PR #3036, opened 2021-02-28
XPRA_ANNOUNCE_TIMEOUT in ui_thread_watcher.py

### LINKED ISSUE #3035: "UI thread is now blocked" when using opengl
Hello,

I am running Debian xpra (`3.0.13+dfsg1-1`), and I'm getting log messages like

```
client   6 @42.106 UI thread is now blocked
client   6 @43.063 UI thread is running again, resuming
```

These fill up the logs (they can be hundreds of megabytes), and do not occur when I use `--opengl=no` on the client.

There are (at least) two situations that trigger these messages

1. Switching to another vt, with any X11 application (e.g. xterm) running.  This (seems to) result in a single blocked/running again pair---corresponding to the switch away from and return to the VT.
2. Allowing X11 DPMS to power down the monitor, with an opengl application (e.g., firefox) running.  Once the screen powers down, I get many pairs of messages with block times upwards of 10 seconds.

I suppose opengl calls block in these situations, and the client is genuinely stalling---but is there maybe a better way to handle this?

aerusso:
The UI_thread_watcher identifies when the client is not responding, and signals to the server.  By default, it also announces this in its log output.  On some systems using OpenGL, this block can occur during the normal enabling of monitor DPMS or a VT switch.  This leads to a large number of these announcements

Provide a mechanism to reduce logging of ui_thread_watcher's detection of by providing a configurable timeout, independent of the ui_thread_watcher's detection threshold, XPRA_ANNOUNCE_TIMEOUT.  

If set to 0, no such announcements will be logged.

Closes #3035

# Testing

I'm currently running this on top of `v3.0.13-r28292`, (Debian testing).

### DIFF
--- a/xpra/platform/ui_thread_watcher.py
+++ b/xpra/platform/ui_thread_watcher.py
@@ -15,6 +15,7 @@ log = Logger("util")
 
 FAKE_UI_LOCKUPS = envint("XPRA_FAKE_UI_LOCKUPS")
 POLLING = envint("XPRA_UI_THREAD_POLLING", 500)
+ANNOUNCE_TIMEOUT = envint("XPRA_ANNOUNCE_BLOCKED", POLLING)
 
 
 class UI_thread_watcher:
@@ -28,11 +29,12 @@ class UI_thread_watcher:
         Beware that the callbacks (fail, resume and alive)
         will run from different threads..
     """
-    def __init__(self, timeout_add, source_remove, polling_timeout):
+    def __init__(self, timeout_add, source_remove, polling_timeout, announce_timeout):
         self.timeout_add = timeout_add
         self.source_remove = source_remove
         self.polling_timeout = polling_timeout
         self.max_delta = polling_timeout * 2
+        self.announce_timeout = announce_timeout/1000.0 if announce_timeout else float('inf')
         self.init_vars()
 
     def init_vars(self):
@@ -40,6 +42,7 @@ class UI_thread_watcher:
         self.fail_callbacks = []
         self.resume_callbacks = []
         self.UI_blocked = False
+        self.announced_blocked = False
         self.last_UI_thread_time = 0
         self.ui_wakeup_timer = None
         self.exit = Event()
@@ -104,7 +107,9 @@ class UI_thread_watcher:
         self.last_UI_thread_time = monotonic_time()
         #UI thread was blocked?
         if self.UI_blocked:
-            log.info("UI thread is running again, resuming")
+            if self.announced_blocked:
+                log.info("UI thread is running again, resuming")
+                self.announced_blocked = False
             self.UI_blocked = False
             self.run_callbacks(self.resume_callbacks)
         return False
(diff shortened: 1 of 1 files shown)


totaam:
I understand why you silence the _UI thread is now blocked_ message only.
Why not also silence the _UI thread is running again, resuming_ message too?

xpra/platform/ui_thread_watcher.py
```diff
@@ -104,7 +107,9 @@ def UI_thread_wakeup(self, scheduled_at=0):
         self.last_UI_thread_time = monotonic_time()
         #UI thread was blocked?
         if self.UI_blocked:
-            log.info("UI thread is running again, resuming")
+            if self.announced_blocked:
+                log.info("UI thread is running again, resuming")
+                self.announced_blocked = False
```
aerusso, line 112: We suppress the `"UI thread is running again, resuming"` message here---`announced_blocked` will never be set unless `announce_timeout` has been exceeded.

totaam: Sorry, my message was not very clear at all!

If `announced_blocked` is not set, then no message is printed - is this intended?

Maybe I misunderstand the symptoms you were seeing?
Can you perhaps post a more complete log sample in #3035 so I can see how often the messages show up?

aerusso: Hey, yeah, I wasn't sure I was understanding your message.  Here's some log output:

    2021-03-02 07:03:14,377 client   2 @03.245 Xpra GTK3 X11 server version 3.0.13-r28292 64-bit
    2021-03-02 07:03:14,377 client   2 @03.245  running on Linux Debian testing bullseye
    2021-03-02 07:03:14,391 client   2 @03.258 Attached to socket:///run/user/1001/xpra/REDACTED-67
    2021-03-02 07:03:14,391 client   2 @03.258  (press Control-C to detach)
    2021-03-02 07:03:14,877 client   2 @03.745 server does not support xi input devices
    2021-03-02 07:03:14,877 client   2 @03.746  server uses: xtest
    2021-03-02 07:03:15,176 New unix-domain connection received
    2021-03-02 07:03:15,176  on '/run/user/1001/xpra/REDACTED-67'
    2021-03-02 07:03:15,177 New unix-domain connection received
    2021-03-02 07:03:15,177  on '/home/REDACTED/.xpra/REDACTED-67'
    2021-03-02 07:03:15,383 client   2 @04.252 Error: cannot access the list of printers
    2021-03-02 07:03:15,384 client   2 @04.252  failed to connect to server
    2021-03-02 07:08:44,496 client   2 @33.364 UI thread is now blocked
    2021-03-02 07:08:48,269 client   2 @37.137 UI thread is running again, resuming
    2021-03-02 07:08:55,999 client   2 @44.868 UI thread is now blocked
    2021-03-02 07:08:56,696 client   2 @45.565 UI thread is running again, resuming

To answer your question directly:

> If `announced_blocked` is not set, then no message is printed - is this intended?

Yes.  I'd like to suppress both the block and resume notifications, since---as I understand it---they are part of the normal operation of XPRA on my system when DPMS engages (at least on my machine).

The block and resume messages always appear in pairs (unbalanced messages should be impossible, right?).

I chose this approach (adding a new environment variable, and not changing the default behavior) because it sounds like these notifications serve a very useful diagnostic role for many people.  If these messages were less important for diagnostics, we might consider changing the loglevel from `.info` to `.debug`, and scrapping this patch.

Issue #2900, opened 2020-10-15
Possible fix for Eclipse repainting problem


totaam:
*Issue migrated from trac ticket # 2900*

**component:** server | **priority:** major | **resolution:** duplicate | **keywords:** environment, eclipse

#### 2020-10-15 06:52:20: wanderyr created the issue

___
> Server: v4.0.4-27557 on Kubuntu 20.04.
> Client: v4.0.4-27557 on Windows 10.
> Using SSH connections.
> 
> I'm running Eclipse 2020-09. Eclipse runs fine (although slowly) via standard X11 forwarding (using MobaXTerm on Windows).
> 
> Within xpra, the eclipse editor windows show the constant repainting mentioned in #2715 and #2070.
> 
> For debugging, I started an xterm within xpra, and then started eclipse from that command line. The flickering is still there.
> `$ eclipse &`
> 
> But, if I do the following within xterm, eclipse works great:
> `$ export GTK_IM_MODULE="ibus"`
> `$ eclipse &`
> 
> That environmental variable is originally set as "xim".
> 
> Is there a way to start the xpra session with this environmental variable preset to "ibus"? I would prefer to use "$ xpra start --start=eclipse" instead of bringing up an intermediate terminal.
> 
> Thanks for all you are doing!
>

totaam:
Can also be improved since fa20fb86f5e678741fc50ac0659a609ef263cacb

PR #3046, opened 2021-03-07
Add atom as text.

cr1901:
Without this change, the Atom editor has latency that makes it unusable (IME :)) over Xpra.


totaam:
FYI: this usually means that `atom` is repainting the screen rapidly and unnecessarily.  
Setting it to `text` mode ensures we don't use a video codec to send its pixel, despite the high framerate and large repaint area.

Eclipse is known to have a similar problem, strangely fixed by switching input method! #2900

PR #3051, opened 2021-03-12
Use sys.executable instead of hardcoded "python3".

arrowd:


totaam:
Thanks.

Note to self: this may cause problems on mingw?

PR #3079, opened 2021-04-09
Dotnet pipe client

TijZwa:
A (very) simple named pipe client which allows us to intercept and mirror certain Xpra messages to dotnet core.
In Dotnetocore we can start a server for incomming connections like this: 

NamedPipeServerStream namedPipeServerStream = new NamedPipeServerStream("xpraDotnetPipe", PipeDirection.In, 1, PipeTransmissionMode.Byte, PipeOptions.Asynchronous);


totaam:
Since the pipe is only used for exposing some messages, we could use dbus events for that instead.
It would make the solution more generic and more generally useful.

The difficulty is choosing how much to expose for each client message.
You have selected the events that interest you (`new-window`, `lost-window`, etc), but we should try to find a solution that is generic enough, yet doesn't just broadcast every packet.

PR #3132, opened 2021-05-22
Wrap Xvfb command string only when generating .conf file

mvnetbiz:
When xvfb command cannot be obtained from a config file, `xpra.scripts.config.xvfb_cmd_str()` is called to generate it, but it contains escaped newlines (to format it nicely when generating `/etc/xpra/conf.d/55_server_x11.conf` at compile time). This makes it only wrap lines at compile time in setup.py.


totaam:
Great find and perfect fix for it!

PR #3137, opened 2021-05-25
Copy 'noicon.png' from xpra-hrml5 to main package

basilgello:
Fixes chatty stacktraces caused by missing 'noicon.png' when
client disconnects from system-wide proxy server.


totaam:
What's the stacktrace you were seeing?

basilgello:
I dropped that container, but the error message was "icon_path must be one of type string, byte string but not NoneType" here: https://github.com/Xpra-org/xpra/blob/2022238ababa40382d2b8d6b019fb2d757ec9955/xpra/server/server_core.py#L1484

And this exception popped like 20 times from http_handler downwards.

I first tried fixing it by adding '/usr/share/xpra/www' to icon path fallback list, but the solution was easier.

PR #3145, opened 2021-05-28
Do not include malloc.h on OpenBSD

marianielias:
Just adding OpenBSD to the chain. :)

Cheers.
Elias.

### DIFF
--- a/xpra/buffers/memalign.c
+++ b/xpra/buffers/memalign.c
@@ -13,7 +13,7 @@
 #ifdef _WIN32
 #define _STDINT_H
 #endif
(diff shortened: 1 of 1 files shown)


totaam:
Thanks.
FYI: [Excessive line breaks are BAD](https://lkml.org/lkml/2020/5/29/1038)

marianielias:
> Thanks.
> FYI: [Excessive line breaks are BAD](https://lkml.org/lkml/2020/5/29/1038)

No problem.
Good reading BTW and I see the point, but some people likes their code "that way".
I got some laugh about the fact that the page on that link uses a 72 chars format... or maybe the e-mail has that...

PR #3152, opened 2021-06-02
Fix lintian warnings

basilgello:
Series of commits fixing Lintian warnings

Another commit that I've postponded is the mismatching license of `fs/share/metainfo/xpra.appdata.xml`:

```xml
 <metadata_license>CC0-1.0</metadata_license>
 <project_license>GPL-2.0+</project_license>
```

Lintian complains on this, but only as warning. To make it happy, one needs to change it like:

```xml
 <metadata_license>GPL-2+</metadata_license>
 <project_license>GPL-2+</project_license>
```

@totaam ?


totaam:
Thanks for doing this. Lots of good things in there.
Can you please separate the whitespace changes in 83bd2cd3a1dc46835f38f8f58068ddd499e60e5c?

aab898fc07037619a8a7c2e46810b41fa1f4b200 says _Fix undefined variable in ignore message_ but there's a whole chunk to do with "png", what is that?

basilgello:
Oh, it is a fragment I squashed by accident that copies local png images during pandoc processing. Should I separate it or just amend the commit msg?

The PNGs are nedded to have full illustrations in local HTML doc. Lintian authors call it `privacy-breach=generic` but it is not the best name for entities loadable from  remote locations in a local help.

basilgello:
Actually I prepared the Debian packaging for master trunk & fixed the issues spotted. Time to package xpra-html5.

PR #3156, opened 2021-06-06
Document 'http-scripts' setting

basilgello:
As discussed, re-opening the PR with only documentation part.

### DIFF
--- a/fs/etc/xpra/conf.d/50_server_network.conf.in
+++ b/fs/etc/xpra/conf.d/50_server_network.conf.in
@@ -91,7 +91,8 @@ ssl = auto
 #html = /var/www/htdocs/
 html = auto
 
-# Which builtin http scripts to allow:
+# Control which HTTP scripts are available to unauthenticated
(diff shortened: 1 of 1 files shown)


totaam (commented):
Good catch. This should be documented in the sample config files.

However, I am reluctant to change the defaults.
If anything, the `Displays` and `Sessions` scripts expose more security sensitive information about the system than the system menus: the system menus only expose what commands are available on the system (none of which can be used without authentication) whereas the sessions and displays have information that can be used for authentication (username).
For a network facing service, the latter is much more sensitive than the former.
The problem is that without those scripts, the connect page loses a lot of its appeal.

Can you change the PR so the defaults remain unchanged for now?

basilgello:
Yes I can try restoring the branch. However, I am thinking about adding the code to HTTP script handlers that checks the authentication mode applied, checks if user is authenticated, returning otherwise 403.

Something like 'all-authenticated' option in addition to 'all'.

PR #3157, opened 2021-06-06
Redirect plain HTTP requests to HTTPS on 'ssl=on'

basilgello:
This PR fixes two issues:

﻿- Use threaded proxy server mode for insecure websockets
- Redirect insecure HTTP GET/HEAD requests to HTTPS

We can of course introduce new possible value to 'ssl' (like
'force', 'required', 'ssl-only') but my bet is that if user
turns on SSL, insecure HTTP requests must be redirected to
HTTPS.

### DIFF
--- a/xpra/net/websockets/handler.py
+++ b/xpra/net/websockets/handler.py
@@ -24,9 +24,12 @@ class WebSocketRequestHandler(HTTPRequestHandler):
 
     def __init__(self, sock, addr, new_websocket_client,
                  web_root="/usr/share/xpra/www/",
-                 http_headers_dir="/etc/xpra/http-headers", script_paths=None):
+                 http_headers_dir="/etc/xpra/http-headers",
+                 script_paths=None,
+                 redirect_https=False):
         self.new_websocket_client = new_websocket_client
         self.only_upgrade = WEBSOCKET_ONLY_UPGRADE
+        self.redirect_https = redirect_https
         super().__init__(sock, addr, web_root, http_headers_dir, script_paths)
 
     def handle_websocket(self):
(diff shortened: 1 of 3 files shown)


totaam (commented):
`Failed to create socket` is a different problem than the one that this code path is meant to solve: ssl and ssh sockets have state that cannot be transferred to a new process and that's why we use threads.
`ws` sockets should work in process mode.
The reason why you get this failure is because the user is not a member of the `xpra` group.
Perhaps we should create the pid file somewhere else?

totaam (changes requested):
For consistency, the `ssl` option should be checked against `TRUE_OPTIONS` and not just `on`.

It would be nice to add an environment toggle so this behaviour can be turned off, ie: `HTTP_HTTPS_REDIRECT` (and make it default to True).

basilgello:
> Perhaps we should create the pid file somewhere else?

It is setup_control_socket, not pidfile. I wonder why control socket can not be created in $XDG_RUNTIME_DIR/xpra for process mode if user is not a member of xpra group (with warning explaining the problem). Let me check :)

totaam:
> It is setup_control_socket, not pidfile

Ah, perhaps this works well only when the proxy run as root.

basilgello:
> Ah, perhaps this works well only when the proxy run as root.

Let me spawn a container running rootless xpra proxy server and check what works there.

basilgello:
> > It is setup_control_socket, not pidfile
> 
> Ah, perhaps this works well only when the proxy run as root.

This works too if `xpra proxy` is started as an unprivileged user. Now trying to implement fallback of socket creation to `$XDG_RUNTIME_DIR` if user is not in `xpra` group and report back what happens.

basilgello:
I re-used the `handle_socket_error` and it appeared that the `xpra` group membership is required for *additional* control socket, not *primary* one (which is in `$XDG_RUNTIME_DIR`). Also I hope I figured out how to get rid of "BrokenPipe" exception inside `Queue` when `stop_proxy` closes queue before the receiver gets the `stop` message.

totaam (commented):
This one is strange, we never clear the `self.message_queue` attribute so the only way this can evaluate to False is if the interpreter is in the process of shutting down? Or am I missing something?

Normally, we should always exit cleanly via one of the methods that call `self.stop(..)` (ie: `signal_quit`) and this will take care of setting `self.exit`.
If there is a path that skips calling `self.stop` then maybe we should fix that instead?

basilgello:
> If there is a path that skips calling self.stop then maybe we should fix that instead?

Interesting possibility! I added the evaluation and so far dozens of test produces no exception.

You can try reproducing yourself:

* Run `xpra proxy :14500 --bind-tcp=127.0.0.1:14500 --tcp-auth allow --daemon no` as unprivileged user
* Attach to running proxy instance with HTML5 client
* Press `Connect` without specifying anything and wait for Xpra display server to spin up.
* Shutdown the Xpra display server via HTML5 and when `connect.html`loads, press `Ctrl+C`in the proxy server console.
* Experiece `BrokenPipe` exception inside feeder after `mq.close()`

Issue #3143, opened 2021-05-28
Unable to start server with version bionic 4.2-r0-1


badmin-c:
I connect to xpra using the following command:
```
xpra start ssh:system.domain.de --env=LANG=de_DE.UTF-8 --attach=True --tray=no --exit-with-children=True --start-child=/usr/bin/firefox
```

```
2021-05-28 09:56:13,826 Xpra GTK3 X11 client version 4.2-r0 64-bit
2021-05-28 09:56:13,959  running on Linux LinuxMint 19.3 tricia
2021-05-28 09:56:13,960  window manager is 'Mutter (Muffin)'
2021-05-28 09:56:14,811 GStreamer version 1.14.5 for Python 3.6.9 64-bit
2021-05-28 09:56:14,861 created unix domain socket '/run/user/1001/xpra/clients/my-desktop-17357'
2021-05-28 09:56:14,946 No OpenGL_accelerate module loaded: No module named 'OpenGL_accelerate'
2021-05-28 09:56:15,206 OpenGL enabled with GeForce RTX 2070/PCIe/SSE2
2021-05-28 09:56:15,296 Connected (version 2.0, client OpenSSH_7.6p1)
2021-05-28 09:56:16,138 Authentication (publickey) successful!
2021-05-28 09:56:16,492  keyboard settings: rules=evdev, model=pc105, layout=de
2021-05-28 09:56:16,497  desktop size is 2560x1440 with 1 screen:
2021-05-28 09:56:16,498   :0.0 (677x381 mm - DPI: 96x96) workarea: 2560x1416
2021-05-28 09:56:16,498     DP-2 (598x336 mm - DPI: 109x109)
2021-05-28 09:56:16,596  SSH: 'Warning: XDG_RUNTIME_DIR is not defined'
2021-05-28 09:56:16,596  SSH: " and '/run/user/1000' does not exist"
2021-05-28 09:56:16,596  SSH: " using '/tmp'"
2021-05-28 09:56:16,721  SSH: 'Warning: XDG_RUNTIME_DIR is not defined'
2021-05-28 09:56:16,722  SSH: " and '/run/user/1000' does not exist"
2021-05-28 09:56:16,722  SSH: " using '/tmp'"
2021-05-28 09:56:16,766  SSH: 'failed to initialize Gtk, no display?'
2021-05-28 09:56:36,630  SSH: 'Error: displayfd failed'
[...]

PR #3164, opened 2021-06-12
Start server before the client

basilgello:
Start server before client if user invokes 'xpra start' or 'start-desktop' with 'attach' option.

For example, the command:
```
xpra start --start=xeyes --attach=yes
```
gets hung because in daemon mode, either client or server is started by 'do_run_mode'.

The following command workedthough:
```
xpra start --start=xeyes --attach=yes --daemon=no
```
Now it works regardless of daemon mode.

### DIFF
--- a/xpra/scripts/main.py
+++ b/xpra/scripts/main.py
@@ -419,22 +419,32 @@ def do_run_mode(script_file, error_cb, options, args, mode, defaults):
 
     if mode in ("start", "start-desktop") and args and parse_bool("attach", options.attach) is True:
         assert not display_is_remote
-        #maybe the server is already running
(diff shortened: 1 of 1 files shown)


totaam:
What version are you using for testing?
I'm not seeing the problem with the commands you specified. (with any branch)

The bug you describe used to exist and was fixed by 583cb7046f968651ea853e41d0f02614a086e2ab.
Around the same time as #3143.
What am I missing here?

basilgello:
Let me rebuild both packages from master trunk & re-check. If the issue goes away, I'll close this PR otherwise I will ACK it exists.

PR #3169, opened 2021-06-14
Add placeholder for COMMIT and BRANCH if build is done

basilgello:
... from a tarball but not Git tree.


totaam:
BTW, tarballs created with `./setup.py sdist` should have the data populated correctly.
Sadly, the ones from github do not as they're automatically generated.

Thanks for fixing this!

PR #3177, opened 2021-06-22
Add python3-distutils as runtime-dep on Debian

basilgello:
'os_util.which()' is called in 'ibus-daemon' search routine
and 'xpra start' fails to start in a pristine Debian bullseye
container due to missing runtime dependency on python3-distutils

### DIFF
--- a/packaging/debian/xpra/control
+++ b/packaging/debian/xpra/control
@@ -40,6 +40,8 @@ Depends: ${misc:Depends}, ${python3:Depends}, ${shlibs:Depends}
         ,python3-gi
         ,gir1.2-gtk-3.0
         ,python3-cairo
+# for spawn
+        ,python3-distutils
         ,python3-gi-cairo
 # for opengl support:
         ,python3-opengl


basilgello:
@totaam pls kick rebuilds of Debian packages after merging this :)

totaam:
> @totaam pls kick rebuilds of Debian packages after merging this :)

Can't do anything until I get home - another 24 hours or so.
I'll try to ensure we don't fail if `which` is missing.

totaam:
@basilgello sorry for the delay.
I've pushed new packages to the beta repo.

But I'm still confused about the error that this is supposed to fix: I've tried to reproduce it by moving `/usr/bin/which` out of the way temporarily and `find_executable` (which is called by `which`) worked without problems.
This isn't surprising since it doesn't actually execute anything and just looks for the executable in `$PATH`.

Exactly how did it fail for you? Do you have the exact backtrace?

basilgello:
Sure!

```python
2021-06-24 20:59:03,565 created tcp socket '0.0.0.0:14500'
/bin/bash: line 1: /usr/bin/python: No such file or directory
2021-06-24 20:59:03,581 Error 127 running source script '/etc/profile'
xpra main error:
  File "/usr/lib/python3/dist-packages/xpra/scripts/main.py", line 128, in main
    return run_mode(script_file, err, options, args, mode, defaults)
  File "/usr/lib/python3/dist-packages/xpra/scripts/main.py", line 408, in run_mode
    return do_run_mode(script_file, error_cb, options, args, mode, defaults)
  File "/usr/lib/python3/dist-packages/xpra/scripts/main.py", line 438, in do_run_mode
    return run_server(script_file, error_cb, options, args, mode, defaults)
  File "/usr/lib/python3/dist-packages/xpra/scripts/main.py", line 2349, in run_server
    return do_run_server(error_cb, options, mode, script_file, args, desktop_display, progress_cb)
  File "/usr/lib/python3/dist-packages/xpra/scripts/server.py", line 694, in do_run_server
    if configure_imsettings_env(opts.input_method)=="ibus":
  File "/usr/lib/python3/dist-packages/xpra/scripts/server.py", line 190, in configure_imsettings_env
    ibus_daemon = which("ibus-daemon")
  File "/usr/lib/python3/dist-packages/xpra/os_util.py", line 826, in which
    from distutils.spawn import find_executable
ModuleNotFoundError: No module named 'distutils'
```

starting `xpra start :1000 --start-child="/bin/bash"`

totaam:
> Sure!

Thanks.
Oh. Never mind, the information was there all along, I just read it wrong!

With the change above, things should work even if distutils is not installed.
It's a trimmed down version of the `find_executable` function, since we don't care about the special cases.

basilgello:
You read my mind on removing distutils ;) I think now we can safely drop that from runtime dependencies again :))))

PR #3215, opened 2021-07-22
use POSIX compliant `command -v` instead of `which`

miojonka:
Xpra currently depends on the tool `which` to be installed to find its own binary on a server. The resulting error message is not helpful as it claims the Xpra binary was not found (i.e. Xpra is not installed when it clearly is), but instead it is the binary-to-find-binaries that was not found. 

On Archlinux for example `which` is not installed by default and `which` in general has different output on different distributions. 

Using `command -v` is POSIX compliant (`which` is not) and will work on every distribution without requiring to have `which` installed. It is also the endorsed way to find the path to a binary in $PATH. 

This PR should fix this (I only tested it on Archlinux).

### DIFF
--- a/xpra/server/ssh.py
+++ b/xpra/server/ssh.py
@@ -147,8 +147,8 @@ class SSHServer(paramiko.ServerInterface):
             cmd = shlex.split(bytestostr(command))
         log("check_channel_exec_request: cmd=%s", cmd)
         # not sure if this is the best way to handle this, 'command -v xpra' has len=3
-        if cmd[0] in ("type", "which", "command") and (len(cmd)==2 or len(cmd)==3):
-            xpra_cmd = cmd[1]   #ie: $XDG_RUNTIME_DIR/xpra/run-xpra or "xpra"
+        if cmd[0] in ("type", "which", "command") and len(cmd) in (2,3):
+            xpra_cmd = cmd[-1]   #ie: $XDG_RUNTIME_DIR/xpra/run-xpra or "xpra"
             if not POSIX:
                 assert WIN32
                 #we can't execute "type" or "which" on win32,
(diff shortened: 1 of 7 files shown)


xpra/server/ssh.py
```diff
@@ -146,7 +146,8 @@ def chan_send(send_fn, data, timeout=5):
         except UnicodeDecodeError:
             cmd = shlex.split(bytestostr(command))
         log("check_channel_exec_request: cmd=%s", cmd)
-        if cmd[0] in ("type", "which") and len(cmd)==2:
+        # not sure if this is the best way to handle this, 'command -v xpra' has len=3
+        if cmd[0] in ("type", "which", "command") and (len(cmd)==2 or len(cmd)==3):
             xpra_cmd = cmd[1]   #ie: $XDG_RUNTIME_DIR/xpra/run-xpra or "xpra"
```
totaam, line 151: OTOH. I think this should be `len(cmd) in (2, 3)`
And then `xpra_cmd = cmd[-1]`?
Otherwise `command -v xpra` would end up with `xpra_cmd = -v`.

miojonka: oops, yes. I changed it, hope it is okay now.

totaam:
Thanks!

totaam:
This had broken the builds that tried to find `nvcc` using the now broken `which()` function.

chewi:
This has broken the tests. Although I agree that `which` is bad, `command` is a shell built-in, so you can't call it with `Popen` via `get_status_output` like you're doing here.

totaam:
@chewi "like you're doing here" - where?

miojonka:
Like here I guess:  https://github.com/Xpra-org/xpra/blob/1c335b5809f9708073ffe5d8c5923aedb312bf99/xpra/os_util.py#L871

For finding the executable maybe `shutil.which` might be more suited. But it requires Python 3.3 at least.

chewi:
Yes, I was going to try `shutil.which`. That's okay, Xpra needs at least Python 3.6 anyway.

totaam:
And yet another breakage is caused by this change: #3610.

PR #3242, opened 2021-08-23
(WIP) Permit compilation with ffmpeg > 4.4

bertiebaggio:
On 2021-04-27, ffmpeg removed some things previously deprecated:
    
- avutil/pixfmt: Remove deprecated VAAPI pixel formats
https://github.com/FFmpeg/FFmpeg/commit/ad524cb9eeb298f7a60d923094fbebb1fda7e0e3
- avformat: Remove deprecated AVFMT_FLAG_MP4A_LATM flag, latm option
https://github.com/FFmpeg/FFmpeg/commit/e8876bdcf3b4a8e5320884ffb9d39ce30e23f0e6
    
These now cause compile errors as the symbols are undefined. This WIP PR comments out those options to permit compilation.

Compilation also fails in some places due to:

    error: assignment discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]

The lines causing these errors have temporarily been commented out to permit compilation.

**Note**: These changes permit compilation, but I have not yet run tests or indeed the application!


totaam:
Thanks for that, I've completely dropped those flags we don't use anyway.
As for the warnings, we now use `const AVCodec*` so this should now build OK on your system.
Can you confirm?

bertiebaggio:
> Thanks for that, I've completely dropped those flags we don't use anyway.
> As for the warnings, we now use `const AVCodec*` so this should now build OK on your system.
> Can you confirm?

Thanks for looking into this.

My commit message omitted to mention `AVFMT_FLAG_KEEP_SIDE_DATA`, deprecated in 2017:

 - https://github.com/FFmpeg/FFmpeg/commit/ca2b7794234cccd6c4ae548c4bd2eac5c53b80f2
 - https://github.com/FFmpeg/FFmpeg/blob/3969c2abfb29c230cfce2ebfdcaaf6d0f0266b30/doc/APIchanges#L718
 - https://lists.ffmpeg.org/pipermail/ffmpeg-cvslog/2017-March/105336.html

I think once that is removed it should compile; that's the only thing my compiler has complained about running against `master`.

Cheers

bertiebaggio:
> I think once that is removed it should compile; it did in my local test which made up this PR, and that's the only thing my compiler complained about running against current `master` after your commits!
> 
> Cheers

Actually to correct myself there's one more `const` change: 

            self.audio_codec = avcodec_find_encoder(AV_CODEC_ID_AAC)                                                    
in

xpra/codecs/enc_ffmpeg/encoder.pyx   
Line 1425

bertiebaggio:
I can confirm that apropos of https://github.com/Xpra-org/xpra/commit/5058864909fdf76652d20c3d555bd91667762f88 xpra now build

thanks for your effort on this, much appreciated!

totaam:
Great!

PR #3261, opened 2021-09-09
Scrub env vars the Python wrapper sets to avoid upsetting the SSH command

DanielO:
If the SSH command is (say) a Python script it may fail with
inscrutable errors about being unable to initialise encoding etc.

### DIFF
--- a/xpra/net/ssh.py
+++ b/xpra/net/ssh.py
@@ -869,6 +869,8 @@ def ssh_exec_connect_to(display_desc, opts=None, debug_cb=None, ssh_fail_cb=ssh_
         cmd = list(display_desc["full_ssh"])
         kwargs = {}
         env = display_desc.get("env")
+        if env is None:
+            env = os.environ.copy()
         kwargs["stderr"] = sys.stderr
         if WIN32:
             from subprocess import CREATE_NEW_PROCESS_GROUP, CREATE_NEW_CONSOLE, STARTUPINFO, STARTF_USESHOWWINDOW
@@ -930,8 +932,6 @@ def ssh_exec_connect_to(display_desc, opts=None, debug_cb=None, ssh_fail_cb=ssh_
                 #sshpass -e ssh ...
                 cmd.insert(0, sshpass_command)
                 cmd.insert(1, "-e")
-                if env is None:
-                    env = os.environ.copy()
                 env["SSHPASS"] = password
                 #the password will be used by ssh via sshpass,
                 #don't try to authenticate again over the ssh-proxy connection,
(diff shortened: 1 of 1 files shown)


DanielO:
I have an SSH wrapper script in Python to connect to exotic systems via bastion hosts and without this change it fails with:
Fatal Python error: initfsencoding: unable to load the file system codec
ModuleNotFoundError: No module named 'encodings'

This was a bit of a pain to debug because it seems stderr gets hidden too which is unfortunate.

xpra/net/ssh.py
```diff
@@ -930,17 +932,18 @@ def ssh_exec_connect_to(display_desc, opts=None, debug_cb=None, ssh_fail_cb=ssh_
...
                 #the password will be used by ssh via sshpass,
                 #don't try to authenticate again over the ssh-proxy connection,
                 #which would trigger warnings if the server does not require
                 #authentication over unix-domain-sockets:
                 opts.password = None
                 del display_desc["password"]
-        if env:
-            kwargs["env"] = env
+        # Scrub out env vars our wrapper set so we don't potentially upset the ssh process
+        for v in ["PYTHONHOME", "PYTHONPATH", "PYTHON", "DYLD_LIBRARY_PATH", "LD_PRELOAD", "LD_LIBRARY_PATH"]:
```
totaam, line 943: Does the python wrapper really set all of these?
I can understand scrubbing `PYTHON*` but the `*LD*` ones could be left alone?
(if the user has set them, he may have a reason)

DanielO: It sets DYLD_LIBRARY_PATH on OSX, so I figured it may set LD_* on FreeBSD/Linux/etc, however I haven't checked.
It would be nice if there was a way to get the "at launch" environment but I guess that's not feasible.

This does only affect the environment for the launched SSH so I thought it would be a reasonable compromise but I agree it's not ideal.

totaam: Ah, so you're on OSX.
The client goes through this script:
https://github.com/Xpra-org/xpra/blob/master/packaging/MacOS/Helpers/PythonExecWrapper

So I see 2 possible solutions:
* modify the script to insert xpra's paths, but preserve any values the variables may already have - this may or may not be enough to make things work as unset variables will still be set to the new value..
* merge your change but only sanitize the environment on MacOS (`if OSX: ..`)

---

FYI: On Fedora, none of these variables are set by the python interpreter:
```
$ python3 -c 'import os;print("\n".join("%s=%s" % (k, v) for k,v in os.environ.items() if k.find("PY")>=0 or k.find("LD")>=0))'
```
(empty output)

I haven't tested MS Windows yet.

DanielO: Ahh, I thought it was a generic tool not something specific to Xpra.
I will try the first approach and see how I go.

totaam:
Done in #3262

PR #3262, opened 2021-09-09
Reset env vars set by Python wrapper on OSX before starting SSH.

DanielO:
The wrapper sets DYLD_LIBRARY_PATH, PYTHON, PYTHONHOME and PYTHONPATH
which can cause inscrutable errors about encoding if our 'ssh' is written
in Python.


totaam:
Thanks!

It would be nice if this chould be generalized so that any command we execute can revert the environment:
https://github.com/Xpra-org/xpra/blob/4d40cc8644352bbafbf131d8cb8343b90ff98152/xpra/server/mixins/child_command_server.py#L198-L205

I'll try to have a go at it later.

Issue #3229, opened 2021-08-01
resolve bytes vs strings conversion issues


totaam:
This change https://github.com/totaam/rencode/commit/4509f95b1041a84cc31ba5fd9af7970cb0bb2605 would allow us to use rencode the way we always wanted to use it: strings come out as strings and bytes come out as bytes...

Before this can happen:
* fix all the issues where we wrongly assumed that bytes and strings are the same thing
* release a new rencode version (and make the same changes to the javascript version: https://github.com/Xpra-org/xpra-html5/issues/84)
* add a toggle so we only enable the new "bytes" mode if the remote end also supports it

totaam:
Already:
* encodings capabilities 63b455761fbd5f982c70a88f74c82c789f84b271
* client handling for `rgb_format` 63c91211ca153193d0db9af1802b31fb4d73132a

Not sure why this existing bug only triggered now: b96aaeecba7636d04bfeb49f59e2b730a0e6cd73

totaam:
After experimenting with removing the lock used to make the non-cythonized version of rencode safe: https://github.com/totaam/rencode/tree/orig-nolock
It's pretty clear that the changes are just too intrusive and make the code a lot slower. (10000 times or more!)
In reality it would not be such a big problem because we could actually instantiate an encoder + decoder context per connection just once - and it is the instantiation that is costly.
Doing the same change to the cython version - the only version actually used - would be much harder.

A better approach is just to "fork" this code directly into the xpra code: `rencodeplus`.
Something I should have done a long time ago.

Supporting multiple connections with different capabilities is only really a problem for servers.
The html5 client has been updated to support bytes without too much trouble: https://github.com/Xpra-org/xpra-html5/commit/1f831636caf39a995f5f5f8296e34320b183235e

totaam:
@basilgello : as per the discussion in bb867cddde5e284658fa45fcc1b7aa895fb84997, I had introduced a surprising bug in a last minute change...

And here's the cause:
```
def f(v):
    print(v)
> x={}
x[f(0)] = f(1)
1
0
```
The expression is evaluated in a different order than I had expected...

basilgello:
Also, Xpra system-wide proxy creates `/run/user/$UID/xpra/$XPRA_SESSION` directories as root… like:

```
$ ls -la /run/user/1000/xpra
загалом 0
drwx------ 3 test test  60 сер  4 12:16 .
drwx------ 9 test test 260 сер  4 12:15 ..
[...]

PR #3286, opened 2021-09-26
nvenc encoder improvements - support a common backing encoding session across windows

winstona:
The current implementation of the nvenc encoder instantiates a new encoding session for each new window or when a window's nvenc video encoder switches to being used. This incurs a (known) setup time each time the encoder is used and prioritized below other encoders. Due to the frequent setup and tear down, the encoder often gets into an unrecoverable state without killing xpra and restarting likely due to 2 reasons:
- **NVIDIA allows only 2 simultaneous encoders** - if 3+ windows are attempting to use the nvenc encoder on a single consumer device, only 2 will ever succeed
- **resources not being freed consistently** - the cuda/nvenc libraries are very sensitive to this and can hold on to an unusable encoder for the duration of xpra running, often completely locking up the nvenc encoder when both available encoders are orphaned

This PR solves the 2 issues above, and in the process adds a few additional benefits:
- **nvenc/hardware accelerated video encoding can easily be done on more than 2 windows simultaneously** since only one encoding session is used
- Due to using a singleton/global encoder, we can **assume a near-zero setup time** allowing more frequent use of the encoder in cases where it would have made more sense to use other encoders
- With the flexibilty of the above, we can **hint the window in focus to always prioritize using the nvenc encoder** for all encoding to keep latency at a minimum

These changes do incur some minor drawbacks however:
- **More frequent IDR keyframes are sent.** Encoding sessions retain state between frame compression except for keyframes (IDR frames). In order to not create image corruption, when the single encoder session is processing images for different windows simultaneously, it needs to send IDR frames more frequently to "reset" the encoding session. The positive side here is that the IDR frame is much faster to compress relative to subsequent frames but at the expense of a much larger size relative - incurring an overall higher bandwidth usage. This is minimized as much as possible though, and if only a single window is using the video encoder it will behave as expected (single IDR frame 0 at the beginning of the encode stream).
- **One large window will slow encoding for smaller windows.** Encoding sessions need to specify dimensions. When a frame is encountered that is larger than the current encoder's dimensions, this will resize the encoder sessions dimensions to be larger. This takes a performance hit on encoding speed for windows that do not need the larger dimension as they still compress the frame with a larger input/output buffer and dimensions. In this implementation, we never resize the encoder back down to an optimum size after the larger window is no longer needed (until xpra is restarted and the encoder is reinitialized). This can easily be fixed in the future by resizing the encoder back down if it no longer needs to render a frame that large.
- **Realtime bandwidth/speed tuning is likely ignored after initialization.** I didn't investigate this much yet, but optimization for changing link speeds/quality is likely ignored for now. This can probably be fixed fairly easily.

**High Level Implementation:**

To minimize code changes, the `Encoder` class is kept mostly intact and can act as a (previous behavior) standalone Encoder reinitialized for each window or as a common singleton instance (abstracted through the `UniqueEncoder` class). There's some python trickery here to transparently pass method calls/variable access from `UniqueEncoder` down to its backing singleton instance thus not requiring any complex rework into how a video encoder is initialized/called. This also allows selective override of any methods/variables in `UniqueEncoder` to override behavior and/or store local data to each `UniqueEncoder`.

Each `UniqueEncoder` has a `unique_id` - this id is used by the singleton `Encoder`'s `compress_image` to determine if it needs to reset the backing encoder session back to frame 0 and send an IDR frame.

`compress_image` checks if the encoder is large enough to render the frame, and if not - triggers a resize of the encoding session (without fully reinitializing it) before continuing on to compress the image.

When `clean()` is called on a `UniqueEncoder` we need to ignore this, as we don't want to cleanup the singleton `Encoder` since it's likely to be used by a future `UniqueEncoder`.

**Notes:**
- Some errors indicate a likely lockup with the encoding session - these when detected will trigger a full reinitialization of the nvenc library or singleton encoder instance, usually successfully recovering within a few seconds.
- A global `encode_lock` is used to ensure parallel `compress_image` calls don't step over each other since they will be using the same backing buffers for IO to the encoder session.
- New environment variable config flags are added:
  - `XPRA_WINDOW_FOCUS_VIDEO_ENCODER` - if `True` try to use the video encoder for all frames of a window in focus, regardless of other hints that would exclude it from using the video encoder (ie text content-type)
  - `XPRA_NVENC_USE_SINGLETON_ENCODER` - if `True` use this PR's implementation of a single backing encoder session for all windows, otherwise use the one encoder session per window implementation
- This PR disables threaded init by default, as I had found (before these changes) that it was much more reliable when not initialized in a separate thread and this PR has only been tested with single-threaded init. In addition, threaded init is likely not even needed anymore for a singleton instance since init will happen very infrequently.

### DIFF
--- a/xpra/codecs/nvenc/encoder.pyx
+++ b/xpra/codecs/nvenc/encoder.pyx
@@ -3328,7 +3328,7 @@ cdef class Encoder:
             msg = "could not open encode session: %s" % (nvencStatusInfo(r) or r)
             log(msg)
             
-            if r == 10:
+            if r == NV_ENC_ERR_OUT_OF_MEMORY:
                 log("dumping lib refs and reloading")
                 global NvEncodeAPICreateInstance, cuCtxGetCurrent
                 NvEncodeAPICreateInstance = None
--- a/xpra/codecs/nvenc/encoder.pyx
+++ b/xpra/codecs/nvenc/encoder.pyx
@@ -11,7 +11,6 @@ from time import monotonic
 import ctypes
 from ctypes import cdll, POINTER
 from threading import Lock
-from pycuda.driver import LogicError
 import random
 import weakref
 import contextlib
@@ -36,6 +35,7 @@ log = Logger("encoder", "nvenc")
 #we can import pycuda safely here,
 #because importing cuda_context will have imported it with the lock
 from pycuda import driver  # @UnresolvedImport
+from pycuda.driver import LogicError
 import numpy
 
 from libc.stdint cimport uintptr_t, uint8_t, uint16_t, uint32_t, int32_t, uint64_t  #pylint: disable=syntax-error
@@ -2703,6 +2703,7 @@ cdef class Encoder:
         assert cuda_context, "missing device context"
         cdef unsigned int w = image.get_width()
         cdef unsigned int h = image.get_height()
+        cdef unsigned int image_stride = image.get_rowstride()
         gpu_buffer = image.get_gpu_buffer()
         cdef unsigned int stride = image.get_rowstride()
         log("do_compress_image(%s) kernel=%s, GPU buffer=%#x, stride=%i, input pitch=%i, output pitch=%i",
@@ -2718,6 +2719,9 @@ cdef class Encoder:
             if w > self.params.encodeWidth:
                 dims_changed = True
             if h > self.params.encodeHeight:
+                dims_changed = True
+            if h*image_stride > len(self.inputBuffer.data):
+                log("compress_image: h_image_stride (%i) is larger than inputbuffer.data (%i), need to resize" % (h*image_stride, len(self.inputBuffer.data)))
                 dims_changed = True
 
             #FIXME: determine when its safe to drop the encoder dimensions back down (no longer need to render a large image)
--- a/xpra/codecs/nvenc/encoder.pyx
+++ b/xpra/codecs/nvenc/encoder.pyx
@@ -1575,9 +1575,9 @@ cdef class UniqueEncoder:
         return Encoder.instance().is_ready()
 
 
-    def compress_image(self, device_context, image, int quality=-1, int speed=-1, options=None, int retry=0):
+    def compress_image(self, image, options=None, int retry=0):
         log("called compress_image with unique_encoder: %s, frames: %i, self id: %i", self, self.frames, id(self))
-        return self.base_encoder().compress_image(device_context, image, quality=quality, speed=speed, options=options, retry=retry, unique_id=id(self))
+        return self.base_encoder().compress_image(image, options=options, retry=retry, unique_id=id(self))
     
     def clean(self, actualClean=False, derefOnly=False):
         Encoder.remove_child(self)
@@ -2670,7 +2670,7 @@ cdef class Encoder:
         assert cuda_device_context, "no cuda device context"
         #cuda_device_context.__enter__ does self.context.push()
         log("called compress_image with common encoder: %s, context: %#x, frames: %i", self, <uintptr_t> self.context, self.frames)
-        log("compress_image%s", (device_context, quality, speed, options, retry))
+        log("compress_image%s", (cuda_device_context, options, retry))
 
         #FIXME: do not lock encode_lock on USE_SINGLETON_ENCODER disabled
         global encode_lock
--- a/xpra/codecs/nvenc/encoder.pyx
+++ b/xpra/codecs/nvenc/encoder.pyx
@@ -3440,6 +3440,7 @@ def init_module():
             #obtain a new cdc, since the previous one could be invalid
             cdc = cuda_device_context(device_id, device)
             with cdc as device_context:
+                options["cuda-device-context"] = cdc
                 test_encodings = []
                 for e in TEST_ENCODINGS:
                     if e in FAILED_ENCODINGS:
--- a/xpra/codecs/nvenc/encoder.pyx
+++ b/xpra/codecs/nvenc/encoder.pyx
@@ -11,6 +11,10 @@ from time import monotonic
 import ctypes
 from ctypes import cdll, POINTER
 from threading import Lock
+from pycuda.driver import LogicError
+import random
+import weakref
+import contextlib
 
 from xpra.os_util import WIN32, LINUX, strtobytes
 from xpra.make_thread import start_thread
@@ -53,10 +57,10 @@ NATIVE_RGB = envbool("XPRA_NVENC_NATIVE_RGB", NATIVE_RGB)
 cdef int LOSSLESS_ENABLED = envbool("XPRA_NVENC_LOSSLESS", True)
 cdef int YUV420_ENABLED = envbool("XPRA_NVENC_YUV420P", True)
 cdef int YUV444_ENABLED = envbool("XPRA_NVENC_YUV444P", True)
-cdef int DEBUG_API = envbool("XPRA_NVENC_DEBUG_API", False)
+cdef int DEBUG_API = envbool("XPRA_NVENC_DEBUG_API", True)
 cdef int GPU_MEMCOPY = envbool("XPRA_NVENC_GPU_MEMCOPY", True)
 cdef int CONTEXT_LIMIT = envint("XPRA_NVENC_CONTEXT_LIMIT", 32)
-cdef int THREADED_INIT = envbool("XPRA_NVENC_THREADED_INIT", True)
+cdef int THREADED_INIT = envbool("XPRA_NVENC_THREADED_INIT", False)
 cdef int SLOW_DOWN_INIT = envint("XPRA_NVENC_SLOW_DOWN_INIT", 0)
 cdef int USE_SINGLETON_ENCODER = envbool("XPRA_NVENC_USE_SINGLETON_ENCODER", True)
 
@@ -1494,23 +1498,42 @@ cdef inline raiseNVENC(NVENCSTATUS ret, msg):
         raise NVENCException(ret, msg)
 
 
+class NVResourceWatch:
+    # cdef int resource_count
+
+    __slots__ = ("resource_count")
+
+
+    def __init__(self):
+        # log("NVResourceWatch init()")
+        self.resource_count = 0
+
+    def __enter__(self):
+        # log("NVResourceWatch enter()")
+        with open("/proc/self/maps", 'r') as f:
+            self.resource_count = len([ a for a in f.read().split("\n") if 'nvid' in a])
+        
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        # log("NVResourceWatch exit()")
+        # return
+        after_resource_count = 0
+        with open("/proc/self/maps", 'r') as f:
+            after_resource_count = len([ a for a in f.read().split("\n") if 'nvid' in a])
+        if self.resource_count <= after_resource_count:
+            log.warn("NVIDIA Resources should have decreased, but didnt, before: %i, after: %i", self.resource_count, after_resource_count)
+
 cdef class UniqueEncoder:
-    cdef unsigned long unique_id
     cdef object base_encoder
 
     cdef object __weakref__
     
     def __init__(self, *args, **kwargs):
         log("init uniqueencoder start")
-        self.unique_id = random.randint(0,999999999)
-        self.base_encoder = Encoder.instance(self)
+        # maintain weakref to base_encoder to allow deallocation
+        self.base_encoder = weakref.ref(Encoder.instance(self))
 
     def __getattribute__(self,name):
         # log("getattribute unique: lookup %s", name)
-
-        # Override specifc names for cdef vars to return correctly
-        if name == 'unique_id':
-            return self.unique_id
 
         # try to get attr for this object (UniqueEncoder)
         try:
@@ -1521,7 +1544,7 @@ cdef class UniqueEncoder:
 
         # try to get attr from Encoder singleton
         try:
-            attr = Encoder.__getattribute__(self.base_encoder, name)
+            attr = Encoder.__getattribute__(self.base_encoder(), name)
             #log("getattribute unique: found attr in base_encoder, using base: %s", name)
             return attr
         except AttributeError:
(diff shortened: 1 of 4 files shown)


totaam:
> NVIDIA allows only 2 simultaneous encoders

Other ways around this problem are:
* using a quadro card
* https://github.com/keylase/nvidia-patch

> These changes do incur some minor drawbacks however

That's going to penalize the quadro users.

I'll try to find the time to review this PR and see if we can at least merge some of it straight away.
ie: things like `can_ignore_dim_changes` can be generalized and renamed, but the removal of `strget` looks like a mistake.
The debug logging and extra asserts look good, the hard-coded 4K limit does not.
And things like `if r == 10:` will need a docstring.
The `has_focus` flag should be updated on every entry, not on every batch delay calculation. It makes no difference now, but it might in the future.
etc.

xpra/codecs/nvenc/encoder.pyx
```diff
@@ -13,6 +13,8 @@ import ctypes
 from ctypes import cdll, POINTER
 from threading import Lock
 from pycuda import driver
+from pycuda._driver import LogicError
```
totaam, line 16: Should be `from pycuda.driver import LogicError` or just refer to it as `driver.LogicError`.

xpra/codecs/nvenc/encoder.pyx
```diff
@@ -1488,7 +1496,132 @@ cdef inline raiseNVENC(NVENCSTATUS ret, msg):
...
 
+cdef class UniqueEncoder:
+    cdef unsigned long unique_id
+    cdef object base_encoder
+
+    cdef object __weakref__
+    
+    def __init__(self, *args, **kwargs):
+        log("init uniqueencoder start")
+        self.unique_id = random.randint(0,999999999)
```
totaam, line 1507: This is not a safe way to generate a unique id. Use `AtomicInteger` instead.

xpra/codecs/nvenc/encoder.pyx
```diff
@@ -1410,10 +1413,15 @@ def get_spec(encoding, colorspace):
     global MAX_SIZE
     max_w, max_h = MAX_SIZE.get(encoding, (4096, 4096))
     has_lossless_mode = colorspace in ("XRGB", "ARGB", "BGRX", "BGRA", "r210") and encoding=="h264"
+    codec_class = Encoder
+    setup_cost = 80
+    if USE_SINGLETON_ENCODER:
+        codec_class = UniqueEncoder
+        setup_cost = 0
```
totaam, line 1420: Even with these changes, I doubt that the setup cost is as low as a pure software encoder.

xpra/codecs/nvenc/encoder.pyx
```diff
@@ -1668,7 +1813,7 @@ cdef class Encoder:
         profile = os.environ.get("XPRA_NVENC_PROFILE", "")
         profile = os.environ.get("XPRA_NVENC_%s_PROFILE" % csc_mode, profile)
         #now see if the client has requested a different value:
-        profile = options.strget("h264.%s.profile" % csc_mode, profile)
+        profile = options.get("h264.%s.profile" % csc_mode, profile)
```
totaam, line 1816: `strget` is needed because we get `options` from the client, through a packet encoder which may mangle bytes and strings. (#3229)

xpra/codecs/nvenc/encoder.pyx
```diff
@@ -2321,15 +2568,34 @@ cdef class Encoder:
...
         raiseNVENC(r, "flushing encoder buffer")
 
-    def compress_image(self, device_context, image, int quality=-1, int speed=-1, options=None, int retry=0):
+    def compress_image(self, device_context, image, int quality=-1, int speed=-1, options=None, int retry=0, unsigned long unique_id=0):
         assert device_context, "no cuda device context"
         #cuda_device_context.__enter__ does self.context.push()
+        log("called compress_image with common encoder: %s, context: %#x, frames: %i", self, <uintptr_t> self.context, self.frames)
+        log("compress_image%s", (quality, speed, options, retry))
+
+        #FIXME: do not lock encode_lock on USE_SINGLETON_ENCODER disabled
```
totaam, line 2577: Actually, I think we do want a lock in all cases.
With multiple concurrent users each having their own encode thread, this could easily break nvenc.
(but the lock would need to protect more than just `compress_image` in that case)

xpra/codecs/nvenc/encoder.pyx
```diff
@@ -2340,11 +2606,34 @@ cdef class Encoder:
...
         log("do_compress_image(%s) kernel=%s, GPU buffer=%#x, stride=%i, input pitch=%i, output pitch=%i",
             image, self.kernel_name, int(gpu_buffer or 0), stride, self.inputPitch, self.outputPitch)
+
+        self.input_width = roundup(w, 32)
+        self.input_height = roundup(h, 32)
+        self.encoder_width = roundup(w, 32)
+        self.encoder_height = roundup(h, 32)
+
+        dims_changed = False
+        if USE_SINGLETON_ENCODER and self.instance_initialized():
```
totaam, line 2616: I think it would be worth having support for `dims_changed` even without the singleton encoder.

xpra/codecs/nvenc/encoder.pyx
```diff
@@ -2357,8 +2646,16 @@ cdef class Encoder:
...
                 stride = self.copy_image(image, False)
+                log("memcpy_htod(cudaInputBuffer=%s, inputBuffer=%s)", self.cudaInputBuffer, self.inputBuffer)
                 driver.memcpy_htod(self.cudaInputBuffer, self.inputBuffer)
-            self.exec_kernel(cuda_context, w, h, stride)
+            try:
+                self.exec_kernel(cuda_context, w, h, stride)
+            except LogicError:
+                if USE_SINGLETON_ENCODER:
+                    log("got kernel error exception, cleaning global instance")
+                    Encoder.clean_instance()
```
totaam, line 2656: Is this going to re-initialize nvenc or just the encoder context?
Do we want to do this for the non singleton case too?

xpra/codecs/nvenc/encoder.pyx
```diff
@@ -2536,16 +2834,17 @@ cdef class Encoder:
...
         picParams.inputPitch = self.outputPitch
         picParams.inputBuffer = input
         picParams.outputBitstream = self.bitstreamBuffer
         #picParams.pictureType: required when enablePTD is disabled
         if self.frames==0:
             #only the first frame needs to be IDR (as we never lose frames)
+            #need to force IDR to get a frame on a dirty encode session
             picParams.pictureType = NV_ENC_PIC_TYPE_IDR
-            picParams.encodePicFlags = NV_ENC_PIC_FLAG_OUTPUT_SPSPPS
+            picParams.encodePicFlags = NV_ENC_PIC_FLAG_OUTPUT_SPSPPS | NV_ENC_PIC_FLAG_FORCEIDR
```
totaam, line 2847: I guess there's no harm in having this for the non-singleton case.

xpra/codecs/nvenc/encoder.pyx
```diff
@@ -2904,6 +3217,15 @@ cdef class Encoder:
             last_context_failure = monotonic()
             msg = "could not open encode session: %s" % (nvencStatusInfo(r) or r)
             log(msg)
+            
+            if r == 10:
```
totaam, line 3221: Why 10?
What does this mean?

xpra/server/window/window_source.py
```diff
@@ -1082,6 +1088,7 @@ def is_cancelled(self, sequence=INFINITY):
 
 
     def calculate_batch_delay(self, has_focus, other_is_fullscreen, other_is_maximized):
+        self.has_focus = has_focus
```
totaam, line 1091: As I mentioned earlier, I don't like having a flag that's updated on a timer when we have better data available.
Also:
* why is focus so important?
* why not take into account OR windows?

We could use `window.get("focused", False)` every time `damage` is entered.

xpra/server/window/window_video_source.py
```diff
@@ -328,8 +330,8 @@ def video_context_clean(self):
         if csce or ve:
             if DEBUG_VIDEO_CLEAN:
                 log.warn("video_context_clean() for wid %i: %s and %s", self.wid, csce, ve)
-                import traceback
-                traceback.print_stack()
+                # import traceback
+                # traceback.print_stack()
```
totaam, line 334: 3d227dc50883d1a2f4a94575704a2ec850df776d

xpra/server/window/window_video_source.py
```diff
@@ -457,13 +461,13 @@ def lossless(reason):
         if self.is_tray:
             return nonvideo(100, "system tray")
         text_hint = self.content_type=="text"
-        if text_hint and not TEXT_USE_VIDEO:
+        if text_hint and not TEXT_USE_VIDEO and not (WINDOW_FOCUS_VIDEO_ENCODER and self.has_focus):
```
totaam, line 464: What is the benefit of using nvenc for text?
Where is `TEXT_USE_VIDEO` defined?

xpra/server/window/window_video_source.py
```diff
@@ -723,6 +727,8 @@ def matches_video_subregion(self, width, height):
         return vr
 
     def subregion_is_video(self):
+        if WINDOW_FOCUS_VIDEO_ENCODER and self.has_focus:
+            return True
```
totaam, line 731: This feels like the wrong place for it.
What is this meant to do?

xpra/server/window/window_video_source.py
```diff
@@ -778,7 +784,7 @@ def send_nonvideo(regions=regions, encoding=coding, exclude_region=None, get_bes
         assert not self.full_frames_only
 
         actual_vr = None
-        if vr in regions:
+        if vr in regions or (WINDOW_FOCUS_VIDEO_ENCODER and self.has_focus):
```
totaam, line 787: As above, this looks wrong.

xpra/server/window/window_video_source.py
```diff
@@ -1079,22 +1087,35 @@ def update_encoding_options(self, force_reload=False):
...
             else:
                 old = vs.rectangle
                 ww, wh = self.window_dimensions
-                vs.identify_video_subregion(ww, wh,
+                
+
+                log("update_encoding_options check setregion override")
+                if WINDOW_FOCUS_VIDEO_ENCODER and self.has_focus:
+                    vs.set_detection(False)
+                    vs.set_region(0, 0, ww, wh)
```
totaam, line 1109: So this also bypasses video region detection.
Why was this needed?
Now the whole window is using nvenc, no matter what?

xpra/server/window/window_video_source.py
```diff
@@ -1237,9 +1262,12 @@ def checknovideo(*info):
...
         elif ve.get_width()!=enc_width or ve.get_height()!=enc_height:
-            scorelog("check_pipeline_score(%s) change of video input dimensions from %ix%i to %ix%i",
-                     force_reload, ve.get_width(), ve.get_height(), enc_width, enc_height)
-            clean = True
+            if hasattr(ve, 'can_ignore_dim_changes') and ve.can_ignore_dim_changes():
+                pass
+            else:
+                scorelog("check_pipeline_score(%s) change of video input dimensions from %ix%i to %ix%i",
+                        force_reload, ve.get_width(), ve.get_height(), enc_width, enc_height)
+                clean = True
```
totaam, line 1270: The `can_ignore_dim_changes` should be added to all the encoders interface after being renamed to a more generic name, ie: `can_compress_size(w, h)`.

xpra/server/window/window_video_source.py
```diff
@@ -1671,9 +1699,12 @@ def do_check_pipeline(self, encodings, width, height, src_format):
...
-            videolog("do_check_pipeline video: window dimensions have changed from %sx%s to %sx%s",
-                                            ve.get_width(), ve.get_height(), encoder_src_width, encoder_src_height)
-            return False
+            if hasattr(ve, 'can_ignore_dim_changes') and ve.can_ignore_dim_changes():
+                pass
+            else: 
+                videolog("do_check_pipeline video: window dimensions have changed from %sx%s to %sx%s",
+                                                ve.get_width(), ve.get_height(), encoder_src_width, encoder_src_height)
+                return False
         return True
```
totaam, line 1708: As above.

totaam (commented):
I have merged a bunch of mostly cosmetic changes.
Please rebase and address or answer the review comments.

winstona:
> Other ways around this problem are:
>     * using a quadro card
>     * https://github.com/keylase/nvidia-patch

I missed mentioning allowing only 2 encoders on consumer cards specifically - the goal here was to have a solution that would work on consumer cards (and working within their limitations). I didn't come across keylase/nvidia-patch in my investigation into this, but that certainly opens up a lot of extra possibilities on its own making this PR less beneficial. I likely would have structured this a bit differently from the beginning had I been aware.

My goal was to eventually add in a pool of encoders with preference/stickiness for the last encoder it used to keep IDR frames to a minimum if possible with the option to specify the number of encoders you'd want to use. This would possibly allow for best of all worlds as this wouldn't limit quadro cards, would be able to use a higher value for patched consumer drivers, and would be able to work within the 2 encoder limit on non-patched consumer cards. I'd imagine running 1 encoder per window even in a quadro/patched case does take up extra VRAM that can't be used for other work or other windows.

I've been trying to use xpra with things like a browser, where its not just straight video encode that initializes once, and instead will be frequent stops and starts - software encoding/compression in this case just isn't responsive enough to keep up, and the shift between software and nvenc encoding seems to cause extra delays and frame jitter (where it jumps back to an old frame for a second). In addition, having at least one encoder ready to go without taking the initialize time hit so far seems to help keep overall perceived latency down.

Thanks for reviewing this PR and adding feedback - I'll run through those soon and make adjustments/comments where appropriate.

xpra/codecs/nvenc/encoder.pyx
winstona: This was chosen merely to give strong preference to nvenc over x264 software encoder - I think a small value like 5 or 10 might still keep it preferred over software encoding. Of course the actual setup cost is > 0, but the setup cost for an already initialized singleton encoder would be = 0 when initializing the `UniqueEncoder`

xpra/codecs/nvenc/encoder.pyx
winstona: I'm not sure exactly what is going on here but months ago using `strget` caused the nvenc encoder to exception out on initialization where I was running it, so I had to add a manual hack changing it to `get` just to get it to work in the first place without crashing on me and disabling nvenc entirely. Currently I'm using the `nvidia/cudagl:11.0-base-ubuntu20.04` docker container, however I believe this was the case with just a vanilla ubuntu 20.04

xpra/codecs/nvenc/encoder.pyx
winstona: I was assuming that multiple threads with their own cuda context and encoder would not need a global lock holding them (this is a global/module level var created) and could operate in parallel without stepping over eachother - the only reason why this is here is to ensure that parallel threads don't step on eachother as they're writing/reading from the same buffers/memory locations (in the singleton case).
Do you suspect that we'd need a global lock in the case where a singleton isn't used?

xpra/codecs/nvenc/encoder.pyx
winstona: this doesn't reload the nvenc libraries in this case - just cleaning up the encoder context through the usual `clean()`, `cuda_clean()`, etc. and letting it reinitialize on the next frame/attempt. While testing this out, I kept running into issues where the kernel wouldn't init correctly and break any future attempts to encode. It was easier to just dump the whole thing and let it start over with a new encoder context.
It probably would be good to detect if this is happening too frequently and do something different, but so far not seeing it happen too much.

xpra/codecs/nvenc/encoder.pyx
winstona: I can go back and dig this up, but I believe this is the scenario where you get `NV_ENC_ERR_OUT_OF_MEMORY` after you've used up all 2 encoders available on consumer cards - it looks like that is already defined, so I should just be able to replace `10` with `NV_ENC_ERR_OUT_OF_MEMORY`

xpra/server/window/window_source.py
winstona: The intent of focus here is to try to ensure we have nvenc initialized and used for windows that are in focus/active regardless of what type of window they are to keep responsiveness high while letting other windows use their previous scoring (ie text windows not using video encoders at all or video subregions holding on to a video encoder during the entire time video is being used).
I'm not very familiar with all aspects of the code, so if there's a better way of getting this info that is performant enough (using window get focused or otherwise) I'm sure that'll work better. This was just where `has_focus` was available in window_source to capture. In this scenario, `has_focus` is only needed as a best effort, so if its not precisely accurate it doesn't hurt anything - but as you mentioned, that could change in the future or other things could use that param that relies on it being accurate.
I'll try out `self.has_focus = window.get("focused", False)` in `damage` and see if that works as expected

xpra/server/window/window_video_source.py
winstona: I took this out because when looking at logs this appeared to be a stack trace from an exception when it actually wasn't and kept masking the actual problems when skimming through logs. Either way works - this one isn't critical

xpra/server/window/window_video_source.py
winstona: I believe `TEXT_USE_VIDEO = envbool("XPRA_TEXT_USE_VIDEO", False)` was already there, so it's nothing new - however the goal here (as also mentioned above) was to push nvenc to be used as much as possible for a window in focus, and things like text content-type would override that and force non-video

xpra/server/window/window_video_source.py
winstona: probably not ideal, yes - here again the intent was to short-circuit any video region detection for a window and always use nvenc for the active/focused window

xpra/server/window/window_video_source.py
winstona: also attempting to short-circuit the logic to use nvenc where possible on the active window

xpra/server/window/window_video_source.py
winstona: yes - the intent was to let the nvenc encoding/compression optimize sending the whole window. I noticed in a lot of cases since partial window refreshes were being sent, the window refreshed in blocks and not all at the same time - since we already are going to try to use nvenc to encode the frame (since the window is in focus) might as well let it handle the optimization and get consistent repaints on the client side.
If there's a better way to do/handle this let me know and I can update

winstona:
moving to draft as I have pending changes and identified a memory leak resulting from leftover nvenc resources

totaam:
Can you update the PR?

winstona:
I've been working on this off and on, and have been trying to validate and ensure any potential long term memory leaks are addressed fully (some of these were a bit hidden and hard to track in the nvenc library).
I've rebased off master, and I'll hopefully have something soon to be considered mergeable - once I get that together (I have a few other changes not committed yet) I'm thinking of getting this merged as disabled by default (enabled by env variable) and follow up with a separate PR adding an encoder pool of n encoders instead of just one so this doesn't limit the quadro cards.

totaam:
Any update?

winstona:
sorry about the delay and updating this one. I've had to step away from this for a bit, but I've been using this daily and working out the last bits if I notice them. I'll need to rebase again off master but will get updates as mentioned soon - I didn't realize that there was also some interest in possibly using this method (at least until nvenc pooling is implemented).

winstona:
rebased off of latest master - it appears to be functional at this point, however I have a few comments to fix/address before considering this ready to merge

totaam:
I would like to merge the resizing part first, and the encoder pool separately - ideally in a separate module.

totaam:
Not heard back :disappointed:

PR #3303, opened 2021-10-14
run_scaled: support passing arguments to Xpra

dancek:
I needed to pass some arguments to Xpra when running Wine software with `run_scaled`. I think this might have other use cases than mine, so I added support for passing generic extra arguments.

While I was at it, I fixed a bug in the manpage.

### DIFF
--- a/fs/bin/run_scaled
+++ b/fs/bin/run_scaled
@@ -67,15 +67,16 @@ def get_screen_spec():
 #default value:
 scale = 2
 command_argv = []
+extra_xpra_args = None
 
-# Split at last "--" and pass the args after that to Xpra
+# If "--" is present, the command is after that
 if "--" in sys.argv:
-    split_ridx = sys.argv[::-1].index("--")
-    args = sys.argv[1:-split_ridx-1]
-    extra_xpra_args = sys.argv[-split_ridx:]
+    split_idx = sys.argv.index("--")
+    args = sys.argv[1:split_idx]
+    command_argv = sys.argv[split_idx+1:]
+    extra_xpra_args = []
 else:
     args = sys.argv[1:]
-    extra_xpra_args = []
 
 for i, x in enumerate(args):
     if x in ("--help", "-h"):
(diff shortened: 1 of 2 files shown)


dancek:
I actually considered that.

Considering the UI of run_scaled, I'd find it confusing because run_scaled is basically a wrapper around an X application. It also wraps Xpra, but a user won't think of Xpra when they write `run_scaled some_application`. And since it's conventional to pass arguments after `--` to the wrapped software, it would feel like `run_scaled some_application -- --param` would call `some_application --param`.

Another point is that running applications that themselves expect the `--` separating argument would be difficult to call if run_scaled itself used that.

Now, if `--xpra-arg=` sounds like a bad idea, I'd consider something like `XPRA_ARGS="--arg1 --arg2" run_scaled some_application` instead. Is that better?

dancek:
I implemented the `--` separator and force-pushed the changed commits. (I think it's cleaner not to have a wrong implementation in git history.)

> In that case, you must specify it before theapp. `run_scaled -- theapp -- --whatever`

This won't work in my implementation. But hopefully no one needs it.

dancek:
The Github web UI is down for me right now, but I can probably comment
with email too? I slept on this and I think the reasonable way to do
it is to split on the last `--` instance. I wrote that and pushed the
commit.

Now with something like

```
run_scaled --scale=1.5 someapp --arg -- --more-args -- other-stuff -- --xpra-arg
```

the command to run is `someapp --arg -- --more-args -- other-stuff`.
This looks workable to me, though not very intuitive.

On Fri, 15 Oct 2021 at 08:06, Antoine Martin ***@***.***> wrote:
>
> @totaam commented on this pull request.
>
> ________________________________
>
> In fs/bin/run_scaled:
>
> >          scale = parse_scale(x[len("--scale="):])
> +    elif x == "--":
> +        extra_xpra_args = sys.argv[i+2:]
>
> This will pass arguments after -- to xpra and does not take into account he possibility that there might be more than one -- in the command line.
> I'll take a stab at it later today.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android.



-- 
Hannu Hartikainen

Issue #3250, opened 2021-08-28
use stricter Cython compilation options and warnings


totaam:
Already started some refactoring in a991f6e2d5a1cc10015b2d3a88c3859a0cbb7af4.

totaam:
The warnings are documented here: https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#warnings
And the compiler directives are here:
https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiler-directives
There are actually many more undocumented ones in the source code:
https://github.com/cython/cython/blob/8af0271186cc642436306274564986888d5e64c8/Cython/Compiler/Options.py#L180

Global compiler options are here:
https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiler-options

With the new warnings, we're now hitting some cython bugs that cause invalid warnings to be emitted:
* https://github.com/cython/cython/issues/3924
* https://github.com/cython/cython/issues/1699

totaam:
So, some real bugs were found. Yay!

But many more spurious warnings are being emitted:
* various exception handlers give a warning like this one: `warning: xpra/codecs/libav_common/av_log.pyx:107:28: Unused result in 'e'`. Which is bogus as `e` is clearly being used in the exception handler.
* `rencodeplus` (592f27ad1a4283c6ca37bcc5abfa1db7e92eb027) triggers many bogus `Unused result` / `Unused argument` warnings
* cython seems to be confused about values used as both input and output, ie: `warning: xpra/net/vsock.pyx:95:9: Unused result in 'socklen'` calling `int getsockname(int sockfd, sockaddr *addr, socklen_t *addrlen)`
* in some cases, variables are initialized to 0 or `NULL` unnecessarily but I just prefer it that way, as the code is easier to read and verify for safety - so the warnings will stay... In other cases (as above), it is the warning that is wrong!

totaam:
As of 797d141, to enable the extra warnings use: `--with-verbose`.

PR #3345, opened 2021-11-05
Fix libavformat does not support mp4

TijZwa:
Fix for the errors;
ffmpeg: h264+mp4 encoding failed: libavformat does not support mp4
ffmpeg: vp8+webm encoding failed: libavformat does not support webm

The error is caused by the fact that no muxers are returned by av_muxer_iterate. The ffmpeg docs state:
_Parameters: opaque a pointer where libavformat will store the iteration state. Must point to NULL to start the iteration._


totaam:
This broke a month ago in 802270f1a9c694f5e7dd37c617207b8391e23b44 for #3250.
These cython compilation warnings are now disabled by default because of all the false positives: 797d1419472d493125c0b5a7d10bacca99b08158

PR #3353, opened 2021-11-13
add parameters to improve/adjust scroll speed on OSX clients

winstona:
Scrolling as an OSX client is incredibly fast and jumpy - turning this down in the OS causes native OS apps to scroll very slowly.

This change (when using an OSX client) tunes the scrolling speed/acceleration to something reasonable while maintaining the defaults on other OS clients.

Additionally I had to use (somewhere between 20 and 30): 

`XPRA_MOUSE_SCROLL_MULTIPLIER=20`

for a reasonably expected scroll speed on OSX.

Note: this applies only when not using uinput scrolling

### DIFF
--- a/xpra/client/mixins/window_manager.py
+++ b/xpra/client/mixins/window_manager.py
@@ -11,6 +11,7 @@ import os
 import errno
 import signal
 import datetime
+import math
 from collections import deque
 from time import sleep, time, monotonic
 from queue import Queue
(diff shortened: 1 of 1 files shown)


totaam:
> Note: this applies only when not using uinput scrolling

Am I understanding this correctly: without `uinput` on the server side, those tweaks are not needed?
If that's the case, then perhaps it is the `uinput` server side that should be adjusted instead?

winstona:
I only mention `uinput` because I believe that enables precise scrolling (ie. `self.server_precise_wheel`) - and if precise wheel is available, we don't even go through this codepath.
I'm not 100% sure if scrolling is more accurate when using precise wheel with an OSX client, it could have the same behavior or it could behave more naturally and not need any tweaks. I haven't been able to get uinput working so far, so since I wasn't able to validate it myself, these tweaks only apply to the non-uinput/precise wheel scenario (I believe that's using `XTest`).

The sqrt idea originally came from this: https://github.com/Xpra-org/xpra/blob/bdc5887f3360ad77871456a7329f645154d99c94/xpra/platform/darwin/gui.py#L567-L569

which is either not used anymore or is used in other UI components. In this case it appears to only apply to OSX as well.

Additionally, since I believe we only see this scrolling behavior on OSX, it seemed like making adjustments on the server side would apply these changes to all clients and may likely cause undesirable behavior on other clients aside from OSX.

If this seems to be more appropriate somewhere else, let me know and I can make those changes

totaam:
Oh, sorry, I had misread your original comment:
> Note: this applies only when not using uinput scrolling

PR #3398, opened 2021-12-24
Fix error that occurs when make_session_dir fallback already exists

chewi:
If the usual session directory cannot be created, a temporary directory is used as a fallback. However, there is no check to ensure that the fallback does not already exist, leading to an error.

Xpra requires at least Python 3.6 now, so it is okay to use Python 3.3's `exist_ok` when calling `os.makedirs`.

Following this, we now always correct the owner of the directory for consistency. This was previously skipped when the usual directory (but not the fallback) already existed.

### DIFF
--- a/xpra/scripts/server.py
+++ b/xpra/scripts/server.py
@@ -374,7 +374,7 @@ def make_session_dir(mode, sessions_dir, display_name, uid=0, gid=0):
         session_dir = osexpand(os.path.join(tempfile.gettempdir(), display_name.lstrip(":")))
         os.makedirs(session_dir, 0o750, exist_ok=True)
     ROOT = POSIX and getuid()==0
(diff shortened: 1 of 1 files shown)


totaam:
Thanks.
Why the other changes? Why not just add `exist_ok=True` alone?

chewi:
If you mean removing `if not os.path.exists(session_dir)`, there is no point checking when `exist_ok` will do it for you. The former is technically subject to a race condition too.

If you mean `if ROOT and (session_dir.startswith(...))`, I mentioned that in the second commit. We want to match `/run/user`, but not `/run/userx`.

totaam:
> If you mean ...

Neither of these things: I meant that your change would end up calling `chown` every time the code went through this function rather than only when it created the new directory.
I wanted to leave this behaviour unchanged, hence the commits I have just pushed.

PR #3399, opened 2021-12-24
Respect XDG_* environment variables when running tests

chewi:
Gentoo sets `XDG_RUNTIME_DIR` to an isolated directory when building and testing packages, so we don't expect test files to be written to the fallback location of `/tmp`.


totaam:
This could cause tests to fail or to behave unpredictably.
If you only need `XDG_RUNTIME_DIR` then please whitelist just this one key.

chewi:
Fair enough, I've adjusted it.

PR #3404, opened 2021-12-29
opts.password_file is a list of strings, not a string itself

chewi:
Note [this line](https://github.com/Xpra-org/xpra/blob/e58b25c071698846c098890a72f66671db6955cf/xpra/scripts/parsing.py#L1580), which sets the action to `append`. Admittedly I'm not sure whether this is the right fix. What is this code supposed to do when more than one password file is given?


totaam:
Thanks, I think this is the right fix but I have simplified it.

PR #3405, opened 2021-12-29
Read password file as a string, not bytes

chewi:
We do comparisons against other strings, such as the URI.


totaam:
I have merged this but can you please provide a command line that exposed the problem?

chewi:
I don't use the feature myself, but the unit tests failed after dealing with a5b5971874f5756cf2a87eb7dd77b51eba34c86e and #3404 IIRC.

PR #3421, opened 2022-01-11
Don't close all open files when redirecting stdio

jktomer:
Conventional wisdom holds that a process about to background should first
close all its open file descriptors, so as to prevent holding on to any open
files it may have inherited from a careless parent.

This might have made sense back in the days of single-threaded programs with
no library dependencies out of its control, but it is most certainly not a
safe thing to do willy-nilly any more.

To wit, on my system, an nsswitch plugin starts a background thread that opens
some sockets on which it depends for its operation. Xpra then cheerfully,
mercilessly closes them all in preparation for daemonizing, and then makes
completely unrelated open calls that recycle those file descriptor numbers,
causing absolute havoc as the main program and the background thread share a
file descriptor they really did not expect to be sharing.

There is no possible way for the xpra server script to know this is happening
and gather a set of file descriptors to preserve. Instead we should close only
file descriptors we know we shouldn't have open any more. In xpra's case these
are only the stdio file descriptors that we are redirecting anyway.

Since we no longer violently close all file descriptors when daemonizing, we
no longer need to carefully track the extra files we open when creating a PAM
session, either; this change removes the code that does so.

### DIFF
--- a/xpra/os_util.py
+++ b/xpra/os_util.py
@@ -559,32 +559,6 @@ def livefds():
                 live.add(fd)
     return live
 
-def get_all_fds():
-    fd_dirs = ["/dev/fd", "/proc/self/fd"]
-    fds = []
-    for fd_dir in fd_dirs:
-        if os.path.exists(fd_dir):
-            for fd_str in os.listdir(fd_dir):
-                try:
-                    fd = int(fd_str)
-                    fds.append(fd)
-                except OSError:
-                    # This exception happens inevitably, because the fd used
-                    # by listdir() is already closed.
-                    pass
-            return fds
-    sys.stderr.write("Uh-oh, can't close fds, please port me to your system...\n")
-    return fds
-
-def close_all_fds(exceptions=()):
-    for fd in get_all_fds():
-        try:
-            if fd not in exceptions:
-                os.close(fd)
-        except OSError:
-            # This exception happens inevitably, because the fd used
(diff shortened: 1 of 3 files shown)


totaam:
Thanks for the PR!

I'm going to let it simmer in master for a while before even contemplating a backport to the stable branches because I have this niggling feeling that this is going to cause problems somewhere unexpected. Like signals? xvfb?

jktomer:
That's an entirely reasonable precaution! For what it's worth, I tried it on two machines: one subject to the problem I described in the commit message, where the difference was "works with the change, not without it," and one not subject to that problem, where the set of open file descriptors reported by lsof after daemonizing was identical with or without the change.

Of course, that's a very small sample size. I wouldn't be shocked to discover someone is starting xpra in a way that does inherit some surprising open file descriptors; hopefully it's easier for them to adjust the way they start it (systemd-run *should* prevent any issues, for example, since IIRC it does take responsibility for setting up the entire set of inherited file descriptors for the child) than to put something like this back into xpra, but if we do need to, I'd suggest that after closing file descriptors we immediately self-exec, so that initialization code from library dependencies has a chance to open anything it needs to again.

totaam:
> I'd suggest that after closing file descriptors we immediately self-exec

Which ones?
We can only self-exec when daemonizing, in other cases the caller may be relying on the process keeping its pid.

PR #3430, opened 2022-01-14
Fix Colorspace table not correctly rendering

vcaen:
The missing line break after the list was preventing the table to be rendered


totaam:
Thanks!

PR #3471, opened 2022-02-21
Websocket auth

mdavidsaver:
This PR has had very little testing.  At this point I'm looking for comment.  Is this a good idea/approach?  Something which could be merged?  Is there already a way to accomplish this already which I've missed?

In looking into ways to integrate authentication of an xpra html5 client with a web application I came across the nginx idea of [sub-request authentication](https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-subrequest-authentication/) and wondered if this could be extended to the xpra server as well.  From what I can tell, the immediate answer is "no".  This PR is an attempt to change the situation by making a `WebSocketRequestHandler` (or `None`) available to authentication plugins.  This gives a plugin access to HTTP header information, especially [Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie) or [Authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization).

`exec_auth.py` is changed to pass some HTTP header information in the manner of a CGI script.  (informed by `http.server.CGIHTTPRequestHandler.run_cgi()`)  So that the command executed could eg. pass the requested user name and a session cookie back to a web application to authorize the request.

This is still a work in progress.  Having done so, I'm less sure that modifying `exec_auth.py` is the way to go.

### DIFF
--- a/xpra/scripts/server.py
+++ b/xpra/scripts/server.py
@@ -1339,7 +1339,7 @@ def _do_run_server(script_file, cmdline,
 
     if opts.chdir:
         log(f"chdir({opts.chdir})")
-        os.chdir(opts.chdir)
+        os.chdir(os.path.expanduser(opts.chdir))
 
     dbus_pid, dbus_env = 0, {}
     if not shadowing and POSIX and not OSX:
--- a/xpra/server/auth/exec_auth.py
+++ b/xpra/server/auth/exec_auth.py
@@ -45,7 +45,7 @@ class Authenticator(SysAuthenticator):
         super().__init__(**kwargs)
 
     def requires_challenge(self) -> bool:
-        return False
+        return bool(self.http_request)
 
     def authenticate(self, caps : typedict) -> bool:
         info = "Connection request from %s" % self.connection_str
--- a/xpra/server/auth/exec_auth.py
+++ b/xpra/server/auth/exec_auth.py
@@ -4,7 +4,9 @@
 # later version. See the file COPYING for details.
 
 import os
-from subprocess import Popen
+import base64, binascii
+from urllib.parse import urlparse
+from subprocess import Popen, DEVNULL
 from gi.repository import GLib
 
 from xpra.util import envint, typedict
@@ -12,6 +14,7 @@ from xpra.os_util import OSX
 from xpra.child_reaper import getChildReaper
 from xpra.server.auth.sys_auth_base import SysAuthenticator, log
 from xpra.platform.features import EXECUTABLE_EXTENSION
+from xpra.net.websockets.handler import WebSocketRequestHandler
 
 TIMEOUT = envint("XPRA_EXEC_AUTH_TIMEOUT", 600)
 
@@ -22,6 +25,7 @@ class Authenticator(SysAuthenticator):
         log("exec.Authenticator(%s)", kwargs)
         self.command = kwargs.pop("command", "")
         self.timeout = kwargs.pop("timeout", TIMEOUT)
+        self.http_request = kwargs.pop('http_request')
         self.timer = None
         self.proc = None
         self.timeout_event = False
@@ -50,7 +54,12 @@ class Authenticator(SysAuthenticator):
     def authenticate(self, caps : typedict) -> bool:
         info = "Connection request from %s" % self.connection_str
         cmd = [self.command, info, str(self.timeout)]
-        with Popen(cmd) as proc:
+        env = stdin = stdout = None
+        if self.http_request:
+            stdin = stdout = DEVNULL
+            env = os.environ.copy()
+            add_cgi_headers(env, self.http_request)
+        with Popen(cmd, env=env, stdin=stdin, stdout=stdout) as proc:
             self.proc = proc
             log("authenticate(..) Popen(%s)=%s", cmd, proc)
             #if required, make sure we kill the command when it times out:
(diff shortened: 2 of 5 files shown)


totaam:
Looks interesting, but like you said, I wouldn't modify `exec` directly, perhaps you can use a subclass instead?

totaam:
Please rebase.
I wouldn't pass the whole request object, maybe just add the headers if that's what you want to access.
And extend exec. And add error handling.

xpra/scripts/server.py
```diff
@@ -1339,7 +1339,7 @@ def check_xvfb(timeout=0):
 
     if opts.chdir:
         log(f"chdir({opts.chdir})")
-        os.chdir(opts.chdir)
+        os.chdir(os.path.expanduser(opts.chdir))
```
totaam, line 1342: Applied in 36fe7466dc8ffecd7e8a0b3939020986d6597285 using `osexpand` instead.

xpra/server/auth/exec_auth.py
```diff
@@ -45,12 +49,17 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
     def requires_challenge(self) -> bool:
-        return False
+        return bool(self.http_request)
```
totaam, line 52: Move this to a subclass that overrides the function.
If the `http_request` is missing then authentication should fail.

xpra/server/auth/exec_auth.py
```diff
@@ -45,12 +49,17 @@ def __init__(self, **kwargs):
...
     def authenticate(self, caps : typedict) -> bool:
         info = "Connection request from %s" % self.connection_str
         cmd = [self.command, info, str(self.timeout)]
-        with Popen(cmd) as proc:
+        env = stdin = stdout = None
+        if self.http_request:
+            stdin = stdout = DEVNULL
+            env = os.environ.copy()
+            add_cgi_headers(env, self.http_request)
+        with Popen(cmd, env=env, stdin=stdin, stdout=stdout) as proc:
```
totaam, line 62: None of this belongs in `exec_auth`, please subclass instead.

xpra/server/auth/exec_auth.py
```diff
@@ -87,3 +96,54 @@ def command_timedout(self):
...
+    env['SERVER_SOFTWARE'] = 'XPRA'
+    env['SERVER_PROTOCOL'] = request.protocol_version
+    env['GATEWAY_INTERFACE'] = 'CGI/1.1'
+    env['REQUEST_METHOD'] = request.command
+    env['PATH_INFO'] = parts[0]
+    env['PATH_TRANSLATED'] = parts[0] # TODO: window-ish path?
+    env['SCRIPT_NAME'] = parts[1]
+    if url.query:
+        env['QUERY_STRING'] = url.query
+    env['REMOTE_ADDR'] = request.client_address[0]
```
totaam, line 121: Perhaps some of these values should be sanitized before being reflected back to the client.

xpra/server/rfb/rfb_server.py
```diff
@@ -68,7 +68,7 @@ def handle_rfb_connection(self, conn, data=b""):
             conn.close()
             return
         def rfb_protocol_class(conn):
-            auths = self.make_authenticators("rfb", "rfb", conn)
+            auths = self.make_authenticators("rfb", "rfb", conn, None)
```
totaam, line 71: No change needed here, just make the argument optional.

xpra/server/server_core.py
```diff
@@ -1831,8 +1832,8 @@ def _process_hello(self, proto, packet):
         #in which case we'll end up here parsing the hello again
         start_thread(self.verify_auth, "authenticate connection", daemon=True, args=(proto, packet, c))
 
-    def make_authenticators(self, socktype, remote, conn):
-        authlog("make_authenticators%s socket options=%s", (socktype, remote, conn), conn.options)
+    def make_authenticators(self, socktype, remote, conn, http_request):
```
totaam, line 1835: Use `http_request=None` default argument.

xpra/server/server_core.py
```diff
@@ -1901,7 +1903,7 @@ def auth_failed(msg):
         if not proto.authenticators:
             socktype = conn.socktype_wrapped
             try:
-                proto.authenticators = self.make_authenticators(socktype, remote, conn)
+                proto.authenticators = self.make_authenticators(socktype, remote, conn, getattr(proto, 'http_request', None))
```
totaam, line 1906: Using `getattr` here is a bit ugly.
Why not stash the `http_request` object in the connection's options instead?
Then the `make_authenticators` function doesn't need any changes and the authenticator can just extract the `http_request` from the connection?

mdavidsaver:
@totaam I hope you did not spent too much time puzzling over my changes to `exec_auth.py`.  I should have mentioned earlier.  Based on your previous comment https://github.com/Xpra-org/xpra/pull/3471#issuecomment-1047013474 I am planning to redo this PR as a new module.  Provisionally `subreq_auth.py`.  This will make HTTP requests directly, instead of relying on an external script/command.

xpra/server/server_core.py
mdavidsaver: > Why not stash the `http_request` object in the connection's options instead?

No good reason.  Mainly my unfamiliarity with this code.  Will `conn` allways be a sub-class of `xpra.net.bytestreams.Connection`?  Are there any namespace consideration to avoid colliding with some other piece of code?

It looks like you have dropped python 2.x support.  Are you now accepting addition of `typing` annotations?  I think a few strategically placed annotations could make this code considerably easier to follow.

totaam: `conn` will always either be a subclass or use the same structure, as for `options` although collisions are technically possible and not actively prevented, there is no sane reason for re-using the `http_request` name for something else.

We have already added a lot of `typing` annotations since python<3.6 was dropped, adding more would be most welcome.

totaam:
> Provisionally `subreq_auth.py`.

`sub` stands for? Perhaps `http_headers_auth.py` or something clearer?

mdavidsaver:
> `sub` stands for?

I got the idea for this for this PR from the NGINX "sub-request" feature.  I guess "sub" in that while handling one HTTP request, a second/subordinate HTTP request is made?

> Perhaps `http_headers_auth.py` or something clearer?

Ok.  `http_headers_auth.py` it is.

xpra/server/server_core.py
mdavidsaver: > ... there is no sane reason for re-using the http_request name for something else.

I should have phrased my question differently.  I am not certain who has control over key names in the `conn.options` dict.  I think your reply implied that only xpra code does.  (as opposed to server configuration, or perhaps even remote clients)

I'll switch to using `conn.options['http_request']`.

totaam:
Not heard back. :disappointed:

PR #3486, opened 2022-03-17
Added keycloak authentication

### LINKED ISSUE #3334: keycloak authentication module
Probably using https://github.com/marcospereirampj/python-keycloak

First hurdle is the keycloak server setup, which is done manually using a UI rather than scripting.
Then we have to choose how deeply to integrate. Do we want to use `get_sessions()`?
The documentation is quite limited too.

iDmple:
Fixes #3334

I added keycloak authentication as discussed. This is needed for it:
```
pip3 install requests==2.25.1 urllib3==1.26.5 oauthlib python-keycloak
```
These specific versions are due to [this issue](https://github.com/marcospereirampj/python-keycloak/issues/196).

Regarding how the XPRA auth is working, I'm unsure how to display the error messages in the client when the authentication fails in the function `check`, but it displays fine when it fails in `__init__`.

Please check this PR together with the associated PR in [xpra-html5](https://github.com/Xpra-org/xpra-html5/pull/164). Let me know if anything needs to be improved.


xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,113 @@
...
+        if "keycloak" not in digests:
+            log.error("Error: client does not support keycloak authentication")
+            return None
+        self.challenge_sent = True
+        return self.salt, "keycloak"
+
+    def check(self, response_json) -> bool:
+        assert self.challenge_sent
+        
+        if response_json is None or response_json == "":
```
totaam, line 66: Use `if not response_json`.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,113 @@
...
+          log.error("keycloak authentication failed: invalid response received from authorization endpoint.")
+          return False
+        
+        #log("response_json: %s", repr(response_json))
+        response = json.loads(response_json)
+
+        if type(response) != dict or ("code" not in response and "error" not in response):
```
totaam, line 73: Use `isinstance` here.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,113 @@
...
+          log.error("keycloak authentication failed: invalid response received from authorization endpoint.")
```
totaam, line 74: Errors are usually prefixed with `Error: `.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,113 @@
...
+          return False
+
+        log("check(%s)", repr(response))
+
+        if "error" in response:
+          log.error("keycloak authentication failed with error %s: %s", response["error"], response["error_description"])
```
totaam, line 80: As above.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,113 @@
...
+    def check(self, response_json) -> bool:
+        assert self.challenge_sent
+        
+        if response_json is None or response_json == "":
+          log.error("keycloak authentication failed: invalid response received from authorization endpoint.")
+          return False
+        
+        #log("response_json: %s", repr(response_json))
+        response = json.loads(response_json)
+
```
totaam, line 72: Since you are accessing this multiple times, it would make sense to do:
```
code = response.get("code")
error = response.get("error")
if code is not None:
    ...
```

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,113 @@
...
+            #log("token_info: %s", repr(token_info))
+
+            if token_info["active"]:
+              # Get userinfo
+              #user_info = keycloak_openid.userinfo(token['access_token'])
+              #log("userinfo_info: %s", repr(user_info))
+
+              log("keycloak authentication succeeded: token is active")
+            else:
+              log.error("keycloak authentication failed: token is not active")
```
totaam, line 109: Maybe `return False` explicitly here is safer / better.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,113 @@
...
+          # Configure client
+          keycloak_openid = KeycloakOpenID(server_url=KEYCLOAK_SERVER_URL,
+                            client_id=KEYCLOAK_CLIENT_ID,
+                            realm_name=KEYCLOAK_REALM_NAME,
+                            client_secret_key=KEYCLOAK_CLIENT_SECRET_KEY)
+
+          try:
+            # Get well_known
+            #config_well_know = keycloak_openid.well_know()
+            #log("well_known: %s", repr(config_well_know))
```
totaam, line 93: Is there any harm in logging these details?
If not, then uncomment that line.
Don't use `%s` with `repr`, just use `%r`.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,113 @@
...
+        
+            # Get token
+            token = keycloak_openid.token(code=response["code"], grant_type=[KEYCLOAK_GRANT_TYPE], redirect_uri=KEYCLOAK_REDIRECT_URI)
+
+            # Verify token
+            token_info = keycloak_openid.introspect(token['access_token'])
+            #log("token_info: %s", repr(token_info))
```
totaam, line 100: As above, logging this might be useful for debugging.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,113 @@
...
+
+            if token_info["active"]:
+              # Get userinfo
+              #user_info = keycloak_openid.userinfo(token['access_token'])
+              #log("userinfo_info: %s", repr(user_info))
```
totaam, line 105: As above.

totaam (changes requested):
This implements keycloak auth purely on the server so this is not safe at all over unencrypted connections.
All the client does is to send the password unencrypted. That's not how keycloak is meant to be used.

I could merge this module as `keycloak_password` or something - similar to the `kerberos_password` we already have.
But not the client changes.

A proper `keycloak` module would do the request from the client and send the token to the server only.

iDmple:
I think there is a misunderstanding, this does not implement the `grant_type` `password`. This implements `authorization_code`. There is no password involved at all!

If you want me to request the token in the client, and then validate it in the server, that's possible, but I'm not sure what this changes security-wise.

totaam:
> There is no password involved at all!

There definitely is:
https://github.com/Xpra-org/xpra/pull/3486/files#diff-1a9bde312dd0af09b7dfac24f0b54286d45366435e1a1a8614f2981f46619398

> I'm not sure what this changes security-wise.

It means that with your authentication module, the password is not sent over the network - often unencrypted.

iDmple:
What is sent through that is the authorization code, this isn't a password. I will improve this file to use another variable name instead.

xpra/server/auth/keycloak_auth.py
iDmple: There is no harm in more logging this, but this does add a useless request making the authentication slower. I uncommented, feel free to comment again.

xpra/server/auth/keycloak_auth.py
iDmple: It does log some sensitive information that can identify the user, so I would have prefered not to log this. It is also extremely verbose. I uncommented, feel free to comment again.

xpra/server/auth/keycloak_auth.py
iDmple: As above.

iDmple:
I'm not sure what the handler file was for. I seem to have needed it at some point, but in the end, it isn't used at all so I removed it.

I hope these changes address your concerns.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,125 @@
...
+          response = json.loads(response_json)
+        except Exception:
+          log.error("Error: keycloak authentication failed as invalid response was received from authorization endpoint.")
+          return False 
+
+        if not isinstance(response, dict):
+          log.error("Error: keycloak authentication failed as invalid response was received from authorization endpoint.")
+          return False
+
+        log("check(%r)", repr(response))
```
totaam, line 81: `%r` means that `repr` is unnecessary.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,125 @@
...
+        #log("response_json: %r", repr(response_json))
+        
+        if not response_json:
+          log.error("Error: keycloak authentication failed as invalid response was received from authorization endpoint.")
+          return False
+        
+        try:
+          response = json.loads(response_json)
+        except Exception:
+          log.error("Error: keycloak authentication failed as invalid response was received from authorization endpoint.")
```
totaam, line 74: Please add:
```
log("json failed to load %r", response_json, exc_info=True)
```
So that the full details of the error can be shown when authentication debugging is enabled.
And perhaps split the long error message into two lines.

The exception type you should be catching here is [json.JSONDecodeError](https://docs.python.org/3/library/json.html#json.JSONDecodeError), not `Exception`.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,125 @@
...
+        auth_code = response.get("code")
+        error = response.get("error")
+
+        if not error and not auth_code:
+          log.error("Error: keycloak authentication failed as invalid response was received from authorization endpoint.")
+          return False
+
+        if error:
+          log.error("Error: keycloak authentication failed with %s: %s", error, response.get("error_description"))
+          return False
```
totaam, line 92: Easier and more readable to do:
```
if error:
   ...
   return False
if not auth_code:
   ...
   return False
```

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,125 @@
...
+          try:
+            # Get well_known
+            config_well_know = keycloak_openid.well_know()
+            log("well_known: %r", repr(config_well_know))
+        
+            # Get token
+            token = keycloak_openid.token(code=auth_code, grant_type=[KEYCLOAK_GRANT_TYPE], redirect_uri=KEYCLOAK_REDIRECT_URI)
+
+            # Verify token
+            token_info = keycloak_openid.introspect(token['access_token'])
```
totaam, line 110: Please use double quotes here like everywhere else.
Could `access_token` be missing from `token`? (which would cause an `KeyError`)
If so:
```
access_token = token.get("access_token")
if not access_token:
   log.error("Error: cannot authenticate, access token is missing")
   return False
```

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,125 @@
...
+            log("token_info: %r", repr(token_info))
```
totaam, line 111: As above, `repr` is redundant.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,125 @@
...
+
+            if token_info["active"]:
```
totaam, line 113: As above: can `active` be missing here?

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,125 @@
...
+              # Get userinfo
+              user_info = keycloak_openid.userinfo(token['access_token'])
```
totaam, line 115: As above: lookup `access_token` just once.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,125 @@
...
+              log("userinfo_info: %r", repr(user_info))
+
+              log("keycloak authentication succeeded: token is active")
+              return True
+            else:
```
totaam, line 120: `else` is unnecessary after `return`.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,125 @@
...
+              log.error("Error: keycloak authentication failed as token is not active")
+              return False
+          except Exception as e:
+            log.error("Error: keycloak authentication failed with code %s: %s", e.response_code, e.error_message)
```
totaam, line 124: You are catching a generic `Exception` type so you cannot access `response_code` and `error_message` on it.
Catch a more specific type that is known to have these attributes. The generic `Exception` type should be caught by the caller.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def __init__(self, **kwargs):
+        self.server_url = kwargs.pop("server_url", KEYCLOAK_SERVER_URL)
+        self.realm_name = kwargs.pop("realm_name", KEYCLOAK_REALM_NAME)
+        self.client_id = kwargs.pop("client_id", KEYCLOAK_CLIENT_ID)
+        self.client_secret_key = kwargs.pop("client_secret_key", KEYCLOAK_CLIENT_SECRET_KEY)
+        self.redirect_uri = kwargs.pop("redirect_uri", KEYCLOAK_REDIRECT_URI)
+        self.scope = kwargs.pop("scope", KEYCLOAK_SCOPE)
+        self.grant_type = kwargs.pop("grant_type", KEYCLOAK_GRANT_TYPE)
+        kwargs["prompt"] = kwargs.pop("prompt", "keycloak")
```
totaam, line 34: It wouldn't hurt to add a comment here, like "_use `keycloak` as default prompt_".

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,125 @@
...
+          return False
+        
+        try:
+          response = json.loads(response_json)
+        except Exception:
+          log.error("Error: keycloak authentication failed as invalid response was received from authorization endpoint.")
+          return False 
+
+        if not isinstance(response, dict):
+          log.error("Error: keycloak authentication failed as invalid response was received from authorization endpoint.")
```
totaam, line 78: As above: full details at debug level, split the long line and please log the `response` datatype we actually received to help debugging.

xpra/client/client_base.py
```diff
@@ -204,7 +204,7 @@ def init_challenge_handlers(self, challenge_handlers):
                 items = (
                     "uri", "file", "env",
                     "kerberos", "gss",
-                    "u2f",
+                    "keycloak", "u2f",
```
totaam, line 207: You define a `keycloak` challenge handler, but you are not including it any more?
How is this going to work?

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,125 @@
...
+    def __repr__(self):
+        return "keycloak"
+
+    def get_challenge(self, digests):
+        assert not self.challenge_sent
+        if "keycloak" not in digests:
+            log.error("Error: client does not support keycloak authentication")
+            return None
+        self.challenge_sent = True
+        return self.salt, "keycloak"
```
totaam, line 61: Shouldn't this be `password` since there is no `keycloak` challenge handler?

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,125 @@
...
+# Xpra is released under the terms of the GNU GPL v2, or, at your option, any
+# later version. See the file COPYING for details.
+
+import os
+import json
+
+from xpra.util import typedict
+from xpra.server.auth.sys_auth_base import SysAuthenticator, log
+from keycloak import KeycloakOpenID
+from oauthlib.oauth2 import WebApplicationClient
```
totaam, line 13: Please move these two imports into the methods that uses them so that the module can be loaded and will only cause an `ImportError` if we actually try to use it and the libraries are missing.

totaam:
Please don't be put off by the number of comments above, it's all fairly minor at this point.

A few more questions about testing:
* how do I go about testing this - preferably in an automated kind of way? (either mocking it, or using automated instructions for setting up a test keycloak instance)
* can you add a trivial command line interface to run a simple authentication test? Something similar to:
https://github.com/Xpra-org/xpra/blob/c80e56c7d4250420fe2b63f5f9dc287579d06a79/xpra/server/auth/pam_auth.py#L61-L75

totaam (changes requested):

xpra/server/auth/keycloak_auth.py
iDmple: Do you want me to remove the `log.error()` and replace by 
```
log("json failed to load %r", response_json, exc_info=True)
```
, or keep both of them?

xpra/server/auth/keycloak_auth.py
iDmple: I don't think it could be missing, but let's do this for extra safety.

xpra/server/auth/keycloak_auth.py
iDmple: Same as above

xpra/server/auth/keycloak_auth.py
totaam: Keep both of them.
`log.error` is for the more generic error message, `log` (which is an alias for `log.debug`) is only shown when debugging is enabled.

totaam:
Bump!

iDmple:
Hey, the work is almost done but I'm currently on holidays. Will finish it up next week!

xpra/server/auth/keycloak_auth.py
iDmple: No, this is correct. We are using this in the client. There is no password.

iDmple:
> Please don't be put off by the number of comments below, it's all fairly minor at this point.
> 
> A few more questions about testing:
> 
> * how do I go about testing this - preferably in an automated kind of way? (either mocking it, or using automated instructions for setting up a test keycloak instance)
> * can you add a trivial command line interface to run a simple authentication test? Something similar to:
>   https://github.com/Xpra-org/xpra/blob/c80e56c7d4250420fe2b63f5f9dc287579d06a79/xpra/server/auth/pam_auth.py#L61-L75

About testing this I'm not sure. I'm not running my own keycloak instance. Maybe we will do so in the future.
I will try to add a command line interface. This is the last missing part.

totaam:
You've "resolved" the review without actually addressing any of the numerous comments I made.
Am I missing something? Did you forget to push your changes?

iDmple:
I'm going to push them soon, it was just easier to understand what I addressed already that way.

I have a question. I would like to implement the logout functionality. Do you have any pointers or examples on how to do that?

totaam:
> I have a question. I would like to implement the logout functionality.

Good point. This wasn't needed until now.
I'll see if we can add a session hook.

iDmple:
> > I have a question. I would like to implement the logout functionality.
> 
> Good point. This wasn't needed until now. I'll see if we can add a session hook.

It's not urgent, just nice to have! Thanks!

iDmple:
> Please don't be put off by the number of comments below, it's all fairly minor at this point.
> 
> A few more questions about testing:
> 
> * how do I go about testing this - preferably in an automated kind of way? (either mocking it, or using automated instructions for setting up a test keycloak instance)
> * can you add a trivial command line interface to run a simple authentication test? Something similar to:
>   https://github.com/Xpra-org/xpra/blob/c80e56c7d4250420fe2b63f5f9dc287579d06a79/xpra/server/auth/pam_auth.py#L61-L75

I added the changes we discussed about.

In order to test you can call it that way:
```
python3 keycloak_auth.py '{"code":"authorization_code"}'
```
The `authorization_code` must be requested from the keycloak auth endpoint. I hope it's suitable.

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,185 @@
...
+        error = response.get("error")
+
+        if error:
+          log.error("Error: keycloak authentication failed")
+          log.error("%s: %s", error, response.get("error_description"))
+          return False
+
+        if not auth_code:
+          log.error("Error: keycloak authentication failed")
+          log.error("Invalid response received from authorization endpoint")
```
totaam, line 109: missing `return False`

xpra/server/auth/keycloak_auth.py
```diff
@@ -0,0 +1,185 @@
...
+          return False 
+
+        if not isinstance(response, dict):
+          log.error("Error: keycloak authentication failed")
+          log.error("Invalid response received from authorization endpoint")
+          log("response is of type %r but dict type is required", type(response), exc_info=True)
+          log("failed to load response %r", response, exc_info=True)
+          return False
+
+        log("check(%r)", response, exc_info=True)
```
totaam, line 97: Don't use `exc_info=True` if there is no exception to use.

xpra/server/auth/keycloak_auth.py
totaam: Just thinking out loud here now that I have figured out what I meant here.

---

The `token` is the "password" in this context (client auth modules) and it is sent unencrypted (see the change to `xpra/net/digest.py`). That's because the token is not meant to be re-usable from another client - how that's enforced is the responsibility of `keycloak`.

What bothers me here is not new to this authentication module so this is not a showstopper: the client will be prompting the user for a "keycloak" (and I think it should say `keycloak token` instead) and then sending it unencrypted without telling the user that this is the case, which could end up leaking a real password the user types in accidentally. (yes, users will do stupid things - I do too)

The `kerberos` and `gss` client handlers acquire the token themselves:
https://github.com/Xpra-org/xpra/blob/137a222006ce28cd18d8def7884e9d6aeb89b38b/xpra/client/auth/kerberos_handler.py#L76
https://github.com/Xpra-org/xpra/blob/137a222006ce28cd18d8def7884e9d6aeb89b38b/xpra/client/auth/gss_handler.py#L54
And I was hoping that a `keycloak` client authentication module could do the same thing.
But the problem with all those modules is that when they fail, we end up also calling the `prompt` client auth handler as fallback..

The `prompt` handler even has an ugly default prompt specific for `kerberos` and `gss`:
https://github.com/Xpra-org/xpra/blob/137a222006ce28cd18d8def7884e9d6aeb89b38b/xpra/client/auth/prompt_handler.py#L24-L25

I am just thinking that it should be more obvious to the user that what they type in the prompt handler can be intercepted in this case.
(and obviously skip this warning if the connection to the server is actually encrypted with `ssl`, `wss`, `ssh` or `aes`)

xpra/server/auth/keycloak_auth.py
totaam: minor: `keycloak token` as default prompt would be clearer IMO.

xpra/net/digest.py
```diff
@@ -67,7 +67,7 @@ def gendigest(digest, password, salt):
         salt = salt.ljust(16, b"\x00")[:16]
         v = generate_response(password, salt)
         return hexstr(v)
-    if digest in ("xor", "kerberos", "gss"):
+    if digest in ("xor", "kerberos", "gss", "keycloak"):
         #kerberos and gss use xor because we need to use the actual token
```
totaam, line 71: `kerberos, gss and keycloak`

totaam (commented):
> `python3 keycloak_auth.py '{"code":"authorization_code"}'`
> The authorization_code must be requested from the keycloak auth endpoint. I hope it's suitable.

It is fine.
A couple of questions:
* if a valid response is just a JSON string, wouldn't a keycloak auth client be trivial to implement?
* do you have sample responses we could add to a unit test? (which is missing BTW)

---

There are just a few nitpicks more that I just found in the PR.
If you don't want to take care of them, just let me know and I'll tweak things after merging this PR.

iDmple:
> > `python3 keycloak_auth.py '{"code":"authorization_code"}'`
> > The authorization_code must be requested from the keycloak auth endpoint. I hope it's suitable.
> 
> It is fine. A couple of questions:
> 
> * if a valid response is just a JSON string, wouldn't a keycloak auth client be trivial to implement?
> * do you have sample responses we could add to a unit test? (which is missing BTW)
> 
> There are just a few nitpicks more that I just found in the PR. If you don't want to take care of them, just let me know and I'll tweak things after merging this PR.

How the client works, if you look at the diagram I sent above, is indeed very simple. We redirect to the `auth` endpoint, the user logs into keycloak. If successful we are redirected back and an authorization code is returned as a url parameter which we pass as a json. Then in the server (this PR), we use this authorization code to get a token on the `token` endpoint, then validate the token using the `introspect` endpoint.

This is exactly what I implemented in the [client PR](https://github.com/Xpra-org/xpra-html5/pull/164). Could you please review it while I take care of the last details here?

Regarding the unit tests, there is only one value that we can give, which is just a random string like above which is going to fail the auth, unless you have access to a keycloak server and can get a valid authorization code during testing.

totaam:
> Regarding the unit tests, there is only one value that we can give, which is just a random string like above which is going to fail the auth, unless you have access to a keycloak server and can get a valid authorization code during testing.

Is the JSON string always just one `code` attribute and nothing else?
Aren't there other attributes that can be echanged?

As for the unit tests, even testing that authentication fails will at least exercise some of the code and increase code coverage.
Feed it:
* invalid JSON string, None value, etc..
* valid JSON string with garbage in it
* valid JSON string with a `code` value (but either empty, a number, a float, etc..)
* valid JSON string with a string `code`

Ideally we would mock a keycloak server to validate fully - but that's too much effort.

iDmple:
> The token is the "password" in this context (client auth modules) and it is sent unencrypted (see the change to xpra/net/digest.py). That's because the token is not meant to be re-usable from another client - how that's enforced is the responsibility of keycloak.

In this case, this isn't the token that is sent unencrypted (it still uses SSL), it's the authorization code. 

> What bothers me here is not new to this authentication module so this is not a showstopper: the client will be prompting the user for a "keycloak" (and I think it should say keycloak token instead) and then sending it unencrypted without telling the user that this is the case, which could end up leaking a real password the user types in accidentally. (yes, users will do stupid things - I do too)

The client isn't prompting anything to the user. The client is redirecting to an URL to get the authorization code. I'm not sure in which conditions we would be able to see this prompt? Even if that happened, what it should prompt the user would be the authorization code, not the token. Keycloak has several `grant_types` and if we were to implement the `password` `grant_type` then we would be in the situation you mention, yes. However this isn't how this works with the `authorization_code` `grant_type`, please see diagram and explanation above.

> The kerberos and gss client handlers acquire the token themselves:
> And I was hoping that a keycloak client authentication module could do the same thing.
Indeed the keycloak authentication module does also aquire the token themselves:
https://github.com/Xpra-org/xpra/blob/3a3e74d4cd48e465c7c3ac56ccd7d0272342c78c/xpra/server/auth/keycloak_auth.py#L125-L126

The way that this authentication module works is not the same as gss/kerberos.

I hope I could clarify.

iDmple:
> > Regarding the unit tests, there is only one value that we can give, which is just a random string like above which is going to fail the auth, unless you have access to a keycloak server and can get a valid authorization code during testing.
> 
> Is the JSON string always just one `code` attribute and nothing else? Aren't there other attributes that can be echanged?
> 
> As for the unit tests, even testing that authentication fails will at least exercise some of the code and increase code coverage. Feed it:
> 
> * invalid JSON string, None value, etc..
> * valid JSON string with garbage in it
> * valid JSON string with a `code` value (but either empty, a number, a float, etc..)
> * valid JSON string with a string `code`
> 
> Ideally we would mock a keycloak server to validate fully - but that's too much effort.

Actually it can also have an error inside. Let me send you an example in a minute.

xpra/server/auth/keycloak_auth.py
iDmple: When we add more grant_types we could have

- keycloak password
- keycloak authorization code
- others

But never token. We could also implement them in different files if too different / complex.

We can add what you prefer here but in this case it will never be shown, or I don't see how. And it will never ask for a token.

totaam:
Stumbled upon this good overview of oauth2: [The complete guide to protecting your APIs with OAuth2 - part 1](https://stackoverflow.blog/2022/04/11/the-complete-guide-to-protecting-your-apis-with-oauth2/)

totaam:
@iDmple please take a look at the obvious fixes from the unit test I have added - since I don't have a keycloak server to test against.
**This** is why we needed unit tests to merge this code.

totaam:
@iDmple can you please try b0834efd279e6574e082ff553b50f3e348920101 or later?
This changes the return value of the `get_challenge` method to match the strict signature used by all the other authentication modules where the "salt" (which is a URI for keycloak) is now a byte string.
https://github.com/Xpra-org/xpra/commit/b0834efd279e6574e082ff553b50f3e348920101#diff-49b8e5880c4ca08664dd75e76c2b961e96ba3aef497cafaae5d133389640ab1a

iDmple:
@totaam We will test this ASAP and report back.

iDmple:
@totaam Sorry it took forever. We've tried it and it works perfectly. Thanks a lot!

PR #3507, opened 2022-03-30
add path-check, add XPRA_SYSCONF_DIR for xposix, add XDG_DATA_DIRS/icons to search path for pixmaps

jhgoebbert:
1) The line
```
LOAD_APPLICATIONS = os.environ.get("XPRA_MENU_LOAD_APPLICATIONS", "%s/share/applications" % sys.prefix).split(":")
```
https://github.com/Xpra-org/xpra/blob/master/xpra/platform/xposix/menu_helper.py#L35
can return `None` and therefor needs to be checked for existance before use.

2) Like it is already possible for OSX ( https://github.com/Xpra-org/xpra/blob/master/xpra/platform/darwin/paths.py#L93 )
 it would help if the environment variable XPRA_SYSCONF_DIR can be used to overwrite the system default path.
 
 Example:
We have multiple installations of Xpra on the same multi-user systems of a few hpc-clusters. The user/workflow can choose which of these installations is loaded using lmod, a framework to load software when needed into the current environment.  
Xpra is therefore no real system-installations as they are installed to special directories each, and they are no user-installations as they are not installed by the user.
Hence it is needed to avoid that xpra looks at system paths (`/etc/xpra`) for configuration files. 

3) XDG_DATA_DIRS/icons should added to the search paths for icons
https://specifications.freedesktop.org/icon-theme-spec/latest/ar01s03.html

### DIFF
--- a/xpra/platform/xposix/menu_helper.py
+++ b/xpra/platform/xposix/menu_helper.py
@@ -177,8 +177,9 @@ def find_resources_icon(*names):
 def find_pixmap_icon(*names):
     if not LOAD_FROM_PIXMAPS:
         return None
+    pixmaps_dirs = [d + '/icons' for d in os.environ.get("XDG_DATA_DIRS").split(":")]
     pixmaps_dir = "%s/share/pixmaps" % sys.prefix
-    pixmaps_dirs = (pixmaps_dir, os.path.join(pixmaps_dir, "comps"))
+    pixmaps_dirs += (pixmaps_dir, os.path.join(pixmaps_dir, "comps"))
     for d in pixmaps_dirs:
         if not os.path.exists(d) or not os.path.isdir(d):
             return None
--- a/xpra/platform/xposix/menu_helper.py
+++ b/xpra/platform/xposix/menu_helper.py
@@ -411,6 +411,8 @@ def load_xdg_menu_data():
         from xdg.Menu import MenuEntry
         entries = {}
         for d in LOAD_APPLICATIONS:
+            if not os.path.exists(d):
+                continue
             for f in os.listdir(d):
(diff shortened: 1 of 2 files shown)


totaam:
> can return None and therefor needs to be checked for existance before use

No. This cannot return a None value.

> Hence it is needed to avoid that xpra looks at system paths

What you want is `XPRA_SYSTEM_CONF_DIRS`:
https://github.com/Xpra-org/xpra/blob/5f5fc0ee55b1e532fc2760841a58eb319226319e/xpra/platform/paths.py#L41-L45

jhgoebbert:
> No. This cannot return a None value.

Yes, you are right. But the check is for another reason required as it might exist:
```
Python 3.9.6 (default, Nov 16 2021, 12:25:13)
[GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os, sys
>>> os.environ.get("XPRA_MENU_LOAD_APPLICATIONS", "%s/share/applications" % sys.prefix).split(":")
['/p/software/juwelsbooster/stages/2022/software/Python/3.9.6-GCCcore-11.2.0/share/applications']
>>>
[goebbert1@jwlogin23 juwels]$ ls /p/software/juwelsbooster/stages/2022/software/Python/3.9.6-GCCcore-11.2.0/share/applications
ls: cannot access '/p/software/juwelsbooster/stages/2022/software/Python/3.9.6-GCCcore-11.2.0/share/applications': No such file or directory
```

jhgoebbert:
I will check if `XPRA_SYSTEM_CONF_DIRS` solves (2) and create a new PR for (1) and (2) then.

Issue #3414, opened 2022-01-08
Error when uploading file over websocket connection


ParanoidVoxel:
**Describe the bug**
When dragging and dropping a file in the xpra web-ui, the filetransfer fails immediately on the server side with the following error:
```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/xpra/server/server_base.py", line 906, in process_packet
    call_handler()
  File "/usr/lib/python3/dist-packages/xpra/server/server_base.py", line 896, in call_handler
    handler(proto, packet)
  File "/usr/lib/python3/dist-packages/xpra/server/mixins/fileprint_server.py", line 250, in _process_send_file
    ss._process_send_file(packet)
  File "/usr/lib/python3/dist-packages/xpra/net/file_transfer.py", line 496, in _process_send_file
    os.write(fd, file_data)
```
**To Reproduce**
1. Start xpra with: `xpra start-desktop :80 --start=/opt/workspace/run_emwm.sh --webcam=no --exit-with-children=no --no-daemon --speaker=on --video-encoders=x264 --speaker-codec=wav,opus+mka --min-speed=75` (`run_emwm.sh` is a script that just launches my window manager)
2. Connect to the xpra web-ui running on the same server
3. Drag and drop file over browser window

**System Information (please complete the following information):**
 - Server OS: Ubuntu Bullseye (Docker)
 - Client OS: Web-UI running in the same container, accessed with chrome and firefox, both on w10 (same issue with both)
 - Xpra Server Version: v4.3.1-r0
 - Xpra Web-Client Version: 4.5.2

**Additional context**
This issue doesn't affect the "full" xpra client connecting with TCP. Could be an issue with only web-socket clients?

However, after implementing this dirty quickfix in the xrpa server, the issue is fixed in the web client, and file upload from the full client (using tcp) isn't broken by this change.

```diff
--- /usr/lib/python3/dist-packages/xpra/net/file_transfer.py    2022-01-07 17:08:03.779462400 +0000
+++ ./file_transfer.py  2022-01-07 17:07:55.389462400 +0000
@@ -491,6 +491,7 @@
         check_digest("sha256", hashlib.sha256)
         check_digest("sha1", hashlib.sha1)
         check_digest("md5", hashlib.md5)
+        file_data = file_data.encode("latin")
         try:
             os.write(fd, file_data)
         finally:
```

[...]

PR #3510, opened 2022-03-31
fix file upload

seidler2547:
file uploads using websocket (HTML5) client always produces a 0 byte file because file_data is a str but os.write needs bytes

### DIFF
--- a/xpra/net/file_transfer.py
+++ b/xpra/net/file_transfer.py
@@ -493,7 +493,7 @@ class FileTransferHandler(FileTransferAttributes):
         check_digest("sha1", hashlib.sha1)
         check_digest("md5", hashlib.md5)
         try:
-            os.write(fd, file_data)
+            os.write(fd, strtobytes(file_data))
         finally:
             os.close(fd)
         self.transfer_progress_update(False, send_id, monotonic()-start, filesize, filesize, None)


totaam:
This should not be the case with current versions.

Please specify what versions you are using.
See https://github.com/Xpra-org/xpra/wiki/Reporting-Bugs

seidler2547:
> This should not be the case with current versions.
> 
> Please specify what versions you are using. See https://github.com/Xpra-org/xpra/wiki/Reporting-Bugs

Are you referring to #3414 ? I am running 4.3.2, so that is already in. But my impression was that it only fixes the other direction, download files from server.

seidler2547:
> > I get an error when I use the floating menu to upload files. Using Firefox on Linux.
> 
> Please specify what version you are using. The current version of the html5 client does not have this problem: https://github.com/Xpra-org/xpra-html5

I wrote above using 4.3.2 for the server. And 4.5.2 for the client. These seem to be the latest released versions. I compared the server code and this file at least has not changed since then.

totaam:
Sorry, I couldn't see the client version.
This problem does not occur for me with the latest builds, please try the beta channel.

totaam:
Not heard back, should work with current versions.

PR #3511, opened 2022-03-31
add XDG_DATA_DIRS/icons to search path for pixmaps

jhgoebbert:
This PR adds `$XDG_DATA_DIRS/icons` to the search paths for icons  
as mentioned in the XDG documentation here https://specifications.freedesktop.org/icon-theme-spec/latest/ar01s03.html


totaam:
This caused a `None` value error and is fixed in beeb22d01b478e851636747102a364101465adf4.

PR #3653, opened 2022-10-14
Allow --max-size and --min-size to work again

SethRobertson:
In commit f4ad1959 the old format-style printing was changes to using join, except join only joins strings and parse_window_size returns ints.

map the output of parse_window_size to a string to allow join to work.


totaam:
Thanks!

PR #3665, opened 2022-10-21
Raise clarity: xpra install on server&client

tux2bsd:


tux2bsd:
Also, on the main website this most vital part is written as an afterthought.

` This is similar to ssh -X, except that performance will be noticeably better and you can disconnect and re-connect to this session as many times as you like. (obviously xpra must be installed on your server host) `

Remove: `(obviously xpra must be installed on your server host)`

And put `xpra must be installed on your server host` as the FIRST line under "Get started with Xpra".  

This is literally step one so shouldn't be written as an afterthought.

README.md
```diff
@@ -44,16 +44,19 @@ For more details, see [building from source](https://github.com/Xpra-org/xpra/tr
 ---
 
 # Usage
+## Initial requirements
+`xpra` must be installed on the client and the host (version matters).
```
totaam, line 48: That's not quite true, you can use the html5 client in which case only the server is required...

tux2bsd: Try this then:

`xpra` must be installed on the client and the host (version matters).  

You can use the html5 client in which case xpra is only required on the host.  People that are using the html5 client would already be somewhat experienced xpra users.

tux2bsd: I've done another commit.  Just using the github interface which is a bit shit, was only supposed to be a quick adjustment.

README.md
```diff
@@ -44,16 +44,21 @@ For more details, see [building from source](https://github.com/Xpra-org/xpra/tr
...
```
totaam, line 48: Please remove "(version matters)" or make it a leak to the supported versions wiki page.

README.md
```diff
@@ -44,16 +44,21 @@ For more details, see [building from source](https://github.com/Xpra-org/xpra/tr
 ---
 
 # Usage
+## Initial requirements
+`xpra` must be installed on the client and the host (version matters).  
+
+You can use the html5 client in which case xpra is only required on the host.  People that are using the html5 client would already be somewhat experienced xpra users.
```
totaam, line 50: Please remove "People..." - this is not helpful or true.

tux2bsd: I've removed it as you don't believe it is helpful.

Issue #2834, opened 2020-07-08
Xpra stop crashes X Display :0


totaam:
*Issue migrated from trac ticket # 2834*

**component:** server | **priority:** major | **resolution:** fixed

#### 2020-07-08 17:59:00: brady created the issue

___
> If I start an xpra session using `xpra start :100`, and then run the command `xpra stop`, xpra will not only stop the xpra session, but will also kill my original x session running on DISPLAY :0. It did not kill the applications that were running on that X Server either.
> 
> This rather annoyingly caused a situation in which a program running on X Display :0 was able to capture my keystrokes that I entered into a tty to try to recover my session.
> 
> Xpra version: `xpra v4.0.2-26625`
> OS: Ubuntu 18.04 amd64
> Desktop Environment: KDE Plasma 5

basilgello:
@totaam

> Let's hope new Debian versions… 

No room for false hopes. I had to revert this commit to make Xorg not deallocate a tty.

totaam:
Reverted in 873060b9ee903d6ed5571cb25ba3df26737fd1cf.

Seeing that a huge number of bugs are being resolved by switching to xrandr Xdummy (#56), the only recommendation I can make is to **avoid Debian with xpra**.

totaam:
Checked again and in 2024, the Debian version of `Xorg` is still butchered and unusable as non-root!
https://github.com/Xpra-org/xpra/wiki/Distribution-Packages#specific-distributions

felipecrs:
> Checked again and in 2024, the Debian version of `Xorg` is still butchered and unusable as non-root! [Wiki: Distribution Packages (specific distributions)](https://github.com/Xpra-org/xpra/wiki/Distribution-Packages#specific-distributions)

Is there some easy way to check? Some specific binary with suid bit?

felipecrs:
No suid bit here, in case that's the binary (Debian 13).

```console
$ stat /usr/bin/Xorg
  File: /usr/bin/Xorg
  Size: 274             Blocks: 8          IO Block: 4096   regular file
Device: 0,292   Inode: 6856730     Links: 1
Access: (0755/-rwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2025-08-22 03:33:46.550588963 +0000
Modify: 2025-06-20 07:47:20.000000000 +0000
Change: 2025-08-22 03:31:41.841512524 +0000
 Birth: 2025-08-22 03:31:41.734511602 +0000
[...]

PR #3667, opened 2022-10-22
add GHA build/test

mdavidsaver:
In looking at adding test coverage for #3471 I started by running the existing tests.  Several are not currently passing.

@totaam Do you think running tests automatically is worth pursuing at present?  I could also scale this back to only check compile/install.

```
FAIL: test_find (__main__.FakeXineramaTest)
----------------------------------------------------------------------
  File "/home/runner/work/xpra/xpra/tests/unittests/unit/x11/fakexinerama_test.py", line 17, in test_find
    assert find_libfakeXinerama()
AssertionError
```

```
 ERROR: test_command_server (__main__.ChildCommandMixinTest)
----------------------------------------------------------------------
  File "/home/runner/work/xpra/xpra/tests/unittests/unit/server/mixins/child_command_test.py", line 26, in test_command_server
    self.do_test_command_server()
  File "/home/runner/work/xpra/xpra/tests/unittests/unit/server/mixins/child_command_test.py", line 55, in do_test_command_server
    self._test_mixin_class(_ChildCommandServer, opts)
  File "/home/runner/work/xpra/xpra/tests/unittests/unit/server/mixins/servermixintest_util.py", line 78, in _test_mixin_class
    x = self.mixin = mclass()
  File "/home/runner/work/xpra/xpra/tests/unittests/unit/server/mixins/child_command_test.py", line 52, in _ChildCommandServer
    ccs = child_command_server.ChildCommandServer()
  File "/home/runner/work/xpra/xpra/dist/python3.8/lib/python/xpra/server/mixins/child_command_server.py", line 75, in __init__
    self.idle_add(self.late_start)
AttributeError: 'ChildCommandServer' object has no attribute 'idle_add'
```

```
error with black vp9 image via enc_vpx and dec_vpx
F
======================================================================
FAIL: test_YUV420P (__main__.Test_CSC_Colorspace)
----------------------------------------------------------------------
  File "/home/runner/work/xpra/xpra/tests/unittests/unit/codecs/video_colorspace_test.py", line 118, in test_YUV420P
    self._test_YUV420P(encoding, encoder, decoder, yuvdata)
  File "/home/runner/work/xpra/xpra/tests/unittests/unit/codecs/video_colorspace_test.py", line 56, in _test_YUV420P
    if in_csc not in encoder_module.get_input_colorspaces(encoding):
  File "xpra/codecs/vpx/encoder.pyx", line 218, in xpra.codecs.vpx.encoder.get_input_colorspaces
AssertionError: invalid encoding: vp9
```

```
 FAIL: test_parse_image_path (__main__.TestCommon)
----------------------------------------------------------------------
  File "/home/runner/work/xpra/xpra/tests/unittests/unit/notifications/common_test.py", line 30, in test_parse_image_path
    assert common.parse_image_path(filename) is not None
AssertionError
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,54 @@
+name: xpra
+
+on: [push, pull_request]
+
+jobs:
+  unittest:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: "apt-get install"
+      run: |
+        # see docs/Build/Debian.md
+        sudo apt-get update
+        sudo apt-get -y install libx11-dev libxtst-dev libxcomposite-dev libxdamage-dev libxres-dev \
+                libxkbfile-dev python-all-dev pandoc libsystemd-dev \
+                libgtk-3-dev python3-dev python3-cairo-dev python-gi-dev cython3 \
+                xauth x11-xkb-utils \
+                xvfb python3-cairo python3-gi-cairo \
+                python3-opengl python3-lz4 python3-rencode python3-pil \
+                libturbojpeg-dev libwebp-dev python3-pil \
+                libx264-dev libvpx-dev yasm \
+                python3-numpy \
+                libavformat-dev libavcodec-dev libswscale-dev \
+                python3-opengl \
+                python3-rencode python3-lz4 python3-dbus python3-cryptography \
+                python3-netifaces python3-yaml python3-lzo \
+                openssh-client sshpass python3-paramiko \
+                python3-setproctitle python3-xdg python3-pyinotify python3-opencv \
+                libpam-dev quilt xserver-xorg-dev xutils-dev xserver-xorg-video-dummy xvfb keyboard-configuration \
+                python3-kerberos python3-gssapi \
+                gstreamer1.0-pulseaudio gstreamer1.0-alsa \
+                gstreamer1.0-plugins-base gstreamer1.0-plugins-good \
+                gstreamer1.0-plugins-ugly \
+                cups-filters cups-common cups-pdf python3-cups \
+                liblz4-dev \
+                python3-coverage \
+                x11-xserver-utils dbus-x11
+
+        # tests/unittests/run wants "coverage", but debian installs as /usr/bin/python3-coverage
+        sudo ln -s /usr/bin/python3-coverage /usr/local/bin/coverage
+
+    - name: "Build/Install/Test"
+      run: >
+        python3 setup.py unittests
+        --skip-fail unit.x11.fakexinerama_test
+        --skip-fail unit.server.mixins.child_command_test
+        --skip-fail unit.server.server_sockets_test
+        --skip-fail unit.scripts.parsing_test
+        --skip-fail unit.notifications.common_test
+        --skip-slow unit.client.x11_client_test
+        --skip-slow unit.server.shadow_server_test
+        --skip-slow unit.server.mixins.start_option_test
+        --skip-slow unit.server.mixins.shadow_option_test
--- a/setup.py
+++ b/setup.py
@@ -472,8 +472,9 @@ if "pdf-doc" in sys.argv:
     convert_docs("pdf")
     sys.exit(0)
 
-if "unittests" in sys.argv:
-    os.execv("./tests/unittests/run", ["run"])
+if sys.argv[1]=="unittests":
+    os.execv("./tests/unittests/run", ["run"] + sys.argv[2:])
+assert "unittests" not in sys.argv, sys.argv
 
 
 #*******************************************************************************
--- a/tests/unittests/run
+++ b/tests/unittests/run
@@ -1,4 +1,5 @@
 #!/usr/bin/bash
+set -e
 
 die() { echo "$*" 1>&2 ; exit 1; }
 
--- a/tests/unittests/run
+++ b/tests/unittests/run
@@ -50,6 +50,6 @@ if [ "${XPRA_TEST_COVERAGE}" == "1" ]; then
 	coverage report -m
 	ls -la
 	coverage html
-	xdg-open ./htmlcov/index.html
+	[ "$GITHUB_ACTIONS" ] || xdg-open ./htmlcov/index.html
 fi
 popd
--- a/tests/unittests/unit/codecs/video_colorspace_test.py
+++ b/tests/unittests/unit/codecs/video_colorspace_test.py
@@ -109,6 +109,11 @@ class Test_CSC_Colorspace(unittest.TestCase):
             if not encoder:
                 print(f"{encoder_name} not found")
                 continue
+            try:
+                encoder.get_input_colorspaces(encoding)
+            except AssertionError:
+                print(f"{encoder_name} does not support {encoding}")
+                continue
             decoder = loader.load_codec(decoder_name)
             if not decoder:
                 print(f"{decoder_name} not found")
--- a/tests/unittests/unit/run.py
+++ b/tests/unittests/unit/run.py
@@ -5,14 +5,23 @@
 #runs all the files in "unit/" that end in "test.py"
 
 import sys
+import time
 import os.path
 import subprocess
 
 
 COVERAGE = os.environ.get("XPRA_TEST_COVERAGE", "1")=="1"
 
+def getargs():
+    from argparse import ArgumentParser
+    P = ArgumentParser()
+    P.add_argument('--skip-fail', action='append', default=[])
+    P.add_argument('--skip-slow', action='append', default=[])
+    P.add_argument('-T', '--test', action='append')
+    P.add_argument('test', nargs='*', default=[])
+    return P
 
-def main():
+def main(args):
     if COVERAGE:
         # pylint: disable=import-outside-toplevel
         #only include xpra in the report,
@@ -36,14 +45,12 @@ def main():
     else:
         run_cmd = [os.environ.get("PYTHON", "python3")]
 
-    paths = []
     #ie: unit_dir = "/path/to/Xpra/trunk/src/unittests/unit"
     unit_dir = os.path.abspath(os.path.dirname(__file__))
-    if len(sys.argv)>1:
-        for x in sys.argv[1:]:
-            paths.append(os.path.abspath(x))
-    else:
-        paths.append(unit_dir)
+    paths = args.test or [unit_dir]
+    skip_fail = set(args.skip_fail)
+    skip_slow = set(args.skip_slow)
+
     #ie: unittests_dir = "/path/to/Xpra/trunk/src/unittests"
     unittests_dir = os.path.dirname(unit_dir)
     sys.path.append(unittests_dir)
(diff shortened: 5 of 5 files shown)


totaam:
> `assert find_libfakeXinerama()`

`libfakeXinerama` is being deprecated, except on Debian / Ubuntu which don't have reliably Xdummy support: https://github.com/Xpra-org/xpra/issues/2834#issuecomment-1179500636

> `AttributeError: 'ChildCommandServer' object has no attribute 'idle_add'`

Already fixed for 4.4.x branch: b5722d304d9825c7adceb34c211fa14dc3d9acff, I would like to do something different / better for trunk.

> `ServerSocketsTest`
> (..)
> `Exception: version client failed to connect, returned 18 : CONNECTION_FAILED`

All of the socket tests can get random failures if we hit timeouts rather than the intended code path.
More tricky.
These tests also take way too long to run, I wouldn't include them as a commit hook test.

> `TestPillow`
> (..)
> `ValueError: wrong palette size`

Eh? News to me. What OS / version / Pillow version?

> `UnboundLocalError: local variable 'e' referenced before assignment`

Fixed in ac0c6dcd31ef26a00811b87fc18bb371b35249c1

> `test_parse_image_path`

Is python pillow installed?

---

> Do you think running tests automatically is worth pursuing at present? I could also scale this back to only check compile/install.

It would be very useful, but certainly not all the tests as they take far too long to run and some of them fail intermittently.

mdavidsaver:
I added a print of individual test execution time, and a `--skip` argument to `run.py`, and added skips for all tests which I have seen take longer than 20 seconds.  This includes some which are failing.  The total run time comes down to ~11 minutes, including ~8 min for compilation.

I've updated the list of failures with the 4 I now see.  It looks like I forgot to copy the `vp8` one previously.

mdavidsaver:
> Eh? News to me. What OS / version / Pillow version?

Currently: Ubuntu `20.04.5` and `python3-pil==7.0.0-4ubuntu0.5`.  The debian binary package name is `python3-pil`, but the source package name is [pillow](https://tracker.debian.org/pkg/pillow).

totaam:
> Note that `TestPillow` is one of the tests now being skipped.

It should not be:
```shell
$ ./unit/codecs/pillow_encoder_test.py 
.......
----------------------------------------------------------------------
Ran 7 tests in 2.396s

OK
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,55 @@
...
+        sudo apt-get update
+        sudo apt-get -y install libx11-dev libxtst-dev libxcomposite-dev libxdamage-dev libxres-dev \
+                libxkbfile-dev python-all-dev pandoc libsystemd-dev \
+                libgtk-3-dev python3-dev python3-cairo-dev python-gi-dev cython3 \
+                xauth x11-xkb-utils \
+                xvfb python3-cairo python3-gi-cairo \
+                python3-opengl python3-lz4 python3-rencode python3-pil \
+                libturbojpeg-dev libwebp-dev python3-pil \
+                libx264-dev libvpx-dev yasm \
+                python3-numpy \
```
totaam, line 23: Why is this one here? I thought it should only be needed with the `nvidia` codecs - which are not available when running the unit tests.

.github/workflows/test.yml
```diff
@@ -0,0 +1,55 @@
...
+        # see docs/Build/Debian.md
+        sudo apt-get update
+        sudo apt-get -y install libx11-dev libxtst-dev libxcomposite-dev libxdamage-dev libxres-dev \
+                libxkbfile-dev python-all-dev pandoc libsystemd-dev \
+                libgtk-3-dev python3-dev python3-cairo-dev python-gi-dev cython3 \
+                xauth x11-xkb-utils \
+                xvfb python3-cairo python3-gi-cairo \
+                python3-opengl python3-lz4 python3-rencode python3-pil \
+                libturbojpeg-dev libwebp-dev python3-pil \
+                libx264-dev libvpx-dev yasm \
```
totaam, line 22: Why do we need `yasm`?

.github/workflows/test.yml
```diff
@@ -0,0 +1,55 @@
...
+        --skip unit.x11.x11_server_test
+        --skip unit.client.splash_test
+        --skip unit.client.x11_client_test
+        --skip unit.server.mixins.start_option_test
+        --skip unit.server.mixins.shadow_option_test
+        --skip unit.server.mixins.startdesktop_option_test
+        --skip unit.server.server_auth_test
+        --skip unit.server.server_sockets_test
+        --skip unit.scripts.main_test
+        --skip unit.codecs.pillow_encoder_test
```
totaam, line 55: Do we have a record of why each one of these tests is disabled?
We should work towards running them all, so having this information would help.

totaam (commented):
Everything looks good!
Can we merged this then?

.github/workflows/test.yml
totaam: The pillow test is fixed: 6f28b0bd28f36a5eb171989967e45da57295bbfc

mdavidsaver:
Updated.  `--skip` is changed to `--skip-fail` or `--skip-slow`.  The former is run, with results printed, but failure is ignored.  The latter is not run.

I've pruned back to a minimal set of skips, with only those tests which individually take more than 60 seconds to run (my previous cutoff was `>=` 20 seconds).  This raises the total execution time to ~20 minutes.

> Can we merged this then?

Certainly, if you think this is good enough for the present.  If it becomes too noisy for you, then just disable or remove it.

.github/workflows/test.yml
mdavidsaver: `--skip-fail` vs. `--skip-slow` should record why a test is ignored/skipped.

.github/workflows/test.yml
mdavidsaver: I started by concatenating all of the build dependency lists in `docs/Build/Debian.md`, then adding some missing pieces (eg. `liblz4-dev`.).  I didn't remove any.  I think it would be nice to prune down to a minimal dependency list.  When I encounter a new OSS project, I tend to look at CI configuration for "as built" documentation of dependencies.

However, I suspect this would need more time than I have at present to try possible combinations and/or learn about the dependency chains.  (I see a bewildering number of possible combinations...)

totaam: Removed yasm: 3c76c3254c4ae402c2b1063714a0e7fd91d9be1b

.github/workflows/test.yml
totaam: Removed yasm: 3c76c3254c4ae402c2b1063714a0e7fd91d9be1b

.github/workflows/test.yml
totaam: A lot of these dependencies aren't needed: c4dca10b3c8aab74442fb3b5fcd7353f8288ad5d
The only test that uses numpy is:
https://github.com/Xpra-org/xpra/blob/c4dca10b3c8aab74442fb3b5fcd7353f8288ad5d/tests/unittests/unit/server/motion_test.py#L108-L113
And it will now just skip this one test case.
(numpy is heavy - don't install it if we don't have to)

totaam:
@mdavidsaver how do I make it honour `skip-checks:true`?
df57f3728c393dd381ac26e0208e770e0167eae4 ended up running the tests.

mdavidsaver:
By my reading of https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

> Alternatively, you can end the commit message with two empty lines followed by either:
> 
>     skip-checks:true
>     skip-checks: true

I only see one blank line in the message of df57f3728c393dd381ac26e0208e770e0167eae4

totaam:
Ah, Doh!

mdavidsaver:
Also worth knowing that you (project owner) can enable/disable workflow running through the web ui.

https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow

totaam:
More fixes in git master:
* [clipboard](https://github.com/Xpra-org/xpra/issues/3824) : was a brotli bindings issue
* file-transfer: c9d66e11b16351c472a2839ae4f26610ef5d933c + 416b65a9998f4430b09cf9227534de46dfcb924b + f1c20373eb9f254c9fe6f046bdf65974c943c0c5
* run as root fix: 2678e4dadbb7f7d181e9fc1a2b217eea330a6367
* test more audio codecs: adcab5a851e986a666300c62528d3a228a3e4499 + 2052710111c2ae4bcdccfe016f4b6b3f0f3da051
* `rencode[plus]` tests: 5c8d2095ef4d1a4c926bc9358069fa6b1a878185 (also silences `AT-SPI` warnings that show up on Ubuntu)
* dependencies cleanup: c4dca10b3c8aab74442fb3b5fcd7353f8288ad5d + 3c76c3254c4ae402c2b1063714a0e7fd91d9be1b

totaam:
The proxy server test now also runs fine.  
The only non-slow test failing on the CI is the "server auth test" using the file backends:
https://github.com/Xpra-org/xpra/blob/eeff82989cb3709dd1b8f2c07db89ae83caf91ed/.github/workflows/test.yml#L44C33-L44C49
```
2023-07-23T11:42:02.6871134Z running unit.server.server_auth_test
2023-07-23T11:42:02.6871759Z
2023-07-23T11:45:09.2263209Z ..EE..
2023-07-23T11:45:09.2266056Z ======================================================================
2023-07-23T11:45:09.2266675Z ERROR: test_file (__main__.ServerAuthTest)
2023-07-23T11:45:09.2267784Z ----------------------------------------------------------------------
2023-07-23T11:45:09.2270526Z Traceback (most recent call last):
2023-07-23T11:45:09.2271020Z   File "/home/runner/work/xpra/xpra/tests/unittests/unit/server/server_auth_test.py", line 83, in test_file
2023-07-23T11:45:09.2272617Z     self._test_auth(f"file:filename={f.name}", "", ExitCode.AUTHENTICATION_FAILED, password+"A")
2023-07-23T11:45:09.2273159Z   File "/home/runner/work/xpra/xpra/tests/unittests/unit/server/server_auth_test.py", line 60, in _test_auth
2023-07-23T11:45:09.2273715Z     raise RuntimeError(f"{auth!r} test error: expected info client to return {estr(exit_code)} but got {estr(r)}")
2023-07-23T11:45:09.2274572Z RuntimeError: 'file:filename=/tmp/xpra-y7_fc1zu' test error: expected info client to return AUTHENTICATION_FAILED but got CONNECTION_FAILED
2023-07-23T11:45:09.2274934Z 
2023-07-23T11:45:09.2275070Z ======================================================================
2023-07-23T11:45:09.2276782Z ERROR: test_multifile (__main__.ServerAuthTest)
2023-07-23T11:45:09.2277332Z ----------------------------------------------------------------------
2023-07-23T11:45:09.2277715Z Traceback (most recent call last):
2023-07-23T11:45:09.2278172Z   File "/home/runner/work/xpra/xpra/tests/unittests/unit/server/server_auth_test.py", line 97, in test_multifile
2023-07-23T11:45:09.2278739Z     self._test_auth(f"multifile:filename={f.name}", "", ExitCode.AUTHENTICATION_FAILED, password+"A")
2023-07-23T11:45:09.2279296Z   File "/home/runner/work/xpra/xpra/tests/unittests/unit/server/server_auth_test.py", line 60, in _test_auth
2023-07-23T11:45:09.2279854Z     raise RuntimeError(f"{auth!r} test error: expected info client to return {estr(exit_code)} but got {estr(r)}")
2023-07-23T11:45:09.2280659Z RuntimeError: 'multifile:filename=/tmp/xpra-uvtmzwv1' test error: expected info client to return AUTHENTICATION_FAILED but got CONNECTION_FAILED
2023-07-23T11:45:09.2281031Z 
2023-07-23T11:45:09.2282705Z ----------------------------------------------------------------------
2023-07-23T11:45:09.2283071Z Ran 6 tests in 186.373s
2023-07-23T11:45:09.2283391Z 
2023-07-23T11:45:09.2284988Z FAILED (errors=2)      
```
No idea why as this works just fine from my local system, repeatedly, whereas it fails every time on the CI.
It only fails on the part of the test that tries to authenticate using an invalid password (wrong error code):
https://github.com/Xpra-org/xpra/blob/eeff82989cb3709dd1b8f2c07db89ae83caf91ed/tests/unittests/unit/server/server_auth_test.py#L83
https://github.com/Xpra-org/xpra/blob/eeff82989cb3709dd1b8f2c07db89ae83caf91ed/tests/unittests/unit/server/server_auth_test.py#L97

totaam:
The CI is now running all the server tests again since e3bce07f1b4031dc390d68efb6a610b19b3d4236.
As for the slow ones... still to slow.

totaam:
But of course I spoke too soon and forgot: the server auth test only fails on the CI and nowhere else.

PR #3669, opened 2022-10-23
Update xpra.1

BillyCroan:
adding reference to --splash=no option
updating url for bugreports and questions

### DIFF
--- a/fs/share/man/man1/xpra.1
+++ b/fs/share/man/man1/xpra.1
@@ -602,6 +602,10 @@ In \fIauto\fP mode, the client will set the bandwidth limit
 value to 80% of the maximum speed of the network interface
 it is using to connect to the server.
 
+\fB--splash\fP=\fIno\fP
+Do not display splash screen while connecting.
+.TP
+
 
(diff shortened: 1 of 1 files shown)


totaam:
The `--splash` command line argument takes more than just `no` as value:
https://github.com/Xpra-org/xpra/blob/fdd9d6fc067dd9ca7d541c1de1b6f0fe73f37aa7/xpra/scripts/parsing.py#L1367-L1371
And it defaults to `auto`.

fs/share/man/man1/xpra.1
```diff
@@ -602,6 +602,10 @@ In \fIauto\fP mode, the client will set the bandwidth limit
 value to 80% of the maximum speed of the network interface
 it is using to connect to the server.
 
+\fB--splash\fP=\fIno\fP
+Do not display splash screen while connecting.
```
totaam, line 606: Also, this is not just used for connecting, this is also shown when starting a server from a context which does have access to a GUI.

totaam:
The commit above contains more correct wording.

PR #3684, opened 2022-11-07
Clarify test failures

mdavidsaver:
Make it easier to search CI output for failures which are not ignored.


totaam:
Great, it was taking me too long to skim through the output.

Issue #3695, opened 2022-11-29
Key based auth with paramiko on Windows not working


aneesh2312:
**Describe the bug**
I am using paramiko client for SSH, along with key based authentication on Windows. I am using the XPRA_SSH_DEFAULT_KEYFILES environment variable. Xpra is not parsing the environment variable correctly as it uses ":" as the separator by default, which is not the case for Windows.

**To Reproduce**
1. xpra start on server side.
2. Set XPRA_SSH_DEFAULT_KEYFILES environment variable on client side.
3. xpra_cmd attach ssh:ubuntu@<host> --ssh=paramiko:auth=key

**System Information (please complete the following information):**
 - Server OS: Linux (Ubuntu)
 - Client OS: Windows 10
 - Xpra Server Version: 4.4.3
 - Xpra Client Version: 4.4.3

totaam:
I don't want to risk breaking another use case, so I only merged the part the uses `pathsep`.
As per my previous comment, trying to load from `.ssh` should be (almost) free.

PR #3696, opened 2022-11-29
Use OS pathsep instead of colon to get default keys

aneesh2312:
Also skip searching in ~/.ssh for key files on Windows.


aneesh2312:
This should fix #3695

totaam:
Ah, right about `pathsep`.
Are you sure that there aren't openssh builds for windows that look for `.ssh` ?
(it should be almost free to look for those paths even if they don't exist)

aneesh2312:
Ack, agree that looking for those paths shouldn't add much overhead. Thanks, closing this.

Issue #3697, opened 2022-11-29
Paramiko client error if display is not specified


aneesh2312:
**Describe the bug**
If the display number is not specified when using paramiko client for SSH, we run into a type conversion error (`connection failed: 'NoneType' object has no attribute 'replace'`).

**To Reproduce**
1. xpra start (on server)
2. xpra_cmd attach ssh:ubuntu@ --ssh=paramiko:auth=key

**System Information (please complete the following information):**
 - Server OS: Linux (ubuntu)
 - Client OS: Windows 10
 - Xpra Server Version 4.4.3
 - Xpra Client Version 4.4.3

**Additional context**
Seems to be from https://github.com/Xpra-org/xpra/commit/e43a738d2c25ce86929656cd1e77a3035cbbeca9

totaam:
PR merged, thanks.

Another bug caused by #3599 - not entirely surprising either.

PR #3698, opened 2022-11-29
Handle missing display number

aneesh2312:
Currently if display number is not specified, a None to str conversion error is seen when using paramiko for SSH


aneesh2312:
This should fix #3697

totaam:
Thanks!

Issue #3531, opened 2022-04-23
Unmet dependencies when installing on Debian testing/bookworm 


markmandel:
**Describe the bug**

Due to a change in version of python3, as well as some other package changes (see below) xpra will no longer install on Debian testing/bookworm (it actually removes it if you do an `apt full-upgrade`).

**To Reproduce**

```bash
➜  ~ sudo apt install xpra
[sudo] password for mark:
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 xpra : Depends: python3 (< 3.10) but 3.10.4-1 is to be installed
        Depends: libwebp6 (>= 0.5.1) but it is not installable
        Depends: libx264-160 (>= 2:0.160.3011+gitcde9a93) but it is not installable
        Recommends: xpra-html5 but it is not going to be installed
        Recommends: python3-rencode but it is not going to be installed
        Recommends: python3-zeroconf but it is not going to be installed
        Recommends: python3-netifaces but it is not going to be installed
        Recommends: python3-gst-1.0 but it is not going to be installed
        Recommends: gir1.2-appindicator3-0.1 but it is not installable
        Recommends: gir1.2-rsvg-2.0 but it is not going to be installed
        Recommends: ibus but it is not going to be installed
        Recommends: python3-qrencode but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

**System Information (please complete the following information):**
 
```bash
➜  ~ lsb_release --all

No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux bookworm/sid
Release:        testing
Codename:       bookworm
```

Xpra is installed via deb repository:

```bash
➜  sources.list.d cat xpra_org.list
# deb https://xpra.org/ bullseye main
deb https://xpra.org/ bookworm main
```

**Additional context**

I've no idea how relevant it is, but it looks like there are the following similar packages in Debian testing now:

* https://packages.debian.org/bookworm/libx264-163
* https://packages.debian.org/bookworm/libwebp7
* https://packages.debian.org/bookworm/python3.9

Just in case that is useful information.

totaam:
I believe that this is similar to #3525: use the beta channel until bookworm is released.
[...]

PR #3738, opened 2023-01-15
Fix build issues on Debian Bookworm and sid

adamnew123456:
Currently neither Xpra stable nor beta install on Debian unstable or testing, with apt blaming two packages:

- `python3`: current Xpra builds require `< 3.11`, but unstable made [3.11 the primary version of `python3`](https://packages.debian.org/sid/python3) recently. The same seems to have happened in testing though the packaging site says it's still at 3.10
- `libprocps8`: no longer exists in either [sid or Bookworm](https://packages.debian.org/bullseye/libprocps-dev), having been replaced by [`libproc2`](https://packages.debian.org/sid/libproc2-dev).

The `python3` dependency doesn't really matter, it's just a case of the build being stale. Xpra builds and runs on Python 3.11 without issue.

`procps` is more involved - the changes here are currently "works on my machine". Some cleanup is needed to make sure they work for Debian stable and other distributions. 3 is the only one I'm not sure about yet - it depends on how many of the supported distributions ship procps-ng vs. procps, and what all the library and header names are.

1. In Debian, `libprocps-dev` doesn't exist outside of sid and unstable. [Easily resolved](https://www.debian.org/doc/debian-policy/ch-relationships.html), I just have to commit the fix for this.
2. The setup script needs to determine which name the library is available under, so that it can be passed to the Cython build flags. This should be doable via `pkgconfig`.
3. The Cython code itself is based on an API that the library hasn't exposed for a while. Both `procps.h` (used to check the dependency) and `readproc.h` (used in the implementation) are omitted from `libproc2-dev`. The new API looks like this, the new Cython procps code is just a translation:

```c
#include <libproc2/pids.h>
#include <stdio.h>
#include <stdlib.h>

// Usage: ./a.out PID

int main(int argc, char **argv) {
        if (argc == 1) return 1;

        struct pids_info *handle = NULL;
        enum pids_item selector = PIDS_ID_PPID;
        if (procps_pids_new(&handle, &selector, 1)) return 1;

        unsigned pid = atoi(argv[1]);
        if (!pid) return 1;

        struct pids_fetch *info = procps_pids_select(handle, &pid, 1, PIDS_SELECT_PID);
        if (!info) return 1;

        int count = info->counts->total;
        printf("Total: %d\n", count);

        for (int i = 0; i < count; i++) {
                struct pids_result *proc_info = info->stacks[i]->head;
                printf("Info[%d]: {type=%d value=%d}\n", i, proc_info->item, proc_info->result.s_int);
        }

        procps_pids_unref(&handle);
}
```


adamnew123456:
All fixed. For 3 there are two versions of the proc extension, one for libproc2 using the procps_pids API and the other for libprocps which uses the readproc API. The setup will compile  one based on which header file is present. `xpra.platform.xposix.proc` itself is now just a shim that loads whatever module is available.

totaam:
Thanks!

> The `python3` dependency doesn't really matter, it's just a case of the build being stale. Xpra builds and runs on Python 3.11 without issue.

Correct, we don't rebuild every time Debian breaks the ABI.
Use the beta channel for that, which has fresh builds regularly enough.

You forgot to update the clean and build files - done in a91a4e44be642fb3c48aec460fcf3a8498925660.
I also made the script executable to make it easier to test it: 29380bff3bd0cb6a57c6d76432345f6000f52653
ie:
```shell
$ ./xpra/platform/xposix/proc.py $$
using `get_parent_pid`=<cyfunction get_parent_pid at 0x7fab2ef35b10>
from xpra.platform.xposix.proc_procps
 get_parent_pid(27453)=11900
```

AGI-chandler:
Hey guys, apt uninstalled `xpra=4.3.4-r0-1` today while I was doing some upgrades, I need that version because it works with the older systems I still need to work with.  However, I can't install any of the newer versions or even the latest beta versions, because of Debian python3 I guess.

All the xpra's from 4.3.4-r0-1 through 5.0-r32584-1 say `Depends: python3 (<< 3.11), python3 (>= 3.10~), python3:any (>= 3.6~)`.  My system has python3.10=3.10.9-1, python3.11=3.11.1-2, and python3=3.11.1-1 installed.

So how could this be fixed?

totaam:
> apt uninstalled `xpra=4.3.4-r0-1` today

Removed or upgraded?

> I need that version because it works with the older systems

There are no known regressions in 4.4.x - why would you need this specific version?

> All the xpra's from `4.3.4-r0-1` through `5.0-r32584-1` say ...

Bookworm is unstable and breaks the python ABI regularly: https://github.com/Xpra-org/xpra/issues/3531#issuecomment-1107525545
Those builds were made against the current python ABI at that point in time.

> So how could this be fixed?

Difficult to say! As I'm not sure what's broken yet.

AGI-chandler:
> > apt uninstalled `xpra=4.3.4-r0-1` today
> >
> Removed or upgraded?
>
Removed

> > I need that version because it works with the older systems
> >
> There are no known regressions in 4.4.x - why would you need this specific version?
>
According to https://github.com/Xpra-org/xpra/discussions/3717, 4.3.4-r0-1 was the latest version still compatible with 1.0.14-1.r24088.el6_10

totaam:
Ah, 1.0.x was EOLed a long time ago, so good luck with that!

AGI-chandler:
Thanks!  I'm sure I'll need it!  Not sure what kind of "enterprise OS" doesn't allow itself to be upgraded but I'm sure the hardware will appreciate it when the time comes.

> Bookworm is unstable 
Technically it's testing and sid is unstable 😉

Now, on my system with bullseye/stable, it has xpra latest stable version 4.4.3-r0-1 installed along with python3 version 3.9.2-3 and python3.9 version 3.9.2-1...huh?
`python3.9 is already the newest version (3.9.2-1).`
`python3 is already the newest version (3.9.2-3).`

On the system in question I guess I need to install python3.9 package... hmm even after that's installed, xpra is still complaining that I have python3 > 3.11 I guess:
` xpra : Depends: python3 (< 3.11) but 3.11.1-1 is to be installed`

Yes but 3.10 and 3.9 are right there for xpra why isn't xpra happy with that?  Is xpra's dependency of `python3 (<< 3.11), python3 (>= 3.10~), python3:any (>= 3.6~)` really necessary or can that be updated/changed at all?  or is this problem with `apt`?

totaam:
> Technically it's testing and sid is unstable wink

You say potato, I say potato. :wink:

> On the system in question I guess I need to install python3.9 package..

Not sure what you're trying to do, but you can't mix packages meant for another distro version.
xpra doesn't declare any specific python version dependency, the Debian packaging tools do that automagically at build time.

adamnew123456:
> On the system in question I guess I need to install python3.9 package... hmm even after that's installed, xpra is still complaining that I have python3 > 3.11 I guess: ` xpra : Depends: python3 (< 3.11) but 3.11.1-1 is to be installed`
> 
> Yes but 3.10 and 3.9 are right there for xpra why isn't xpra happy with that? Is xpra's dependency of `python3 (<< 3.11), python3 (>= 3.10~), python3:any (>= 3.6~)` really necessary or can that be updated/changed at all? or is this problem with `apt`?

No, apt is correct here - you can't build an Xpra package using Python 3.11 and have it run under 3.10 or some other version. Parts of it will "work" but it'll break hard because some [core pieces](https://github.com/Xpra-org/xpra/blob/master/xpra/rectangle.pyx) are done as Cython modules, and those are always built for a specific version of Python.

AIUI the (only?) supported way to run Xpra on rolling distros like unstable and testing is to use https://xpra.org/beta. It's updated more frequently to track the packages as they change. The catch is that it only tracks the most recent release of Xpra. The joy of the rolling release is that you get the latest version, whether you want it or not :)

If you're comfortable building the package yourself (it's easier than you think), it should be possible to build 4.3.4 against a newer Debian release by fiddling with the package dependencies like I did. Otherwise you might just want to setup CentOS 6 or Debian Bullseye in a container and run Xpra that way - [Toolbox](https://github.com/toolbx-images/images) would probably be the easiest method.

AGI-chandler:
> Not sure what you're trying to do

@totaam Just trying to get _any_ of the xpra packages installed...
<br>
<br>

> you can't build an Xpra package using Python 3.11 and have it run under 3.10 or some other version

@adamnew123456 Then how is it that xpra 4.4.3-r0-1 is installed and running on my Debian bullseye machine with only python 3.9?

> AIUI the (only?) supported way to run Xpra on rolling distros like unstable and testing is to use https://xpra.org/beta

Yes but, as I said:
> All the xpra's from 4.3.4-r0-1 through 5.0-r32584-1 say `Depends: python3 (<< 3.11), python3 (>= 3.10~), python3:any (>= 3.6~)`.  My system has python3.10=3.10.9-1, python3.11=3.11.1-2, and python3=3.11.1-1 installed.

and now it also has python3.9=3.9.2-1.  Plenty of snakes around but xpra doesn't want any of them...
<br>

> If you're comfortable building the package yourself (it's easier than you think), it should be possible to build 4.3.4 against a newer Debian release by fiddling with the package dependencies like I did. Otherwise you might just want to setup CentOS 6 or Debian Bullseye in a container and run Xpra that way - [Toolbox](https://github.com/toolbx-images/images) would probably be the easiest method.

It's looking like I don't have much of a choice, my comfort isn't the problem, it's the time I have unfortunately...

totaam:
FYI regarding `libprocps`: #4113

PR #3782, opened 2023-03-02
fix focal beta component

zatricky:
Other sources files were already corrected in a previous commit ; but the source of the error, focal, was not fixed. :-)


totaam:
Thanks!

PR #3794, opened 2023-03-16
Gstreamer vaapih264enc params

TijZwa:
`stream-format=avc` causes `transform could not transform video/x-raw, width=(int)128, height=(int)128, format=(string)NV12, framerate=(fraction)60/1, interlace=(string)progressive in anything we support` (on Intel)

`"tune"  : 3` does not work on AMD Radeon Vega.
`"tune"  : 1` works on Intel and AMD


totaam:
OK, let's make these changes and also it more easily configurable.

PR #3836, opened 2023-04-27
Fixed many typos

polluks2:


totaam:
That was fun!
At least I am semi consistent in the words I misspell: `overriden` and `limitting` contributing a significant proportion of this PR.
A few are likely to come back as they're copied from external header files.

I'm not sure what tool you have used to generate these changes, can you specify?
(I was expecting more false positives with "simplified English")

polluks:
Your new friend is "codespell".

PR #3859, opened 2023-05-20
Add a shortkey for the keyboard layout manual switching

graph-inc:
The usual way for switching the keyboard layouts is to ask the platform (i.e., Windows, POSIX) to perform the change and update the xpra keymap when the platform layout switching is detected. This merge request enables the opposite workflow by introducing the `next_keyboard_layout(self, update_platform_layout: bool)` method, so users can press a platform-independent shortkey and ask the xpra to update its keymap, and optionally inform the platform to switch its keyboard layout.

This patch set requires three CLI options to work independent of the platform settings (of course, except the local support for the chosen keyboard layout). For example, to have "us" and "ir" keyboard layouts, starting by the "us" layout and switching to "ir" and back to the "us" again whenever the "Control+space" shortkey is pressed, following options may be passed:

```bash
xpra attach :1234 --keyboard-layout="us" --keyboard-layouts="us,ir" --key-shortcut=Control+space:'next-keyboard-layout(true)'
```

Usage of the `false` boolean argument is to avoid race conditions when a shortkey is chosen which is monitored by the platform too. For example, if `Alt+Shift+Shift_L` is chosen in Windows platform, `--key-shortcut=Alt+Shift+Shift_L:'next-keyboard-layout(false)'` is appropriate since Windows will switch the layout automatically.
The platform layout switching is only supported in POSIX and Win32 platforms by this patch set.

A side-benefit of this opposite workflow (changing the xpra layout first and updating the platform layout later) is that it is more resilient against the platform layout switching detection bugs and complements the keyboard layout locking feature.


xpra/client/gui/keyboard_shortcuts_parser.py
```diff
@@ -114,6 +114,10 @@ def parse_shortcuts(strs=(), shortcut_modifiers=(), modifier_names=()):
                         args.append(None)
                     elif x.find(".") != -1:
                         args.append(float(x))
+                    elif x in ("yes", "true", "on"):
+                        args.append(True)
+                    elif x in ("no", "false", "off"):
```
totaam, line 119: best to use `x.lower() in TRUE_OPTIONS` / `x.lower() in FALSE_OPTIONS`

xpra/keyboard/layouts.py
```diff
@@ -58,7 +58,7 @@
            1061: ("ETI", "Estonia",        "Estonian",                 1257,   "ee", ["nodeadkeys", "dvorak", "us"]),
            1062: ("LVI", "Latvia",         "Latvian",                  1257,   "lv", ["apostrophe", "tilde", "fkey", "modern", "ergonomic", "adapted"]),
            1063: ("LTH", "Lithuania",      "Lithuanian",               1257,   "lt", ["std", "us", "ibm", "lekp", "lekpa"]),
-           1065: ("FAR", "Iran",           "Farsi",                    1256,   "", []),
+           1065: ("FAR", "Iran",           "Farsi",                    1256,   "ir", []),
            1066: ("VIT", "Viet Nam",       "Vietnamese",               1258,   "vn", []),
```
totaam, line 62: This looks safe and should fix `ir` layout support.

xpra/keyboard/layouts.py
```diff
@@ -296,7 +296,7 @@
     0x00000463  : ("af", "Pashto (Afghanistan)"),
     #duplicate of 'ku'
     #0x00000429  : ("ir", "Persian"),
-    0x00050429  : ("ir", "Persian (Standard)"),
+    0xa0000429  : ("ir", "Persian (Standard)"),
```
totaam, line 299: This conflicts with:
```
    0x00000429  : ("ku", "Central Kurdish"),
```
Which is why it is commented out in the line just above - I have no idea which one is correct.
If it is indeed this one, then the other one should be commented out.

xpra/platform/posix/keyboard.py
```diff
@@ -37,6 +40,55 @@ def init_vars(self):
...
+
+        self.__dbus_helper.call_function(
+            "org.gnome.Shell",
+            "/org/gnome/Shell",
+            "org.gnome.Shell",
+            "Eval",
+            [native_to_dbus(ism + cmd)],
+            ok_cb,
+            err_cb,
+        )
```
totaam, line 82: Doesn't this trigger an error on non-gnome-shell desktops? Or is it only going to trigger `err_cb`?

xpra/platform/win32/keyboard.py
```diff
@@ -60,6 +94,16 @@ def init_vars(self):
...
+        self.__x11_layouts_to_win32_hkl = x11_layouts_to_win32_hkl()
+
+    def set_platform_layout(self, layout):
+        hkl = self.__x11_layouts_to_win32_hkl.get(layout)
+        if hkl is None:
+            return 0
+        # https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-activatekeyboardlayout
+        # KLF_SETFORPROCESS|KLF_REORDER = 0x108
+        old_hkl_or_zero_on_failure = ActivateKeyboardLayout(hkl, 0x108)
+        return old_hkl_or_zero_on_failure
```
totaam, line 106: None of the other `set_platform_layout` functions return a value.
If anything, they should all return a boolean. (no point in exposing the "old hkl" to callers - they won't know what to do with it)

xpra/platform/win32/keyboard.py
```diff
@@ -37,6 +38,39 @@ def _GetKeyboardLayoutList():
...
+                        _layout, kbid, _variants, code, hkli)
+                    if _layout not in layout_to_hkl:
+                        layout_to_hkl[_layout] = hkl
+                        break
+    except Exception:
+        log("x11_layouts_to_win32_hkl()", exc_info=True)
+    return layout_to_hkl
+
+
+
```
totaam, line 73: Small nitpick: one CR too many!

totaam (changes requested):
Very good PR.

If you can address the minor points in this review, LGTM!

xpra/keyboard/layouts.py
graph-inc: For the Persian language, the 0xa0000429 is the correct one in Windows. Before introduction of the "Persian (Standard)" layout, the "ku" layout was used as a workaround, but since Windows 10, the default is changed (and matches with the Linux choice thereafter).

```python
            for hkl in l:                                               
                for mask, bitshifts in KMASKS.items():                  
                    kbid = 0                                            
                    for bitshift in bitshifts:                          
                        kbid = (hkl & mask)>>bitshift                   
                        if kbid in WIN32_LAYOUTS:                       
                            break                                       
                    if kbid in WIN32_LAYOUTS:                           
                        code, _, _, _, _layout, _variants = WIN32_LAYOUTS.get(kbid)
                        log("found keyboard layout '%s' / %#x with variants=%s, code '%s' for kbid=%#x",                
                            _layout, kbid, _variants, code, hkl)        
                        if _layout not in layouts_defs:                 
                            layouts_defs[_layout] = hkl                 
                            break
```

Shouldn't we keep both uncommented since they are describing different layout names? ir vs ku?
Users either have **ku** (before Windows 10) or **ir** (after Windows 10) and if a user enable both layouts manually, since the iteration on `KMASKS.items()` is ordered, the ku=0x00000429 will not catch the ir=0xa0000429 layout.
That is, an `hkl` which is supposed to match with **ir**, will match with it in the first iteration (and in the second iteration, `if _layout not in layouts_defs:` prevents the **ir** detected layout to be overridden) while an `hkl` which is going to match with **ku** will never match with the **ir** (due to its extra 0x0a byte).

xpra/platform/posix/keyboard.py
graph-inc: In absence of the `org.gnome.Shell` target, only the `err_cb` will be invoked which logs the failure.
The `DBusHelper.call_function` stops and calls `err_cb` if

1. Target bus name/path does not exists at all (`self.bus.get_object(bus_name, path)`),
2. It exists, but they have removed the chosen function or an ACL is blocking our access to that function (`obj.get_dbus_method(function, interface)`),
3. That function is accessible, but fails to run due to any other internal reason (`fn.call_async`).

xpra/platform/win32/keyboard.py
graph-inc: I removed the return value and just logged the failures because in the "gnome dbus request" scenario, successful/failed response will be determined by a callback and it does not seem good to block the shortkey processing until the callback is invoked.

xpra/keyboard/layouts.py
totaam: You're absolutely right, I had missed the `0xa` prefix.

totaam:
Thanks!

totaam:
Also got complaints that this broke the keyboard for users who don't have / want dbus installed, the commit above fixes that.

PR #3866, opened 2023-05-26
repos: define available architectures (amd64, arm64) for deb822-format

xalt7x:
On multi-arch systems (e.g. i386<->amd64) without specification of supported architectures "apt update" warns about missing packages
(e.g. "Skipping acquire of configured file 'main/binary-i386/Packages' as repository 'https://xpra.org jammy InRelease' doesn't support architecture 'i386'")


totaam:
Thanks!

Issue #3937, opened 2023-07-25
Proxy cannot establish connection: `'SocketProtocol' object has no attribute 'encryption'`


goekce:
Using proxy according to the example [in the proxy-server docs](https://github.com/Xpra-org/xpra/blob/master/docs/Usage/Proxy-Server.md#diagram) leads to an error. The client waits in the loop afterwards.

1. server: `xpra start`
2. server: `xpra proxy :20 --tcp-auth=allow --bind-tcp=0.0.0.0:14501 --daemon=no`
3. client: `xpra attach tcp://SERVER_ADDR:14501`

- ArchLinux
- version (server & client): latest (g27b450d87)

The client logs look typical. Only problem: it waits for the session to be established.

<details><summary>Server logs:</summary>
<pre><code>
2023-07-25 12:25:05,988 created tcp socket '0.0.0.0:14501'
2023-07-25 12:25:05,989 no uinput module (not usually needed)
2023-07-25 12:25:06,098 serving html content from '/usr/share/xpra/www'
2023-07-25 12:25:06,099 wrote pid 124326 to '/run/user/1000/xpra/20/server.pid'
2023-07-25 12:25:06,132 Warning: socket directory '/run/xpra'
2023-07-25 12:25:06,132  expected permissions 775 but found 777
2023-07-25 12:25:06,134 created unix domain socket '/run/user/1000/xpra/20/socket'
2023-07-25 12:25:06,154 Warning: cannot watch for application menu changes without pyinotify:
2023-07-25 12:25:06,154  No module named 'pyinotify'
2023-07-25 12:25:06,161 Warning: cannot load menu data                              
2023-07-25 12:25:06,161  No module named 'xdg'                                
2023-07-25 12:25:06,164 xpra is ready.             
2023-07-25 12:25:11,828 python-zeroconf: '::' IPv6 loopback address is not supported
2023-07-25 12:25:13,444 xpra proxy server version 5.0-r33870 (g27b450d87) beta
2023-07-25 12:25:13,468  uid=1000 (u), gid=1000 (u)
2023-07-25 12:25:13,468  running with pid 124326 on Linux Arch rolling n/a
2023-07-25 12:25:23,277 New tcp connection received                          
2023-07-25 12:25:23,277  from '10.10.10.10:52928'                                        
2023-07-25 12:25:23,277  on '0.0.0.0:14501'              
2023-07-25 12:25:23,280 Authentication required by allow authenticator module      
[...]

PR #3940, opened 2023-07-26
#3937 make `packet` mutable for reprocessing

goekce:
`_packet_recompress` in `xpra/server/proxy/proxy_instance.py` modifies the `packet`. So make `packet` mutable for reprocessing.


totaam:
Definitely not the correct fix.

Please see https://github.com/Xpra-org/xpra/issues/3937#issuecomment-1651382822

PR #3942, opened 2023-07-26
Fix immutable packet modifications

goekce:
I was getting the following error when I use Firefox:

```
TypeError: 'tuple' object does not support item assignment 
2023-07-26 11:59:37,799 error encoding packet              
  File "/usr/lib/python3.11/site-packages/xpra/server/proxy/proxy_instance.py", line 614, in encode_loop
    if self.process_draw(packet):                          
       ^^^^^^^^^^^^^^^^^^^^^^^^^                           
  File "/usr/lib/python3.11/site-packages/xpra/server/proxy/proxy_instance.py", line 644, in process_draw
    packet[7] = Compressed(f"{encoding} pixels", pixels)   
    ~~~~~~^^^                                              
TypeError: 'tuple' object does not support item assignment
```

After this patch I get:

```
...
2023-07-26 12:18:05,391 Warning: found a large uncompressed item
2023-07-26 12:18:05,391  in packet 'draw' at position 7: 245194 bytes
...
```

Probably this is not the right fix.

### DIFF
--- a/xpra/server/proxy/proxy_instance.py
+++ b/xpra/server/proxy/proxy_instance.py
@@ -371,7 +371,7 @@ class ProxyInstance:
         # make the packet data mutable and replace the contents at `index`:
         assert index>0
         lpacket = list(packet)
-        list_packet[index] = Compressed("file-data", packet[index])
+        lpacket[index] = Compressed("file-data", packet[index])
         # noinspection PyTypeChecker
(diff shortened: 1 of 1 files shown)


totaam:
The fixes are correct. Good work!

As for the warning, I will try to take a look.

PR #3998, opened 2023-09-11
Support microphone for Chromium, improve audio source detection, and fix pactl output parsing

graph-inc:
This pull request addresses one server-side and three client-side issues.

In the server side, "Xpra Speaker" and "Xpra Microphone" are created as two sink devices, so client programs may write into the speaker while the xpra server reads from the "Monitor of Xpra Speaker" device and transmit them to the client side to be played. Similarly, xpra server writes into the "Xpra Microphone" and expects the programs to read their audio input stream from the "Monitor of Xpra Microphone" device.
This setup works for Mozilla Firefox flawlessly.
However, Chromium refuses to read from a monitor device and reports that there is no connected microphone. This issue is resolved by creating a remapped source device which reads from the "Monitor of Xpra Microphone" and provides it for readers as
a non-monitor source device.

In the client code,
the output of `pactl list` command is parsed line by line, looking for non-indented lines as text-block beginners, in order to find out about the input/source and output/sink devices. The following indented lines in each text-block provide attributes such as name and description for each input/output device. When a new text block is observed, the previous block (which its attributes should have been read so far) will be processed, variables which denote those attributes are reset, and then the next input/output device line is checked in order to accumulate its attributes.
The issue is that whenever a non-interesting device is detected, its processing is stopped with a `continue` which bypasses the initialization of the attribute variables. This allows a variable to keep its value during the processing of the next block attributes and if the next block is not going to provide any value for that attribute, the old value from the previous block will be used instead, causing false matches.

Second client-side issue is an improvement in detection of some virtual devices which do not report the `device.class="monitor"` attribute, even though they are monitor devices. These devices can be detected by looking for the `Monitor of Sink` line which reports the corresponding sink device or `n/a` if there is no such a sink device. This pull request, in the special scenario that `device.class` is missing, checks for presence of the `Monitor of Sink` line and only ignores a device if its `is_monitor` state may not be detected with both methods.

Third and last, the computation of the `microphone_enabled` was correct for a single value as it was compared with the `"on"` constant. However, if a device was specified in the same option (i.e., "on:device"), it failed to enable the microphone by default (although the "device" part was parsed out correctly and enabling the microphone via a menu could work as expected).

### DIFF
--- a/xpra/audio/pulseaudio/pactl_impl.py
+++ b/xpra/audio/pulseaudio/pactl_impl.py
@@ -137,14 +137,18 @@ def get_pa_device_options(monitors=False, input_or_output=None, ignored_devices=
 
 def do_get_pa_device_options(pactl_list_output, monitors=False, input_or_output=None,
                              ignored_devices=("bell-window-system",)):
-    def are_properties_acceptable(name: Optional[str], device_class: Optional[str]) -> bool:
-        if name is None or device_class is None:
+    def are_properties_acceptable(name: Optional[str], device_class: Optional[str],
+                                  monitor_of_sink: Optional[str]) -> bool:
+        if (name is None) or (device_class is None and monitor_of_sink is None):
             return False
         if name in ignored_devices:
             return False
         #Verify against monitor flag if set:
         if monitors is not None:
-            is_monitor = device_class == '"monitor"'
+            if device_class is not None:
+                is_monitor = device_class == '"monitor"'
+            else:  # monitor_of_sink is not None
+                is_monitor = monitor_of_sink != "n/a"
             if is_monitor != monitors:
                 return False
         #Verify against input flag (if set):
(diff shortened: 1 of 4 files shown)


totaam:
Thanks!

And congratulations for figuring it all out.
(apologies for this ugly old code you've had to read through - it's one of the worst subsystems, I swear!)

--

I have backported this to the v5 branch.

totaam:
In hindsight, I could / should have skipped the backporting of 16343d39e58aecde78bf91f7f185783171ec0cb7.

graph-inc:
Maybe, it is good to backport [3aebb0c0c](https://github.com/Xpra-org/xpra/pull/3998/commits/3aebb0c0c9eddcee2ae3702ececab0c2775dbbde) too (so `--microphone=on:device-name` option can be used).

totaam:
@graph-inc done!

PR #4014, opened 2023-09-25
Fix limitation for size of inbound\outbound clipboard

dexion:
Hi, you already had a draft for this functionality, but it was not fully implemented. I have added validation logic, and fixed the logging process.

Now in `/etc/xpra/xpra.conf` you can specify a config like:
```
env=XPRA_MAX_CLIPBOARD_RECEIVE_SIZE=200
env=XPRA_MAX_CLIPBOARD_SEND_SIZE=50
```

Up to 200 bytes can be sent **TO** the clipboard on a machine running xpra
**FROM** a machine running xpra you can send up to 50 bytes to the clipboard.

### DIFF
--- a/xpra/clipboard/clipboard_core.py
+++ b/xpra/clipboard/clipboard_core.py
@@ -532,7 +532,16 @@ class ClipboardProtocolHelperCore:
         return _filter_targets(local_targets)
 
     def _munge_raw_selection_to_wire(self, target:str, dtype:str, dformat:int, data) -> Tuple[Any,Any]:
-        log("_munge_raw_selection_to_wire%s", (target, dtype, dformat, repr_ellipsized(bytestostr(data))))
+        log.info("_munge_raw_selection_to_wire, target=%s, type=%s, format=%s, len(data)=%s",
+                 target, dtype, dformat, len(data or b""))
+        if self.max_clipboard_send_size > 0:
+            log("perform clipboard limit checking - datasize - %d, %d", len(data), self.max_clipboard_send_size)
+            max_send_datalen = self.max_clipboard_send_size * 8 // get_format_size(dformat)
+            if len(data) > max_send_datalen:
+                olen = len(data)
+                data = data[:max_send_datalen]
+                log.info("clipboard data copied out truncated because of clipboard policy %d to %d",
+                         olen, max_send_datalen)
         # Some types just cannot be marshalled:
         if dtype in ("WINDOW", "PIXMAP", "BITMAP", "DRAWABLE",
                     "PIXEL", "COLORMAP"):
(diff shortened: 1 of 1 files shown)


xpra/clipboard/clipboard_core.py
```diff
@@ -580,14 +589,15 @@ def _do_munge_raw_selection_to_wire(self, target:str, dtype:str, dformat:int, da
...
+        log.warn("wire selection to raw, encoding=%s, type=%s, format=%s, len(data)=%s",
             encoding, dtype, dformat, len(data or b""))
         if self.max_clipboard_receive_size > 0:
+            log.debug("perform clipboard limit checking - datasize - %d, %d", len(data), self.max_clipboard_send_size)
             max_recv_datalen = self.max_clipboard_receive_size * 8 // get_format_size(dformat)
             if len(data) > max_recv_datalen:
                 olen = len(data)
                 data = data[:max_recv_datalen]
-                log.info("clipboard data copied out truncated because of clipboard policy %d to %d",
+                log.warn("clipboard data copied in truncated because of clipboard policy %d to %d",
```
totaam, line 600: As above.

xpra/clipboard/clipboard_core.py
```diff
@@ -532,7 +532,16 @@ def remote_targets(self, local_targets:Iterable[str]):
...
+        log.warn("_munge_raw_selection_to_wire, target=%s, type=%s, format=%s, len(data)=%s",
+                 target, dtype, dformat, len(data or b""))
+        if self.max_clipboard_send_size > 0:
+            log.debug("perform clipboard limit checking - datasize - %d, %d", len(data), self.max_clipboard_send_size)
+            max_send_datalen = self.max_clipboard_send_size * 8 // get_format_size(dformat)
+            if len(data) > max_send_datalen:
+                olen = len(data)
+                data = data[:max_send_datalen]
+                log.warn("clipboard data copied out truncated because of clipboard policy %d to %d",
+                         olen, max_send_datalen)
```
totaam, line 544: Just a note - not something to deal with now: we should probably do something smarter when handling strings.
Chopping the byte array might give an invalid UTF8 sequence.

dexion:
I fixed all your comments

totaam:
Can you squash these two cookies into one please?

dexion:
Branch was squashed into one commit. BTW, you could also do squash when you merge branch (you would have an option for it in interface)

totaam:
Only just realized that this was merged in v5.x, in the future please submit all PRs against git master instead.

totaam:
I should have been more careful, especially with the v5.x stable branch.
The commit you forced push doesn't address the review comments: using `log.info` is going to spam the log file.

dexion:
These info messages could be used to send additional info/reactions to external systems and you didn't need to enable really huge log with `-d clipboard`. These log-strings didn't appear to often, they couldn't spam the main log. 

BTW you are still have log.info at https://github.com/Xpra-org/xpra/blob/master/xpra/clipboard/core.py#L598 and https://github.com/Xpra-org/xpra/blob/master/xpra/clipboard/core.py#L541 (if you want to move all of these logs to debug). I would prefer log.info on both strings: https://github.com/Xpra-org/xpra/blob/master/xpra/clipboard/core.py#L534 and https://github.com/Xpra-org/xpra/blob/master/xpra/clipboard/core.py#L541

Issue #3657, opened 2022-10-18
replace opencv with libcamera?


totaam:
[libcamera](https://github.com/kbingham/libcamera) is _A complex camera support library for Linux, Android, and ChromeOS_

It is probably more suitable than the humongous dependencies we get with opencv.
Or perhaps we should use gstreamer?

See also: MacOS (#1231), capturing video (#1627), MS Windows (#3336).

totaam:
We can do `libcamera` via gstreamer!

totaam:
WIP.
For lack of a better place, here's a bit of code that saves a VGA preview of my webcam to file:

```py
import os
import mmap
import time
import selectors

from libcamera import (
    Camera, CameraManager, CameraConfiguration,
    StreamRole, Request, PixelFormat, Size,
    FrameBuffer, FrameBufferAllocator, FrameMetadata,
)

cm = CameraManager.singleton()


class MappedFrameBuffer:
    """
    Provides memoryviews for the FrameBuffer's planes
    """
    def __init__(self, fb: FrameBuffer):
        self.__fb = fb
        self.__planes = ()
        self.__maps = ()

    def __enter__(self):
        return self.mmap()

    def __exit__(self, exc_type, exc_value, exc_traceback):
        self.munmap()

    def mmap(self):
        if self.__planes:
            raise RuntimeError('MappedFrameBuffer already mmapped')
        fb = self.__fb

        # Collect information about the buffers

        bufinfos = {}
        for plane in fb.planes:
            fd = plane.fd

            if fd not in bufinfos:
                buflen = os.lseek(fd, 0, os.SEEK_END)
                bufinfos[fd] = {'maplen': 0, 'buflen': buflen}
            else:
                buflen = bufinfos[fd]['buflen']

            if plane.offset > buflen or plane.offset + plane.length > buflen:
                raise RuntimeError(f'plane is out of buffer: buffer length={buflen}, ' +
                                   f'plane offset={plane.offset}, plane length={plane.length}')

            bufinfos[fd]['maplen'] = max(bufinfos[fd]['maplen'], plane.offset + plane.length)

        # mmap the buffers
        maps = []
        for fd, info in bufinfos.items():
            map = mmap.mmap(fd, info['maplen'], mmap.MAP_SHARED, mmap.PROT_READ | mmap.PROT_WRITE)
            info['map'] = map
            maps.append(map)

        self.__maps = tuple(maps)

        # Create memoryviews for the planes
        planes = []
[...]

Issue #3336, opened 2021-11-01
MS Windows native webcam capture


totaam:
Split from #1231.
The opencv version breaks regularly and requires a huge list of dependencies.
It should be possible to capture frames with a fairly simple cython wrapper.

Perhaps something similar to:
[DirectShow Virtual Video Capture Source Filter in C#](https://www.codeproject.com/Articles/437617/DirectShow-Virtual-Video-Capture-Source-Filter-in)

totaam:
Removing this huge list of unused dependencies will also avoid vulnerabilities in those libraries, such as [CVE-2023-51257](https://www.cvedetails.com/cve/CVE-2023-51257/)
(we are not vulnerable in any meaningful way - but rather than having to explain why, just avoid the issue altogether)

totaam:
Works well, even with virtualbox USB passthrough!
Can be tested with:
`Xpra_cmd.exe webcam -d webcam`

Issue #1627, opened 2017-08-17
webcam video support


totaam:
Currently the [Webcam](../wiki/Webcam) uses jpeg for each frame, but there is no particular reason why we can't use video encodings or even pass-through:
* capture raw H264 from the webcam if available
* re-inject this directly into the loopback device (requires this patch: [https://github.com/umlaeute/v4l2loopback/pull/142/commits/8d7a6bc766261276a7cb780fe2f4ab3ac847f1f9] for HEVC and VP9)

See also #1231

Issue #1231, opened 2016-06-16
MacOS webcam capture


totaam:
A native port should solve #1113 on OSX (also blocked by #840) and will allow us to improve platform support and remove the dependency on opencv.

Also remove the hard-coded csc dependency.

PR #4015, opened 2023-09-26
Reverse the webcam aspect ratio in webcam-start message

graph-inc:
The OpenCV library, expects to retrieve `size = (width, height)` tuple, for example, when resizing a frame. However, it reports `(height, width)` when reporting the frame size as a NumPy array. Here is the [relevant quotation](https://learnopencv.com/reading-and-writing-videos-using-opencv/):

    If you are using the NumPy shape method to retrieve frame size,
    remember to reverse the output as OpenCV will return height x width x channels.

This reversed aspect ratio reporting subtlety was correctly handled in the `webcam-frame` messages, however, it was missed in
the `webcam-start` message.


totaam:
The `opencv` webcam support is going to be replaced with something better and more lightweight: #3657, #1231, #1627, #3336

PR #4019, opened 2023-10-01
qrcode: gracefully fail during menu creation

aerusso:
If xpra is built without `qrencode.h` available, `xpra/net/qrcode` is not copied over because `qrencode_ENABLED` is set to false (line 2266 of `setup.py`).

In the modified file, `from xpra.net.qrcode.gtk_qr import show_qr` will then fail, causing `xpra/client/gtk3/menu_helper.py::MenuHelper::build` to fail with `Error: failed to setup menu`.  The symptom is a missing tray icon, and while the best fix is may be to build with qrencode available, some people may not want that enabled.

The minimal fix below moves the failsafe logic before the first import, which is presumably what was intended anyway.

### DIFF
--- a/xpra/client/gtk3/menu_helper.py
+++ b/xpra/client/gtk3/menu_helper.py
@@ -452,16 +452,16 @@ class MenuHelper:
 
 
     def make_qrmenuitem(self) -> Gtk.ImageMenuItem:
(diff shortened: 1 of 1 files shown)


totaam:
Thanks.
Merged and applied to v5.x

PR #4033, opened 2023-10-15
Minor man page fixes

aerusso:
This is a composite of several minor fixes:

 - dashes in command line arguments are mangled by man, and have to be escaped to be searchable.  This is accomplished by the VIM command:

```
:%s/--\([a-z]\+\)\(-[a-z]\+\)*/\=substitute(submatch(0),'-','\\-','g')/g
```

 - Some other dashes were manually identified and fixed
 - A URL suffered from the mangled dash as well.  It is also now bold
 - A grammatical error was corrected
 - A font selection `\f` was used without a font, so it ate up a character to be displayed.  Some more `\fI`-ed text was added around there, as well.

If you want, I can separate these out into different commits.


totaam:
Thanks!

PR #4059, opened 2023-11-22
setup.py FreeBSD tweaks

arrowd:


setup.py
```diff
@@ -234,7 +234,7 @@ def has_header_file(name, isdir=False):
 gtk_x11_ENABLED = DEFAULT and not WIN32 and not OSX
 gtk3_ENABLED = DEFAULT and client_ENABLED
 opengl_ENABLED = DEFAULT and client_ENABLED
-pam_ENABLED = DEFAULT and (server_ENABLED or proxy_ENABLED) and POSIX and not OSX and (find_header_file("/security", isdir=True) or pkg_config_ok("--exists", "pam", "pam_misc"))
+pam_ENABLED = DEFAULT and (server_ENABLED or proxy_ENABLED) and LINUX and not OSX and (find_header_file("/security", isdir=True) or pkg_config_ok("--exists", "pam", "pam_misc"))
```
totaam, line 237: `LINUX and not OSX` is redundant here.
Please use `LINUX` alone.

arrowd: Yep, that was rather silly of me.

PR #4185, opened 2024-04-04
Adding keycloak validation of authorization groups support

iDmple:
Motivation: previously a user could log in with the OIDC client as long as they had a valid token. This PR adds support for validating the keycloak groups we want to permit authentication for this XPRA session.

- `claim_field` is the field that contains the `groups_claim` (groups that the user belongs to)
- `auth_groups` are the groups we want to permit authentication in this XPRA session
- `auth_condition` is the authentication condition

If `auth_condition` is equal to `and`, the user must belong to all groups defined in  `auth_groups` for the authentication to succeed.
If `auth_condition` is equal to `or`, the user belonging to one of the `auth_groups` is sufficient for the authentication to succeed.

If `claim_field` is not defined or empty, the previous functionality is retained and the checks above are not performed.

Independently, I noticed there was a log error because of the %s used in the log function of `xpra/server/auth/sys_auth_base.py`, this was also fixed.

PR #4186, opened 2024-04-04
Adding keycloak validation of authentication groups support

iDmple:
Follow up of #4185 after squashing.

### DIFF
--- a/xpra/server/auth/keycloak.py
+++ b/xpra/server/auth/keycloak.py
@@ -15,6 +15,9 @@ KEYCLOAK_REALM_NAME = os.environ.get("XPRA_KEYCLOAK_REALM_NAME", "example_realm"
 KEYCLOAK_CLIENT_ID = os.environ.get("XPRA_KEYCLOAK_CLIENT_ID", "example_client")
 KEYCLOAK_CLIENT_SECRET_KEY = os.environ.get("XPRA_KEYCLOAK_CLIENT_SECRET_KEY", "secret")
 KEYCLOAK_REDIRECT_URI = os.environ.get("XPRA_KEYCLOAK_REDIRECT_URI", "http://localhost/login/")
+KEYCLOAK_CLAIM_FIELD = os.environ.get("XPRA_KEYCLOAK_CLAIM_FIELD", "")  # field containing groups claim
+KEYCLOAK_AUTH_GROUPS = os.environ.get("XPRA_KEYCLOAK_AUTH_GROUPS", "")  # authorized groups
+KEYCLOAK_AUTH_CONDITION = os.environ.get("XPRA_KEYCLOAK_AUTH_CONDITION", "and")  # authentication condition
 KEYCLOAK_SCOPE = os.environ.get("XPRA_KEYCLOAK_SCOPE", "openid")
 KEYCLOAK_GRANT_TYPE = os.environ.get("XPRA_KEYCLOAK_GRANT_TYPE", "authorization_code")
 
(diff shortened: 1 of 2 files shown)


xpra/server/auth/keycloak.py
```diff
@@ -158,9 +164,37 @@ def check(self, response_json: bytes) -> bool:
...
+                    log.error("Error: keycloak authentication failed as auth_groups is invalid")
+                    return False
+
+                def get(data_dict, keys):
+                    return get(data_dict[keys.pop(0)], keys) if len(keys) else data_dict
+
+                groups_claim = get(user_info, self.claim_field.split('.'))
+                log("groups_claim: %r", groups_claim)
+                if self.auth_condition == "or":
+                    if len(self.auth_groups.intersection(set(groups_claim))) == 0:
```
totaam, line 186: How can that work since `self.auth_groups` is a list?

totaam: Also, `len( whatever ) == 0` is better written as `not whatever`

xpra/server/auth/keycloak.py
```diff
@@ -158,9 +164,37 @@ def check(self, response_json: bytes) -> bool:
...
+            log("claim_field: %r", self.claim_field)
+            log("auth_groups: %r", self.auth_groups)
+            # if claim_field is defined, check that the auth_groups match the groups_claim condition
+            if self.claim_field:
+                if not self.auth_groups:
+                    log.error("Error: keycloak authentication failed as auth_groups is invalid")
+                    return False
+
+                def get(data_dict, keys):
+                    return get(data_dict[keys.pop(0)], keys) if len(keys) else data_dict
```
totaam, line 181: It's still not clear what that does.
Please include docstrings with an example to make it easier for someone reading this code.

xpra/server/auth/keycloak.py
```diff
@@ -158,9 +164,37 @@ def check(self, response_json: bytes) -> bool:
...
+
+                groups_claim = get(user_info, self.claim_field.split('.'))
+                log("groups_claim: %r", groups_claim)
```
totaam, line 184: `groups_claim` is always used as a `set(groups_claim)` so perhaps this should be done here, only once.
What type of value does `get` return? Is it guaranteed to always return an `Iterable`?

iDmple: It will return a list of groups.

xpra/server/auth/keycloak.py
```diff
@@ -27,6 +30,9 @@ def __init__(self, **kwargs):
         self.client_id = kwargs.pop("client_id", KEYCLOAK_CLIENT_ID)
         self.client_secret_key = kwargs.pop("client_secret_key", KEYCLOAK_CLIENT_SECRET_KEY)
         self.redirect_uri = kwargs.pop("redirect_uri", KEYCLOAK_REDIRECT_URI)
+        self.claim_field = kwargs.pop("claim_field", KEYCLOAK_CLAIM_FIELD)
```
totaam, line 33: Does this variable only drive the auth group check?
If so, wouldn't it be easier to remove it and check for `auth_condition` being set?

xpra/server/auth/keycloak.py
```diff
@@ -27,6 +30,9 @@ def __init__(self, **kwargs):
...
+        self.auth_groups = kwargs.pop("auth_groups", KEYCLOAK_AUTH_GROUPS).split()
```
totaam, line 34: It could be worth doing: `self.auth_groups = set(...)` here instead of later. No risk either as `split()` always returns a list.

totaam:
LGTM, can you squash the commits?

FWIW: I usually stash my changes, `git reset HEAD~` then redo the commit(s), finally `git push --force`.

xpra/server/auth/keycloak.py
iDmple: No, this is used to store the list of keys used to retrieve the groups claim within userinfo, this is done with `extract_groups_claim()`.

iDmple:
> LGTM, can you squash the commits?
> 
> FWIW: I usually stash my changes, `git reset HEAD~` then redo the commit(s), finally `git push --force`.

This didn't seem to work. I have to go for a bit, I will try to find a way when I come back!!
Thanks a lot for reviewing the PR!

totaam:
Applied. Thanks!

This will be considered for merging into LTS 5.1.x after 6.0 is released.

PR #4216, opened 2024-05-07
Add compatibility with FFMPEG 7.0

eclipseo:
channel_layout has been replaced with ch_layout

avcodec_close has been deprecated in favor of avcodec_free_context

avio: Constify data pointees of write callbacks

FF_API_FRAME_PICTURE_NUMBER has been deprecated

AV_PIX_FMT_XVMC has been deprecated

This is just a proposal. I didn't find an easy way to put condition on the FFmpeg version in the Cython code.

The most recent incompatible feature would be "avio: Constify data pointees of write callbacks" which is available for LIBAVFORMAT_VERSION_MAJOR >= 61, so FFmpeg v7. 

As such merging that would break compilation with previous FFmpeg. Maybe we could add a decoder_v7.pyx and encoder_v7.pyx, and then find out FFmpeg version in setup.py and compile the proper files.

### DIFF
--- a/xpra/codecs/ffmpeg/decoder.pyx
+++ b/xpra/codecs/ffmpeg/decoder.pyx
@@ -203,7 +203,6 @@ cdef extern from "libavutil/pixfmt.h":
     AVPixelFormat AV_PIX_FMT_BAYER_GBRG16BE
     AVPixelFormat AV_PIX_FMT_BAYER_GRBG16LE
     AVPixelFormat AV_PIX_FMT_BAYER_GRBG16BE
-    AVPixelFormat AV_PIX_FMT_XVMC
     AVPixelFormat AV_PIX_FMT_YUV440P10LE
     AVPixelFormat AV_PIX_FMT_YUV440P10BE
     AVPixelFormat AV_PIX_FMT_YUV440P12LE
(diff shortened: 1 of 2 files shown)


totaam:
Some things can be merged as is, ie: removing `AV_PIX_FMT_XVMC`.
Otherwise, the differences can be hidden away in a compat header file, just like:
https://github.com/Xpra-org/xpra/blob/v5.x/xpra/codecs/ffmpeg/register_compat.c

What is the replacement for:
```
self.av_frame.coded_picture_number = self.frames+1
self.av_frame.display_picture_number = self.frames+1
```

eclipseo:
> Some things can be merged as is, ie: removing `AV_PIX_FMT_XVMC`. Otherwise, the differences can be hidden away in a compat header file, just like: https://github.com/Xpra-org/xpra/blob/v5.x/xpra/codecs/ffmpeg/register_compat.c
> 
> What is the replacement for:
> 
> ```
> self.av_frame.coded_picture_number = self.frames+1
> self.av_frame.display_picture_number = self.frames+1
> ```

I haven't seen any replacement for these. They were removed and no mention of a replacement was mentioned, even when they were just deprecated.

eclipseo:
What is the minimal FFmpeg version you want to support?

totaam:
For each supported branch, the minimal version shipped by any of the supported distro: https://github.com/Xpra-org/xpra/wiki/Download#-linux
In practical terms, that means Ubuntu 20.04 / RHEL8 for v5.x LTS and Ubuntu 22.04 / RHEL8 for v6 + git master.

totaam:
Not heard back.

totaam:
FWIW we no longer use ffmpeg directly: 20bb5f04233dc650022bc67d5904566d1b158af9

PR #4257, opened 2024-06-16
Use nvidia-smi to check for GPU in a containerized environment

EmanH:


totaam:
Thanks.

There are a lot of callers. `has_nvidia_hardware` is used in the nvidia and gstreamer codecs.
This change potentially changes how long this function takes to run from milliseconds to... a lot more, because it shells out.
I don't think it would help much to introduce a Lock, and the codecs are loaded sequentially anyway.
And even if we end up running `nvidia-smi` twice, I don't see a problem.

totaam:
You forgot to add an import statement, the commit above fixes that.

PR #4260, opened 2024-06-20
A GNOME Shell extension for input sources management

graph-inc:
Previously, we used `org.gnome.Shell.Eval` D-Bus method in order to list and activate input sources, when it was required to update the Linux platform keyboard layout. However, that method is not accessible since GNOME Shell v41.0 due to [this commit](https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/a628bbc4).

This pull request installs a new GNOME Shell extension, namely `input-source-manager@xpra_org`, which exposes the `xpra_org.InputSourceManager.List` and `xpra_org.InputSourceManager.Activate` D-Bus methods and calls them instead of the internal `Eval` method.

### DIFF
--- a/packaging/debian/xpra/control
+++ b/packaging/debian/xpra/control
@@ -219,13 +219,22 @@ Suggests: xpra-codecs-nvidia (= ${binary:Version})
         ,python-psutil
 # webcam support:
         ,python3-opencv
-# to make the system tray work again with gnome-shell:
-        ,gnome-shell-extension-top-icons-plus
-        ,gnome-shell-extension-appindicator
+        ,xpra-client-gnome
 # password prompts:
         ,pinentry-gnome3
 Description: tool to detach/reattach running X programs,
  this is the xpra client user interface using GTK3
+
+
+Package: xpra-client-gnome
+Architecture: any
+Depends: xpra-client-gtk3 (= ${binary:Version})
+        ,gnome-shell-extension-top-icons-plus
+        ,gnome-shell-extension-appindicator
+Description: This package installs the GNOME Shell extensions
+ that can help in restoring the system tray functionality.
+ It also includes the input-source-manager@xpra_org extension which
+ is required for querying and activating keyboard input sources.
 
 
 Package: xpra-audio
--- a/packaging/rpm/xpra.spec
+++ b/packaging/rpm/xpra.spec
@@ -38,6 +38,8 @@ autoprov: no
 %define DEFAULT_BUILD_ARGS --with-Xdummy --without-Xdummy_wrapper --without-csc_cython --without-evdi --without-cuda_rebuild
 %endif
 
+%global gnome_shell_extension input-source-manager@xpra_org
+
 %{!?nthreads: %global nthreads %(nproc)}
 %{!?update_firewall: %define update_firewall 1}
 %{!?run_tests: %define run_tests 0}
@@ -343,8 +345,10 @@ Requires:			gnome-shell-extension-appindicator
 Requires(post):		gnome-shell-extension-common
 %endif
 %description -n%{package_prefix}-client-gnome
-This meta package installs the gnome shell extensions
+This package installs the GNOME Shell extensions
 that can help in restoring the system tray functionality.
+It also includes the %{gnome_shell_extension} extension which
+is required for querying and activating keyboard input sources.
 
 
 %package -n %{package_prefix}-x11
@@ -536,7 +540,10 @@ rm -rf $RPM_BUILD_ROOT
 #meta package
 
 %files -n %{package_prefix}-client-gnome
-#meta package without any files
+%{_datadir}/gnome-shell/extensions/%{gnome_shell_extension}/COPYING
+%{_datadir}/gnome-shell/extensions/%{gnome_shell_extension}/README.md
+%{_datadir}/gnome-shell/extensions/%{gnome_shell_extension}/extension.js
+%{_datadir}/gnome-shell/extensions/%{gnome_shell_extension}/metadata.json
 
 %files -n xpra-filesystem
 %defattr(-,root,root)
@@ -839,8 +846,10 @@ for uid in `ls /run/user/`; do
     if [ -S "${BUS}" ]; then
 %if 0%{?el8}
 sudo -i -u "#$uid" DBUS_SESSION_BUS_ADDRESS="unix:path=$BUS" gnome-shell-extension-tool -e TopIcons@phocean.net  &>/dev/null || :
+sudo -i -u "#$uid" DBUS_SESSION_BUS_ADDRESS="unix:path=$BUS" gnome-shell-extension-tool -e %{gnome_shell_extension}  &>/dev/null || :
 %else
 sudo -i -u "#$uid" DBUS_SESSION_BUS_ADDRESS="unix:path=$BUS" gnome-extensions enable appindicatorsupport@rgcjonas.gmail.com  &>/dev/null || :
+sudo -i -u "#$uid" DBUS_SESSION_BUS_ADDRESS="unix:path=$BUS" gnome-extensions enable %{gnome_shell_extension}  &>/dev/null || :
 %endif
     fi
 done
--- a/setup.py
+++ b/setup.py
@@ -282,6 +282,7 @@ uinput_ENABLED = x11_ENABLED
 dbus_ENABLED = DEFAULT and (x11_ENABLED or WIN32) and not OSX
 gtk_x11_ENABLED = DEFAULT and not WIN32 and not OSX
 gtk3_ENABLED = DEFAULT and client_ENABLED
+ism_ext_ENABLED = DEFAULT and gtk3_ENABLED and data_ENABLED
 opengl_ENABLED = DEFAULT and client_ENABLED
 has_pam_headers = has_header_file("/security", isdir=True) or pkg_config_ok("--exists", "pam", "pam_misc")
 pam_ENABLED = DEFAULT and (server_ENABLED or proxy_ENABLED) and LINUX and has_pam_headers
@@ -437,6 +438,7 @@ SWITCHES += [
     "server", "client", "dbus", "x11", "xinput", "uinput", "sd_listen",
     "gtk_x11", "service",
     "gtk3", "example",
+    "ism_ext",
     "pam", "xdg_open",
     "audio", "opengl", "printing", "webcam", "notifications", "keyboard",
     "rebuild",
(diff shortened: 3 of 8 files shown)


README.md
```diff
@@ -52,6 +52,14 @@ To contribute to the project, please try to use pull-requests and follow our [co
...
 [![xpra](https://github.com/Xpra-org/xpra/actions/workflows/test.yml/badge.svg)](https://github.com/Xpra-org/xpra/actions/workflows/test.yml)
 
+## GNOME Shell Extension
+
+The `input-source-manager@xpra_org` GNOME Shell extension (see `gnome_shell_extension/` directory)
+must be installed if it is desired to update the platform keyboard layout when the `next_keyboard_layout` is called.
+This extension may be installed by `/usr/bin/gnome-extensions enable input-source-manager@xpra_org` command
+or alternatively, it may be enabled manually by searching for the `Extensions` (by pressing the `Super` key)
+and finding that extension in the filtered list.
+
```
totaam, line 62: This does not belong on the home page!

gnome_shell_extension/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+If the extension is installed and enabled, above methods may be tested
+from CLI using the `gdbus` program as follows:
+
+```
+# List available input sources
+gdbus call --session --dest org.gnome.Shell --object-path /org/xpra/InputSourceManager --method xpra_org.InputSourceManager.List
+# Activate the first available input source (with index 0)
+gdbus call --session --dest org.gnome.Shell --object-path /org/xpra/InputSourceManager --method xpra_org.InputSourceManager.Activate 0
+```
```
totaam, line 25: The gnome shell extension should be placed somewhere in `fs/share/gnome-shell/extensions/input-source-manager@xpra_org/`, not at the top level.

gnome_shell_extension/README.md
```diff
@@ -0,0 +1,25 @@
+# Input Source Manager (ISM) Extension for GNOME Shell
+
+The GNOME Shell has restricted D-Bus interface access to some of its
+private APIs since [v41.0](https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/a628bbc4)
```
totaam, line 4: Is there any point in trying so hard to support gnome shell if they keep restricting their pseudo-APIs to their own applications only?
If Microsoft did this sort of thing, they would get sued.

gnome_shell_extension/metadata.json
```diff
@@ -0,0 +1,17 @@
...
+    "40",
+    "41",
+    "42",
+    "43",
+    "44"
+  ],
+  "url": "https://github.com/Xpra-org/xpra/",
+  "uuid": "input-source-manager@xpra_org",
+  "version": 1
+}
```
totaam, line 17: And this extension will have to be maintained and updated every time they break something or add a new release?
(similar to the churn I see in other very useful extensions)
Oh, joy.

packaging/debian/xpra/control
```diff
@@ -214,6 +214,7 @@ Recommends: xpra-codecs (= ${binary:Version})
 # hope that this will allow our status icon to be shown:
         ,gir1.2-appindicator3-0.1
         ,gir1.2-ayatanaappindicator3-0.1
+        ,gnome-shell
```
totaam, line 217: This is not acceptable, we cannot depend on `gnome-shell`.
Xpra is DE agnostic.
Ideally this would live in another package that users can choose to install to support gnome shell.
This may warrant adding a `xpra-client-gnome` package.

packaging/rpm/xpra.spec
```diff
@@ -317,6 +319,7 @@ Recommends:			%{package_prefix}-codecs = %{version}-%{release}
 Recommends:			%{package_prefix}-x11 = %{version}-%{release}
 Recommends:			pinentry
 Recommends:			%{package_prefix}-audio
+Recommends:			gnome-shell
```
totaam, line 322: Just like above, we cannot recommend `gnome-shell`.

packaging/rpm/xpra.spec
```diff
@@ -828,6 +835,7 @@ fi
 %post -n %{package_prefix}-client-gtk3
 /usr/bin/update-desktop-database &> /dev/null || :
 /bin/touch --no-create %{_datadir}/icons/hicolor &>/dev/null || :
+echo "Enable the '%{gnome_shell_extension}' extension for keyboard layout management"
```
totaam, line 838: Why is this here?

xpra/platform/posix/keyboard.py
```diff
@@ -51,7 +51,7 @@ def init_vars(self) -> None:
             self.__dbus_helper = None
             log.info(f"gnome input sources requires dbus: {e}")
         else:
-            self._dbus_gnome_shell_eval_ism(".inputSources", self._store_input_sources)
+            self._dbus_gnome_shell_ism("List", [], self._store_input_sources)
```
totaam, line 54: This breaks backwards compatibility for any installation that has not been update to gnome-shell 41, doesn't it?

xpra/platform/posix/keyboard.py
```diff
@@ -93,7 +92,7 @@ def set_platform_layout(self, layout: str) -> None:
         if index is None:
             log(f"asked layout ({layout}) has no corresponding registered input source")
             return
-        self._dbus_gnome_shell_eval_ism(f".inputSources[{index}].activate()")
+        self._dbus_gnome_shell_ism("Activate", [native_to_dbus(index)])
```
totaam, line 95: A better approach to these chunks would be to test which interface is available and then choose the most appropriate one.

totaam (changes requested):

gnome_shell_extension/README.md
graph-inc: They have restricted the **Eval** API for security reasons as it allows arbitrary codes to be executed in context of another process. It is a bit similar to the `--start-new-commands` option of Xpra itself.
By installing our extension, end-users become eligible to enable/disable our extension.
An alternative method for end-users is to put gnome-shell in "unsafe" mode which enables access to **Eval** method again.

packaging/rpm/xpra.spec
graph-inc: It's just a reminder. The extension should be enabled manually because all newly installed extensions are disabled by default. When the extension is disabled, Xpra will log an error for the D-Bus query and moves on (similar to the now restricted **Eval** method in the **safe** mode).

And for the same reason, I thought that following explanation in the README.md file may be useful for end-users:
> This extension may be enabled by `/usr/bin/gnome-extensions enable input-source-manager@xpra_org` command or alternatively, it may be enabled manually by searching for the `Extensions` (by pressing the `Super` key) and finding that extension in the filtered list.

xpra/platform/posix/keyboard.py
graph-inc: No, it is backward-compatible. I tested it on `v3.32.2` too.
In older versions, both methods (**Eval** and our new extension) will work equally.
In newer versions, only the extension works.

README.md
graph-inc: I removed this part and integrated its information with the README.md file which exists in `fs/share/gnome-shell/extensions/input-source-manager@xpra_org/` directory.

gnome_shell_extension/README.md
graph-inc: It is now moved there and `setup.py` is updated accordingly.

xpra/platform/posix/keyboard.py
graph-inc: Good idea. Updated. Now, it falls back to the **Eval** method and works as while as one of these conditions is true:

1. new extension is enabled by end-user,
2. the GNOME Shell is running in **unsafe** mode,
3. the GNOME Shell is older than v41.0.

totaam: I must be missing something because I can't see where it does the fallback.

gnome_shell_extension/README.md
graph-inc: I agree that exposing the list of input sources is not a security risk. Their point is that **Eval** method which allows running arbitrary codes is restricted.
Although we used **Eval** method in order to get the list of `.inputSources`, but others could use **Eval** for shutting down the computer :smile:  like this `imports.misc.gnomeSession.SessionManager().ShutdownRemote()`

xpra/platform/posix/keyboard.py
graph-inc: Code runs asynchronously, sending a D-Bus request and calling one of the `ok_cb` or `err_cb` functions based on the successful or failed invocation. So, if extension is not installed or is not enabled, `err_cb` will be called which will send the fallback D-Bus request. It's around lines 75-85:
```python
        def err_cb(msg) -> None:
            log("_dbus_gnome_shell_ism(method=%s, args=%s): %s", method, args, msg)
            if method == "List":
                self._dbus_gnome_shell_eval_ism(".inputSources", self._store_input_sources)
            elif method == "Activate":
                try:
                    index = dbus_to_native(args[0])
                except Exception:
                    log("_dbus_gnome_shell_ism(method=%s, args=%s): the Eval fallback", method, args, exc_info=True)
                else:
                    self._dbus_gnome_shell_eval_ism(f".inputSources[{index}].activate()")
```

totaam: Ah, thanks, now I see it.

setup.py
```diff
@@ -1563,6 +1563,10 @@ def add_build_info(*args) -> None:
         # ensure it is now included in the module list
         add_modules("xpra.src_info")
 
+if POSIX and gtk3_ENABLED and data_ENABLED:
+    ism_dir = "share/gnome-shell/extensions/input-source-manager@xpra_org"
+    data_files.append((ism_dir, glob(f"fs/{ism_dir}/*")))
+    data_files.append((ism_dir, ["COPYING"]))
```
totaam, line 1570: The extension needs its own switch.

gnome_shell_extension/README.md
totaam: No way!
So for years Wayland desktops had an `Eval()` function accessible to all!? (and some deployed desktops probably still do - RHEL8?)
Where is the CVE for that?
And if it still exists, even restricted to gtk / gnome prefixed apps, this is ripe for exploitation.

graph-inc: Good news is that it was not open to **all local users** because a session D-Bus checks the **effective UID** of the caller and it has to match with the session daemon. Bad news is that **all processes** of the **same local user** can still use it on RHEL8.
I cannot find a CVE for it :thinking:

setup.py
graph-inc: It is replaced by `ism_ext` switch.

packaging/debian/xpra/control
graph-inc: The xpra-client-gnome DEB package is added similar to its RPM counterpart.

totaam:
@graph-inc can you help with #4724 ?

PR #4261, opened 2024-06-21
use --ssh when reconnecting

hxtmdev:
Reconnects did not use the `--ssh` args which can lead to using paramiko and/or missing options.

In my case it prompted for a password on reconnect even though there was a perfectly fine key just waiting to be used.

Reproduce:
- set up ssh with pubkey
- `xpra --ssh=... start`
- wait until connected
- kill ssh process to trigger reconnect (also caused by wifi problems)
- observe password prompt (not anymore with this patch)


totaam:
Thanks!

PR #4276, opened 2024-07-02
fix PDFIUM_Print.exe crash

JuergenGleiss:
* change order of fields within FPDF_LIBRARY_CONFIG(Structure)
* define return of FPDF_LoadPage

resolve #4275

PR #4278, opened 2024-07-03
POINTER(None) is not a valid construct.

JuergenGleiss:
sonarqube static analyzer

Change this argument; Function "POINTER" expects a different type Arguments given to functions should be of an expected type[python:S5655](https://sonarqube-xpra.vpo.nl/coding_rules?open=python%3AS5655&rule_key=python%3AS5655)

fixes #4276


totaam:
Thanks!

PR #4296, opened 2024-07-09
Bugfix: When checking for existing user config, expand tilde to HOME

luke-jr:
Without this, os.path.exists looks for a literal directory named '~' and returns False, causing Xpra to overwrite the existing user config

### DIFF
--- a/xpra/scripts/config.py
+++ b/xpra/scripts/config.py
@@ -473,7 +473,8 @@ def may_create_user_config(xpra_conf_filename:str = DEFAULT_XPRA_CONF_FILENAME):
     if udirs:
         has_user_conf = None
         for d in udirs:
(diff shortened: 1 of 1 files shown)


totaam:
Please use https://github.com/Xpra-org/xpra/blob/d0f325c8981dde8258e9b08a109ffb6275501e96/xpra/util/env.py#L90 instead.
How did this get triggered?

luke-jr:
idk, it always happens every time for me

totaam:
How, what specific command?

luke-jr:
`xpra --help` is enough

totaam:
Done in 11a2239ca89f7457ab0ca91097d777e3ac17f981

PR #4310, opened 2024-07-26
Windows printing: only one page of a multipage document is printed

JuergenGleiss:
Bugfix: not all pages of an multipage pdf document is printed.

### Solution
Before a new page is sent to the printer, prepare the printer to accept date After rendering of a page is finished the page is closed and the printer device is notified. 

### Some links on how I came to my solution
[StartPage](https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-startpage) 
[EndPage](https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-endpage)

[FoxIt PDF SDK User Manual](http://cdn01.foxitsoftware.com/pub/foxit/manual/enu/FoxitPDF_SDK20_Guide.pdf)

### DIFF
--- a/xpra/platform/win32/pdfium.py
+++ b/xpra/platform/win32/pdfium.py
@@ -16,7 +16,7 @@ from ctypes import (
 from xpra.util.str_fn import Ellipsizer, strtobytes
 from xpra.platform.win32.common import GetDeviceCaps
 from xpra.platform.win32 import win32con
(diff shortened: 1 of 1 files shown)


totaam:
How odd, did I never test multi-page printing on MS Windows? Perhaps not!

The changes look fine but have you tested these changes with the libpdfium that we ship with the windows binaries?
You seem to be linking to the non-free `FoxIt` SDK.

PR #4312, opened 2024-07-30
Fix starting stderr_reader as a thread in exec_client.py

alukichev:
When an ssh program is given with "--ssh" option, its stderr may be connected through a pipe. To manage reading the stream, a separate thread is started that has child process (ssh impl) as a parameter. Starting the thread, however, does not provide the corresponding argument, leading to an exception.

This fixes the exception by providing required argument.


totaam:
Thanks!

PR #4316, opened 2024-08-03
README.md: fix download url

khipp:
This PR fixes the download URL for the macOS PKG installer package and further adds language identifiers to the fenced code blocks to enable syntax highlighting.


totaam:
Thanks!

PR #4340, opened 2024-08-31
Fix --keyboard-sync command line option

unisons:
The "sync" is not in prop but prop['keymap']

Check out the code below:
https://github.com/Xpra-org/xpra/blob/b16a48f6137e08f43fc38b4d96d16d49b3392ed7/xpra/server/mixins/input.py#L58


totaam:
Thanks. 
I thought this had just been fixed a couple of weeks ago, but clearly not.

PR #4381, opened 2024-10-10
docs: update freebsd dependency packages

spmzt:
Binary packages available from the official FreeBSD Package repositories are built using a [DEFAULT_VERSION of Python](https://cgit.freebsd.org/ports/tree/Mk/bsd.default-versions.mk#n140), which evolves over time.


totaam (approved):

totaam:
Strange how the default version doesn't have a short alias like `py3-XXXX`, which would make things so much easier.

Issue #4503, opened 2025-02-13
SSH connection host should strip comments


timsu92:
**Describe the bug**
If there's a comment in SSH's config file, xpra connections using host aliases will also include the part as SSH command.
Normally, say just use SSH to connect to another host, comments in `Hostname` line will be ignored and connected. 

**To Reproduce**
1. Start server in any way
2. Setup `%userprofile%\.ssh\config` to have comments in `Hostname` line
   ```
   Host myhost
       Hostname <address-here>  # put some comments behind
       Port 12345
   ```
3. Connect to the server like `Xpra_cmd attach ssh://myhost/1`
4. See the error message:
   ```
   Warning: failed to connect: 
    cannot get any address for <address-here>  # put some comments behind:12345 : [Errno 11001] getaddrinfo failed
   ```
**System Information (please complete the following information):**
 - Server OS: Ubuntu 24.04.2 LTS
 - Client OS: Windows 11
 - Xpra Server Version 6.2.3-r0
 - Xpra Client Version 6.2.3-r1

**Additional context**
Client version:
![client version](https://github.com/user-attachments/assets/433ca062-ea7c-499c-b535-cd7faf55b922)

totaam:
This should be done in paramiko, but seeing that a one line changed has not been merged in two years: https://github.com/Xpra-org/xpra/issues/3868#issuecomment-1643612561
I don't expect this will be fixed upstream anytime soon.

Please try cab89fb18724096bee22c5c3d95e82f7862f9292

timsu92:
I tried to build for Windows but failed😢
However, I have copied out the `safeget` function and tested on both `127.0.0.1` and `127.0.0.1 # comments`.

Do you think that I should try to build client on Ubuntu?

totaam:
> and I found [cx-freeze](https://packages.msys2.org/packages/mingw-w64-x86_64-python-cx-freeze) package instead of cx_Freeze, so I changed it

Thanks!
Fixed: dc216ad0e1ee6e0d7731fa6ab1abed61c64c65c1

> AttributeError: 'NoneType' object has no attribute 'group'

Ah, that's a bug, the parsing failure should not have been fatal.
It is just failing to identify which version of MS Windows you are running.
What does `systeminfo` print when you run it from a prompt?

a44631dd109164b9e78b1b77da98021b5b10ef5d + aef04a39a7a0b59db2dacb869990d9e186f3b532 should make it more resilient.

timsu92:
[...]

PR #4507, opened 2025-02-15
Fix windows systeminfo codepage when setting up

timsu92:
This is a byproduct when dealing with the problem in https://github.com/Xpra-org/xpra/issues/4503#issuecomment-2659339897

In the last discussion, we found that when running `python3 packaging/MSWindows/BUILD.py`, there are problems when calling `systeminfo`:
1. The titles (like "OS Name") are not in English if user's computer is in other language
2. I got text's encoding in `cp950` since I'm using a computer in Traditional Chinese
3. Output of "OS Name"("作業系統名稱" in my case) contains non-English contents

For the first two problems, I've solved in this PR and currently the output is something like "Microsoft Windows 10 企業版 LTSC".
I don't see the string being used other than showing in build info, so I guess texts not in English here works, as long as it's utf-8?

---

Meanwhile, there's still a little got-cha regarding `cx-freeze`😉
Fixed in this PR as well

### DIFF
--- a/fs/bin/add_build_info.py
+++ b/fs/bin/add_build_info.py
@@ -9,13 +9,12 @@
 
 import datetime
 from shutil import which
-from subprocess import Popen, PIPE, STDOUT, getstatusoutput
+from subprocess import Popen, PIPE, STDOUT, getstatusoutput, run
 from typing import Any
 from collections.abc import Sequence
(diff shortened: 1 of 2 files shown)


totaam:
Thanks, will test it later.

PR #4532, opened 2025-03-17
Bugfix: Include pam_*.h within the security/ subdirectory

luke-jr:
Including these headers was never supposed to work without the security/ subdirectory in the include directive, and PAM 1.7.0 (specifically [941bfc1de4c8d4963c8b9d63d32fed391f1476c8](https://github.com/linux-pam/linux-pam/commit/941bfc1de4c8d4963c8b9d63d32fed391f1476c8)) fixed the pkgconfig files so it no longer works without that

### DIFF
--- a/xpra/platform/pam.pyx
+++ b/xpra/platform/pam.pyx
@@ -32,12 +32,12 @@ cdef extern from "string.h":
 cdef extern from "stdlib.h":
     void *calloc(size_t nitems, size_t size)
 
-cdef extern from "pam_misc.h":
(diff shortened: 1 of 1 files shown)


totaam:
IMNSHO, if they're going to break 10+ years worth of C examples, they should include a message in the pkgconfig file explaining what to do.
Now I can revert ff44b3fef6a3507c619b0f68c70d6b03a5055dd2, and all its backports.

totaam:
Looks like this broke Debian: #4568

PR #4600, opened 2025-06-04
fix non-standard port host key verification

### LINKED ISSUE #4599: SSH hostkey verification fails on non-standard ports
**Describe the bug**
When I try to connect to a host via ssh on a non standard port, then the host key verification fails.
The reason for the failure is that in the file `known_hosts`, the format is
```
[the-hostname]:1234 ssh-ed25519 AAAAC<rest-of-fingerprint>
```
instead of the standard entry
```
the-hostname ssh-ed25519 AAAAC<rest-of-fingerprint>
```

**To Reproduce**
1. Connect with `ssh the-host-config` first, and make sure the entry is in the file `known_hosts`
2. Connect with `xpra ssh://the-host-config ...`

finkandreas:
Fixes #4599


totaam:
Thanks!

PR #4645, opened 2025-08-21
Fix apt-get install --no-install-recommends xpra when there's no adduser package

felipecrs:
xpra-server postinst depends on addgroup command, but does not list dependency on adduser package which provides such command.

This issue can be replicated in container, please let me know if you'd like me to provide a reproduction.


totaam:
Thanks!

Issue #2944, opened 2020-11-17
wayland clipboard extra data types


totaam:
#2927 only added support for text, add uris and images.

totaam:
60f6d1efd74a910ee27664f48bbf5395217ef167 already added `image/png`

totaam:
As of today, practically recognized formats - the shared clipboard code knows about:

| Category | MIME/targets |
|---------------|----------------------|
|Plain text |`UTF8_STRING`, `TEXT`, `STRING`, `text/plain;charset=utf-8`, `text/plain` |
|HTML        |`text/html` |
|URI/file list | `text/uri-list` |
|RTF            | `text/rtf, application/rtf` |
|PDF             | `application/pdf` |
| Images   | `image/png`, `image/jpeg`, `image/tiff`

These definitions are in:
https://github.com/Xpra-org/xpra/blob/854ded570974468cb5c78f35ec9b3a2fba8b5b5f/xpra/clipboard/targets.py#L61-L82

Adding more loose matching helps when client / server run completely different OSes with different datatype exepectations.
This will do for now.

PR #4717, opened 2025-11-20
clipboard/gtk: support other targets, like image/png

snyh:
In wayland/xwayland, client will use gtk/clipboard.py which isn't handle the ATOM of `image/png`

This PR add it, so image can be copied and pasted to remote app

### DIFF
--- a/xpra/gtk/clipboard.py
+++ b/xpra/gtk/clipboard.py
@@ -8,7 +8,7 @@ from time import monotonic
 from xpra.util.gobject import n_arg_signal, one_arg_signal
 from xpra.clipboard.common import ClipboardCallback
 from xpra.clipboard.targets import TEXT_TARGETS
-from xpra.clipboard.proxy import ClipboardProxyCore
+from xpra.clipboard.proxy import ClipboardProxyCore, filter_data
 from xpra.clipboard.timeout import ClipboardTimeoutHelper
 from xpra.os_util import gi_import
 from xpra.util.str_fn import Ellipsizer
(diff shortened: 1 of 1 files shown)


xpra/gtk/clipboard.py
```diff
@@ -155,8 +155,14 @@ def get_contents(self, target: str, got_contents: ClipboardCallback, time=0) ->
                 got_contents(target, 8, text)
                 return
         else:
-            # data = wait_for_contents(target)?
-            pass
+            r = self.clipboard.wait_for_targets()
+            if r and len(r) == 2 and r[0]:
+                atom = next((x for x in r[1] if x.name() == target), None)
+                if atom != None:
+                    sel = self.clipboard.wait_for_contents(atom)
```
totaam: According to the docs, this function can return `None`:
https://lazka.github.io/pgi-docs/Gtk-3.0/classes/Clipboard.html#Gtk.Clipboard.wait_for_contents

totaam:
Slight tangent here, but how hard would it be to hook up the image watermark code here?
Looks trivial based on:
https://github.com/Xpra-org/xpra/blob/17ace255ae129e6fe53aca5d66f1ebf75e9761c6/xpra/platform/win32/ctypes_clipboard.py#L382-L386

FYI: it's not just the watermark, it also sanitizes images:
https://github.com/Xpra-org/xpra/blob/17ace255ae129e6fe53aca5d66f1ebf75e9761c6/xpra/clipboard/proxy.py#L192

snyh:
If target is valid and is image/png or image/jpeg then send it to self.filter_data before invoking got_contents.

Test result: 
 image

xpra/gtk/clipboard.py
```diff
@@ -142,11 +142,16 @@ def do_owner_changed(self) -> None:
...
             r = self.clipboard.wait_for_targets()
             if r and len(r) == 2 and r[0]:
-                targets = r[1]
-                atoms = tuple(x.name() for x in targets)
+                return r[1]
+            return []
+
+        if target == "TARGETS":
+            atoms = tuple(x.name() for x in get_targets())
+            if len(atoms) != 0:
```
totaam: `if atoms:` is sufficient.

totaam:
Merged and then found a minor issue: 965f83b752503524a6104f4ab38628f086fd944b, then fixed the cosmetic one: 8abd01e9fb9106edf434368008f404686442905d

FYI: enable the pre-commit hooks with `pre-commit install` and `ruff` will tell you about these.

snyh:
> Merged and then found a minor issue: [965f83b](https://github.com/Xpra-org/xpra/commit/965f83b752503524a6104f4ab38628f086fd944b), then fixed the cosmetic one: [8abd01e](https://github.com/Xpra-org/xpra/commit/8abd01e9fb9106edf434368008f404686442905d)
> 
> FYI: enable the pre-commit hooks with `pre-commit install` and `ruff` will tell you about these.

I'll keep an eye on the project guidelines moving forward—really love how cool this project is.

totaam:
See also #2944

Issue #4724, opened 2025-12-01
input-source-manager extension outdated for gnome-shell 49


sagitter:
Hi all.

**Describe the bug**
`input-source-manager` extension from `xpra-6.3.6` is too old on Fedora 43+ that currently uses gnome-shell 49
Is it acceptable upgrading the extension?

**To Reproduce**
Installing xpra-client-gnome-6.3.6-1 on Fedora 43 from Fedora repository

<img width="872" height="586" alt="Image" src="https://github.com/user-attachments/assets/aacd3bce-5a11-4cff-9d54-f517f562e682" />

totaam:
> Is it acceptable upgrading the extension?

The extension was added in #4260 to fix something that broke in gnome shell 41.0
I can't say that I'm surprised that gs broke something again, but I have absolutely no idea what broke or how to fix it, sorry.
@graph-inc ?

ndias:
I'm the reported of this problem in Fedora, and toke some time to make a pull request
https://github.com/Xpra-org/xpra/pull/4725
The problem with the extension was first the metadata.json doesn't have the version of the current gnome-shell, second right now gnome-shell is more complain about some things like import things in the midle of the code and no default class, my patch solve this problems and now run and doesn't complain anymore

sagitter:
@ndias @totaam 
Thank you

totaam:
@ndias thanks!
This PR was included in the [6.4 release](https://github.com/Xpra-org/xpra/releases/tag/v6.4)

PR #4725, opened 2025-12-02
Update extension to work with gnome-shell 49

ndias:
This patch resolve the issue described in this bug https://github.com/Xpra-org/xpra/issues/4724


totaam:
Thanks!
That was quick.

PR #4737, opened 2025-12-16
feat: add support for Arch Linux distribution

JakkuSakura:


setup.py
```diff
@@ -594,6 +605,57 @@ def build_package() -> int:
...
+build() {{
+  cd "$srcdir/xpra-$pkgver"
+  python3 ./setup.py build
+}}
+
+package() {{
+  cd "$srcdir/xpra-$pkgver"
+  python3 ./setup.py install --root="$pkgdir" --prefix=/usr --skip-build --optimize=1
+}}
+""".format(pkgver=XPRA_VERSION, tar_name=tar_name, pkg_arch=pkg_arch))
```
totaam: One could make the case for storing this template in https://github.com/Xpra-org/xpra/tree/master/packaging, which would make it easier to find.
Especially since this build file is humongous.

totaam (commented):
LGTM!
If you can answer these minor points, I can merge this.

totaam:
Thanks!

Issue #511, opened 2014-02-04
AMD hardware accelerated h264 encoding: VCE Video Codec Engine


totaam:
[AMD Open-Sources VCE Video Encode Engine Code](http://www.phoronix.com/scan.php?page=news_item&px=MTU5MTc) and [AMD Publishes New Code For Open-Source VCE Video Encode](http://www.phoronix.com/scan.php?page=news_item&px=MTYwMTE)

[initial VCE support](http://lists.freedesktop.org/archives/dri-devel/2014-February/053191.html) based on [OpenMAX](http://www.khronos.org/openmax/)

totaam:
The easier way of supporting this is via #3706

totaam:
That said, on MS Windows (ie: #4418), it may be easier to implement native bindings.
Some links:
* [wikipedia: Video Core Next](https://en.wikipedia.org/wiki/Video_Core_Next)
* [Advanced Media Framework (AMF) SDK](https://github.com/GPUOpen-LibrariesAndSDKs/AMF)

Apparently, my new AMD 5700G is a "Cezanne" part, which means it supports VCN 2.2

totaam:
I've tried again on Fedora because that's a much more comfortable environment for me, so here are some notes:
* `AMD` is part of the proprietary components
* `vainfo` shows:
  ```
  Trying display: wayland
  libva info: VA-API version 1.22.0
  libva info: Trying to open /usr/lib64/dri-nonfree/radeonsi_drv_video.so
  libva info: Trying to open /usr/lib64/dri-freeworld/radeonsi_drv_video.so
  libva info: Trying to open /usr/lib64/dri/radeonsi_drv_video.so
  libva info: Found init function __vaDriverInit_1_22
  libva info: va_openDriver() returns 0
  vainfo: VA-API version: 1.22 (libva 2.22.0)
  vainfo: Driver version: Mesa Gallium driver 24.3.2 for AMD Radeon Graphics (radeonsi, renoir, LLVM 19.1.5, DRM 3.59, 6.12.7-200.fc41.x86_64)
  vainfo: Supported profile and entrypoints
        VAProfileMPEG2Simple            :	VAEntrypointVLD
        VAProfileMPEG2Main              :	VAEntrypointVLD
        VAProfileJPEGBaseline           :	VAEntrypointVLD
        VAProfileVP9Profile0            :	VAEntrypointVLD
        VAProfileVP9Profile2            :	VAEntrypointVLD
[...]

PR #4743, opened 2025-12-19
Fix wrong include path for AMF headers

icedream:
I have noticed that the xpra package on Arch Linux will not build on the latest versions if the AMF headers are installed. The build would fail with the message that it cannot find `core/Result.h`, included from [here](https://github.com/Xpra-org/xpra/blob/master/xpra/codecs/amf/amf.pxd#L36).

I noticed the generated compiler flags include `-I/usr/include/AMF/AMF` - a path that does not exist on my system, however `/usr/include/AMF` does, and within that path `core/Result.h` does exist. The change in this PR seems to fix it up.

I have no idea why `/AMF` would ever need to be explicitly appended here, but maybe I missed something.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2958,7 +2958,7 @@ if amf_ENABLED:
         amf_kwargs = pkgconfig("amf")
     except ValueError:
         amf_kwargs = {
-            "extra_compile_args": "-I" + find_header_file("AMF", isdir=True) + "/AMF",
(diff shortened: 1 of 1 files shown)


totaam:
Thanks.
Merged - looks reasonable to me.
No idea why this was done this way!
Please be aware that I have never managed to get the AMF encoder to work on anything other than MS Windows: https://github.com/Xpra-org/xpra/issues/511#issuecomment-2692195830

PR #4747, opened 2026-01-01
Add .md extension to links referring to local files.

aluaces:
Some internal links in the documentation were failing because absent `.md` extensions.


totaam:
Thanks!

Issue #4384, opened 2024-10-11
server xfce-4.16 desktop WM-resolution fixed at SXGA 1280x1024 (despite --resize-display=yes), for 6.1.3-10.r0.el8 (any > xpra-server-5.0.9-10.r5.el8), rocky-linux-8.9


sto6:
Using xpra repo, I noticed 6.1.x xpra-server now available so tried them. But the desktop does not resize to the client, always remains fixed at 1280x1024. Going backwards I found xpra-server-5.0.9-10.r5.el8 works.

For working resize-display:
```
dnf erase --noautoremove xpra\*
dnf install xpra-server-5.0.9-10.r5.el8 xpra-x11-5.0.9-10.r5.el8
```
Versions that work: `xpra-server-5.0.7-10.r0.el8, xpra-server-5.0.9-10.r5.el8`
Versions that fail: `xpra-server-5.0.10-10.r0.el8, xpra-server-6.1.2-10.r1.el8, xpra-server-6.1.3-10.r0.el8`

Server-start command:
```
xpra start-desktop :33 --exit-with-children=yes --start-child=startxfce4 --resize-display=yes --desktop-scaling=no --dpi=96 --audio=no
```
Client used: xpra v6.1.2-r1 (macOS, Intel Monterey 12.7.6).
Client attach command:
```
/Applications/Xpra.app/Contents/MacOS/Xpra attach ssh://sto@cdsr0/33 --desktop-fullscreen=yes --desktop-scaling=no --dpi=96 --swap-keys=no --opengl=yes --ssh=ssh --ssh-upgrade=no
```
Contents of client-side `~/.xpra/xpra.conf`:
```
pulseaudio=no
file-transfer=off
speaker=disabled
microphone=disabled
bell=no
printing=no
html=off
webcam=no
chdir=~/Desktop
dpi=96
opengl=yes
mode=ssh
ssh=ssh
desktop-fullscreen=yes
desktop-scaling=no
swap-keys=no
ssh-upgrade=no
```
 - Server OS: Rocky-Linux 8.9, Intel
 - Client OS: macOS, Intel Monterey 12.7.6
 - Xpra Server Version 6.1.3-10.r0.el8 (any newer-than 5.0.9-10.r5.el8).
 - Xpra Client Version v6.1.2-r1

sto6:
[5.0.9-attach-geom.log](https://github.com/user-attachments/files/17344495/5.0.9-attach-geom.log)
[...]

PR #4767, opened 2026-01-23
#4384 fix --resize-display=yes:none not working

extfs:
## Summary

Fixes dynamic display resizing not working with XFCE4 (and similar desktop environments) when using the HTML5 client.

The workaround `--resize-display=yes:none` mentioned in #4384 doesn't actually work due to two bugs. This PR fixes both.

## Problem

When connecting to an XFCE4 desktop session via HTML5 client, resizing the browser window does not resize the remote desktop. The resolution stays stuck at the default (1920x1080).

The expected behavior with `--resize-display=yes:none` is that **no** initial resolution is set, allowing the client to dynamically control the display size.

## Root Cause

Two bugs prevent `--resize-display=yes:none` from working:

### 1. IndexError in `server.py`

```python
vfb_geom = parse_resolutions(sizes, opts.refresh_rate)[0]
```

When `parse_resolutions("none")` returns `()`, accessing `[0]` crashes:
```
IndexError: tuple index out of range
```

### 2. Empty tuple treated as falsy in `display.py`

```python
self.initial_resolutions = parse_resolutions(...) or self.get_default_initial_res()
```

Since `()` is falsy in Python, `() or default` always returns `default`. So even when `:none` is specified, the 1920x1080 fallback is silently used, preventing dynamic resizing.

## Testing

```bash
xpra start-desktop :100 \
    --bind-tcp=127.0.0.1:15151 \
    --html=on \
    --start=xfce4-session \
    --resize-display=yes:none
```

**Before fix:**
- Server crashes with IndexError, OR
- Resolution stuck at 1920x1080, resizing browser has no effect

**After fix:**
- Server starts without initial resolution
- Resizing browser window dynamically resizes XFCE4 desktop

## Related

- #4384 - Desktop WM-resolution fixed despite `--resize-display=yes`

### DIFF
--- a/xpra/scripts/server.py
+++ b/xpra/scripts/server.py
@@ -1221,7 +1221,9 @@ def do_run_server(script_file: str, cmdline: list[str], error_cb: Callable, opts
                 # "off:1080p" -> "1080p"
                 # "4k" -> "4k"
                 sizes = opts.resize_display.split(":", 1)[-1]
-                vfb_geom = parse_resolutions(sizes, opts.refresh_rate)[0]
+                resolutions = parse_resolutions(sizes, opts.refresh_rate)
+                if resolutions:
+                    vfb_geom = resolutions[0]
             fps = 0
             if opts.refresh_rate:
                 fps = get_refresh_rate_for_value(opts.refresh_rate, 60)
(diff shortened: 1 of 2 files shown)


totaam:
Thanks!

PR #4792, opened 2026-02-25
win32: match title bar theme to Windows dark/light mode setting

andrewachen:
Xpra should respect Windows' dark vs. light mode theme setting.

Read the Windows registry `AppsUseLightTheme` setting and apply `DwmSetWindowAttribute(DWMWA_USE_IMMERSIVE_DARK_MODE)` to each window handle during add_window_hooks(). Controlled by the env var `XPRA_WIN32_NATIVE_TITLE_BAR_THEME` (default: enabled).


totaam:
Very nice!
Thanks.

Issue #921, opened 2015-07-17
native win32 opengl client


totaam:
This may help with testing and with things like #916, #917.

It shouldn't be too hard to make a simple win32 client using pywin32 + opengl.
Doing so would also help to ensure that the separation between the gtk client, ui client class and platform code is clean.

Here's some example code:
* [glglue](https://github.com/ousttrue/glglue): [wgl.py](https://github.com/ousttrue/glglue/blob/master/glglue/wgl.py)
* [pywin32 openGLDemo.py](https://github.com/Travis-Sun/pywin32/blob/master/Pythonwin/pywin/Demos/openGLDemo.py)

Things like the clipboard are due to be re-written anyway (#812), and sound is out of process already.
The keyboard support can't be too difficult either.

totaam:
Found an important note WRT pointer position values in [`WM_MOUSEMOVE`](https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-mousemove):
_Do not use the [LOWORD](https://learn.microsoft.com/en-us/windows/win32/winmsg/loword) or [HIWORD](https://learn.microsoft.com/en-us/windows/win32/winmsg/hiword) macros to extract the x- and y- coordinates of the cursor position because these macros return incorrect results on systems with multiple monitors. Systems with multiple monitors can have negative x- and y- coordinates, and LOWORD and HIWORD treat the coordinates as unsigned quantities._

And [`AdjustWindowRectEx`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-adjustwindowrectex): _his API is not DPI aware, and should not be used if the calling thread is per-monitor DPI aware. For the DPI-aware version of this API, see [AdjustWindowsRectExForDPI](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-adjustwindowrectexfordpi)._

totaam:
Issue explaining why gtk rendering is messy so we can avoid making the same mistakes: https://gitlab.gnome.org/GNOME/gtk/-/issues/105

totaam:
See also #1172, which we should support.
See also #1682 and [High DPI Desktop Application Development on Windows](https://learn.microsoft.com/en-us/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows)

totaam:
# Where you actually stand today

Two mechanisms currently set awareness, and they disagree in strength:

[...]

PR #4793, opened 2026-02-25
add middle-click config option to suppress mouse button 2

andrewachen:
## Summary
Add a `middle-click` boolean config option (default: `yes`) so users can disable button 2 forwarding with `middle-click = no` in their client config.

## Problem
ThinkPad TrackPoint (and similar pointing stick devices) use the middle button for scroll mode — you hold the middle button and move the stick to scroll. On native Win32 windows, the Synaptics/Lenovo driver intercepts `WM_MBUTTONDOWN` and converts it to `WM_MOUSEWHEEL` events. However, GTK3 on Windows creates windows that receive `WM_POINTER` messages instead of legacy `WM_MBUTTONDOWN`, and the TrackPoint driver does not intercept these. The result is that every scroll attempt also generates a spurious middle-click, pasting clipboard content in terminal windows.

This is a GTK3-on-Windows issue, not specific to Xpra. A minimal GTK3 test window with zero Xpra code reproduces the problem.

## What was tried
The following were all tested and ruled out as causes:
- `XPRA_WIN32_WINDOW_HOOKS=0` — disabling all Win32 window hooks
- `XPRA_WHEEL=0` — disabling Xpra's wheel handling
- `XPRA_SIMULATE_MOUSE_DOWN=0` — disabling synthetic mouse-down events
- `--opengl=no` — ruling out OpenGL native child windows
- `GDK_WIN32_TABLET_INPUT_API=none` — disabling GDK's WinPointer tablet API
- `XPRA_DPI_AWARE=0 XPRA_DPI_AWARENESS=0` — disabling DPI awareness calls
- `XPRA_FORWARD_WINDOWS_KEY=0` — disabling the low-level keyboard hook
- `XPRA_WIN32_WINDOW_EVENTS=0 XPRA_CONSOLE_EVENT_LISTENER=0 XPRA_TRAY=0` — disabling all background Win32 infrastructure
- `xpra start-desktop` (non-seamless mode) — same behavior, confirming process-level issue
- Bare GTK3 window (`TEST_LEVEL=0`) with zero Xpra code — still receives button 2 events, confirming this is fundamental to GTK3 on Windows

Since the root cause is in GTK3's Win32 backend and cannot be fixed from Xpra, this option provides a client-side workaround.


totaam:
> Since the root cause is in GTK3's Win32 backend and cannot be fixed from Xpra

FYI: it can, by removing GTK: #921 - WIP can be tested in v6.5 with `--backend=win32`.
But this will take longer.

---

Please add command line toggles for this option, preferably near:
https://github.com/Xpra-org/xpra/blob/7b3ce57dc546ab9248a8335943b4e894c3a62655/xpra/scripts/parsing.py#L1356
And a manual page update for the same.

PR #4796, opened 2026-03-04
fix audio forwarding: three PulseAudio routing bugs

andrewachen:
Audio forwarding silently fails in Xpra sessions when applications (e.g. browsers)
write audio to the wrong PulseAudio sink. Three related bugs:

**Bug 1 — `pactl_impl.py`: Sink sections parsed as Sources**

`do_get_pa_device_options()` iterated all `pactl list` output without tracking which section type it was in. Sink entries were collected as if they were Sources, so null sinks like `Xpra-Microphone` were returned as valid capture devices. This causes GStreamer `pulsesrc` to fail with "No such entity" when a client has no real audio hardware (containers, VMs, CI). On typical desktops the bug is masked by `USE_DEFAULT_DEVICE`.

Fix: track `in_source_section` and only record devices encountered inside `Source #N` sections.

**Bug 2 — `pulseaudio.py`: `PULSE_SINK` set in child process environment (primary)**

`get_child_env()` called `get_pulse_env()`, which sets `PULSE_SINK=Xpra-Microphone` and `PULSE_SOURCE=Xpra-Speaker` for all child processes. `PULSE_SINK` is a standard PulseAudio client variable that overrides the default sink — so every application
launched inside the session (including browsers) was directed to write audio to `Xpra-Microphone` instead of `Xpra-Speaker`. The audio was never captured and forwarded to the client.

The internal audio pipeline selects devices via `XPRA_PULSE_SINK_DEVICE_NAME` and `XPRA_PULSE_SOURCE_DEVICE_NAME`, not `PULSE_SINK`/`PULSE_SOURCE`, so removing these from the child environment does not affect audio capture or mic forwarding.

Fix: remove `env.update(self.get_pulse_env())` from `get_child_env()`.

**Bug 3 — `pulseaudio.py`: wrong default sink at socket open time (defense-in-depth)**

`Xpra-Microphone` was loaded before `Xpra-Speaker` in `get_default_pulseaudio_command()`. PulseAudio makes the last-loaded sink the default, so `Xpra-Microphone` was the default in the window between `module-native-protocol-unix` loading (socket opens, clients can connect) and `xpra.pa` running `set-default-sink`. Any client that reconnected quickly during a restart would attach to `Xpra-Microphone`.

Fix: swap load order so `Xpra-Speaker` is loaded last.

**Tests**

Added `tests/unittests/unit/audio/pactl_test.py` and `tests/unittests/unit/server/subsystem/pulseaudio_test.py` covering all three fixes.

This was tested on Ubuntu 24.04 on the 6.4 branch and ported to master.

### DIFF
--- a/tests/unittests/unit/audio/pactl_test.py
+++ b/tests/unittests/unit/audio/pactl_test.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+# This file is part of Xpra.
+# Copyright (C) 2026 Antoine Martin <antoine@xpra.org>
+# Xpra is released under the terms of the GNU GPL v2, or, at your option, any
+# later version. See the file COPYING for details.
+
+import unittest
+
+from xpra.audio.pulseaudio.pactl_impl import do_get_pa_device_options
+
+# Minimal pactl list output with one Sink section and one Source section
+SINK_AND_SOURCE = b"""Sink #0
+\tName: Xpra-Microphone
+\tdevice.class = "abstract"
+\tdevice.description = "Dummy Output"
+
+Source #0
+\tName: Xpra-Speaker.monitor
+\tMonitor of Sink: Xpra-Speaker
+\tdevice.description = "Monitor of Xpra-Speaker"
+
+"""
+
+# Output with both a monitor source and a regular (non-monitor) source
+MIXED_SOURCES = b"""Source #0
+\tName: alsa.monitor
+\tMonitor of Sink: alsa_output
+\tdevice.description = "Monitor of ALSA"
+
+Source #1
+\tName: alsa_input.mic
+\tdevice.class = "sound"
+\tdevice.description = "Built-in Mic"
+
+"""
+
+
+class TestDoGetPaDeviceOptions(unittest.TestCase):
+
+    def test_sink_sections_excluded(self):
+        # Sink sections must not appear as audio sources
+        devices = do_get_pa_device_options(SINK_AND_SOURCE, monitors=None, input_or_output=None)
+        assert "Xpra-Microphone" not in devices
+        assert "Xpra-Speaker.monitor" in devices
+
+    def test_sink_only_output_returns_empty(self):
+        # Output with only Sink sections and no Sources returns empty dict
+        sink_only = b"""Sink #0
+\tName: Xpra-Microphone
+\tdevice.class = "abstract"
+\tdevice.description = "Dummy Output"
+
+"""
+        devices = do_get_pa_device_options(sink_only, monitors=None, input_or_output=None)
+        assert devices == {}
+
+    def test_monitor_filter_true(self):
+        # monitors=True returns only monitor sources
+        devices = do_get_pa_device_options(MIXED_SOURCES, monitors=True, input_or_output=None)
+        assert "alsa.monitor" in devices
+        assert "alsa_input.mic" not in devices
+
+    def test_monitor_filter_false(self):
+        # monitors=False excludes monitor sources
+        devices = do_get_pa_device_options(MIXED_SOURCES, monitors=False, input_or_output=None)
+        assert "alsa.monitor" not in devices
+        assert "alsa_input.mic" in devices
+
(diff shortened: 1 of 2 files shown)


totaam:
> Sink sections parsed as Sources

This one looks legit, I'm not sure why it wasn't causing problems until now.

> PULSE_SINK is a standard PulseAudio client variable that overrides the default sink — so every application

We do want to set the default source and sink, and this has been working fine for years.
What has changed?
If the only problem was caused by the parsing of sections, then this can be left alone.

> wrong default sink at socket open time (defense-in-depth)

I don't see the change that would fix this, or even how it could be fixed.

---

Again, please specify the AI and prompts.

andrewachen:
> > Sink sections parsed as Sources
> 
> This one looks legit, I'm not sure why it wasn't causing problems until now.
> 
> > PULSE_SINK is a standard PulseAudio client variable that overrides the default sink — so every application
> 
> We do want to set the default source and sink, and this has been working fine for years. What has changed? If the only problem was caused by the parsing of sections, then this can be left alone.

My understanding is that this is only triggered if there are no other audio devices on the host system, as is the case for the VM that I am running Xpra on.

> > wrong default sink at socket open time (defense-in-depth)
> 
> I don't see the change that would fix this, or even how it could be fixed.

Here's Claude's explanation of this:

> The key is in get_default_pulseaudio_command() — it builds a list of arguments that PulseAudio executes sequentially as it starts up. The relevant sequence is:
>                                                                                                                                                                                                                      
>   1. --load=module-null-sink sink_name=Xpra-Microphone   ← PA default = Xpra-Microphone
>   2. --load=module-null-sink sink_name=Xpra-Speaker      ← PA default = Xpra-Speaker                                                                                                                                 
>   3. --load=module-remap-source ...                                                                                                                                                                                  
>   4. --load=module-native-protocol-unix socket=...       ← socket opens HERE                                                                                                                                         
>      ...                                                                                                                                                                                                             
>   (2 seconds later, via GLib.timeout_add at line 354):
>   5. configure_pulse() runs pulseaudio_configure_commands ← set-default-sink Xpra-Speaker
> 
>   PulseAudio's rule: when no default has been explicitly set, the most recently loaded null sink becomes the implicit default. So after step 2, the default is Xpra-Speaker. After step 4, the socket is open and
>   clients can connect. Any app that connects and plays audio before step 5 completes gets whatever the current PA default is.
> 
>   Before the fix, the order was reversed — Xpra-Speaker was loaded first, Xpra-Microphone second. So:
> 
>   1. Load Xpra-Speaker      ← default = Xpra-Speaker
>   2. Load Xpra-Microphone   ← default = Xpra-Microphone  ← WRONG
>   3. Load remap-source
>   4. Socket opens           ← default = Xpra-Microphone  ← apps connect here
>   5. set-default-sink       ← default = Xpra-Speaker (2s later)
> 
>   Any app connecting between steps 4 and 5 would attach to Xpra-Microphone — its audio would go into the void.
> 
>   After the fix, Xpra-Microphone loads first, Xpra-Speaker loads second, so Xpra-Speaker is already the PA default when the socket opens. The set-default-sink at step 5 is now redundant for the normal path — it
>   just confirms what's already true. Hence "defense-in-depth": the correct state is established before the race window even opens, so the 2-second timer is no longer load-bearing.


> Again, please specify the AI and prompts.

This was Anthropic Claude. It is listed as a Co-author in all commits for transparency. This was a long-running conversation about audio forwarding not working (I gave it this error message from Microsoft Edge `ALSA lib pcm_dmix.c:1000:(snd_pcm_dmix_open) unable to open slave [76134:76134:0303/161035.109685:ERROR:media/audio/alsa/alsa_util.cc:204] PcmOpen: default,No such file or directory ALSA lib pcm_dmix...`).

totaam:
Thanks, makes sense now.

I'll probably cherry-pick just the re-ordering for older branches.

totaam:
Looks like this completely broke audio forwarding - and it wasn't broken for me to begin with!
See #4984

PR #4797, opened 2026-03-04
fix Window Encoders section in Statistics tab

andrewachen:
The "Window Encoders" section in the Statistics tab of the session info dialog
has a bug that make it invisible and non-functional:

`populate_statistics` called `get_children()` and removed *all* children of
`encoder_info_box` including the "Window Encoders" title widget added in
`__init__`, then never re-added it. The section became invisible after the first
refresh cycle.

**Fix:** Only remove and replace encoder labels (children at index 1+), leaving
the title widget at index 0 untouched. Labels are only replaced when fresh data
is available.

### Additional improvements

- **Layout:** Changed `encoder_info_box` from `HBox` to `VBox` so multiple
  windows stack vertically rather than overflowing horizontally.
- **Spacing:** Added `margin_top=20` to match the padding used by the stats
  grid above.
- **Non-video encodings:** When no video encoder is active, falls back to
  `last_used` (e.g. `webp`, `png`) so the section always shows the current
  encoding rather than going blank.
- **Window titles:** Each entry now shows the window title (truncated to 40
  characters) alongside the window ID and encoder, making it easier to identify
  which window is using which encoder.
- **Tooltip:** The encoder detail tooltip now excludes the internal `"title"`
  key in addition to the encoder type name already shown in the label.

### DIFF
--- a/xpra/gtk/dialogs/session_info.py
+++ b/xpra/gtk/dialogs/session_info.py
@@ -1201,16 +1201,19 @@ class SessionInfo(Gtk.Window):
 
     def get_window_encoder_stats(self) -> dict:
         window_encoder_stats = {}
-        # info["client"] is always {int: source_info} (one entry per connected client)
+        # info["client"] is {int: source_info} — find our client by uuid
         server_last_info = self.last_info()
-        window_dict = {}
         client_info = server_last_info.get("client", {})
+        window_dict = {}
         for source in (v for v in client_info.values() if isinstance(v, dict)):
+            if bytestostr(source.get("uuid", "")) != self.client.uuid:
+                continue
             w = source.get("window")
             if isinstance(w, dict):
-                window_dict.update(w)
+                window_dict = w
+                break
         if window_dict:
-            id_to_window = getattr(self.client, "_id_to_window", {})
+            id_to_window = self.client._id_to_window
             for k, v in window_dict.items():
                 with log.trap_error("Error: cannot lookup window dict"):
                     wid = int(k)
@@ -1225,7 +1228,7 @@ class SessionInfo(Gtk.Window):
                             continue
                         stats = {"": last_used}
                     win = id_to_window.get(wid)
-                    title = (win.get_title() if win and hasattr(win, "get_title") else "") or ""
+                    title = (win.get_title() if win else "") or ""
                     stats["title"] = (title[:40] + "…") if len(title) > 40 else title
                     window_encoder_stats[wid] = stats
         return window_encoder_stats


xpra/server/base.py
```diff
@@ -526,6 +526,14 @@ def up(prefix, d) -> None:
...
             cinfo[i] = sinfo
         up("client", cinfo)
+        # expose window encoder info at the top level so the session info dialog can display it:
+        window_info: dict = {}
+        for sinfo in cinfo.values():
+            w = sinfo.get("window")
+            if isinstance(w, dict):
+                window_info.update(w)
+        if window_info:
+            info["window"] = window_info
```
totaam: Is this not already exposes somewhere else?

andrewachen:
> Which AI did you use to generate this? And using what prompt?

Anthropic Claude. It's listed as a coauthor on the commit for transparency.

It was a somewhat long-ish conversation around trying to debug whether or not NVENC was working, so it wasn't one specific prompt.

xpra/gtk/dialogs/session_info.py
```diff
@@ -1172,33 +1175,68 @@ def populate_statistics(self) -> bool:
...
-        window_dict = server_last_info.get("window")
-        if window_dict and isinstance(window_dict, dict):
+        window_dict = {}
+        client_info = server_last_info.get("client", {})
+        if "window" in client_info:
+            # single client: client_info IS the source info
+            sources = [client_info]
+        else:
+            # multiple clients: client_info is indexed by integer
+            sources = [v for v in client_info.values() if isinstance(v, dict)]
```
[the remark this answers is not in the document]
andrewachen: Fixed -- take a look.

xpra/gtk/dialogs/session_info.py
```diff
@@ -1172,33 +1175,60 @@ def populate_statistics(self) -> bool:
...
-        window_dict = server_last_info.get("window")
-        if window_dict and isinstance(window_dict, dict):
+        window_dict = {}
+        client_info = server_last_info.get("client", {})
+        for source in (v for v in client_info.values() if isinstance(v, dict)):
+            w = source.get("window")
+            if isinstance(w, dict):
+                window_dict.update(w)
+        if window_dict:
+            id_to_window = getattr(self.client, "_id_to_window", {})
```
totaam: This still doesn't look right to me.
The `client_info` is a dictionary for all the clients it exposes:
https://github.com/Xpra-org/xpra/blob/7a2f57970267b954c2566a00793958ec4978267b/xpra/server/base.py#L523-L528
And by default, that's all of them:
https://github.com/Xpra-org/xpra/blob/7a2f57970267b954c2566a00793958ec4978267b/xpra/server/base.py#L487-L491
In the case of multiple clients connected, this code would end up updating with the last client processed by this loop instead of the one we are interested in.
See:
```
xpra info | grep client | grep -i uuid
```
The client uuid is defined here:
https://github.com/Xpra-org/xpra/blob/7a2f57970267b954c2566a00793958ec4978267b/xpra/client/base/client.py#L117
And exposed to the server here:
https://github.com/Xpra-org/xpra/blob/7a2f57970267b954c2566a00793958ec4978267b/xpra/client/base/client.py#L305-L306

totaam:
Thanks!

PR #4799, opened 2026-03-05
Send full encoding caps to clients after background codec init

andrewachen:
🤖 This PR was generated by Anthropic Claude (claude-sonnet-4-6).

**How this was found:** The Server Picture Encodings section of the client Session Info dialog
was showing only default encodings (`rgb32`, `rgb24`, `jpeg`, `png`) instead of the full set
including h264/nvenc. Claude diagnosed and applied a local module override in a prior session to
restore functionality. In a later session, Claude revisited the patch as a starting point for
upstreaming, diagnosed the root cause more deeply (tracing why `threaded_init_complete()` was
never called — dead in both v6.4.x and master, but for different reasons), determined the local
patch's approach would not work on master, and designed and implemented a correct fix using
`GLib.idle_add` + a `threaded_encoding_done` flag. Tests were written first (TDD), then
implementation.

### Problem

After `threaded_encoding_setup()` loads video codecs, clients never receive an updated encoding
capability packet. The Statistics dialog shows only the default encodings (`rgb32`, `rgb24`,
`jpeg`, `png`) and the manual encoding selection menu is limited to those — even though the server
is actually encoding with h264/nvenc.

The actual encoding *does* switch correctly: `reinit_encoders()` is called on connected window
sources when init completes, so the server starts using hardware encoders. The missing piece is
the informational `threaded_init_complete()` packet back to the client, which updates
`server_core_encodings` for display and manual selection.

`threaded_init_complete()` has never been called from anywhere in the codebase. In master,
`threaded_encoding_setup()` called `init_encodings()` directly when done with no client
notification of any kind. The `init-thread-ended` signal that `reinit_encodings` was connected
to was also never emitted after the per-subsystem threading refactor, making that path doubly
dead.

### Fix

- After background codec loading completes, dispatch `reinit_encodings()` on the GLib main thread
  via `GLib.idle_add()` instead of calling `init_encodings()` directly. This ensures connected
  clients get `reinit_encoders()` followed by `threaded_init_complete()`, which sends the full
  encoding capability packet.
- Add `add_new_client()` to handle clients that connect after init has already completed:
  `reinit_encodings()` will have run with an empty source list and those clients would never
  receive `threaded_init_complete()`. Track `threaded_encoding_done` to distinguish this case —
  there is no other observable signal that `reinit_encodings()` has already fired.

### Tests

Three new tests in `encoding_test.py`:

- `test_reinit_encodings_sends_caps`: verifies `reinit_encodings()` sends the capability packet
  to connected clients
- `test_add_new_client_sends_caps_when_init_done`: verifies `add_new_client()` sends caps
  immediately when background init is already complete
- `test_add_new_client_defers_caps_when_init_pending`: verifies `add_new_client()` does not
  send caps while background init is still running

### DIFF
--- a/tests/unittests/unit/server/subsystem/encoding_test.py
+++ b/tests/unittests/unit/server/subsystem/encoding_test.py
@@ -6,8 +6,23 @@
 
 import unittest
 
-from xpra.util.objects import AdHocStruct
+from xpra.util.objects import AdHocStruct, typedict
 from unit.server.subsystem.servermixintest_util import ServerMixinTest
+
+
+def _make_opts():
+    opts = AdHocStruct()
+    opts.encoding = ""
+    opts.encodings = ["rgb", "png"]
+    opts.quality = 0
+    opts.min_quality = 20
+    opts.speed = 0
+    opts.min_speed = 20
+    opts.video = True
+    opts.video_scaling = "auto"
+    opts.video_encoders = []
+    opts.csc_modules = []
+    return opts
 
 
 class EncodingMixinTest(ServerMixinTest):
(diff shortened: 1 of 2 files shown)


totaam:
Thanks!

The refactoring was clearly incomplete - I had hit this bug but not investigated it yet.
I may want to rename `threaded_init_complete` later, it is remnant of the way things were done in earlier versions.

PR #4808, opened 2026-03-12
Xpra HTML5 Keyboard Input Fix

diyism:
# Xpra HTML5 Keyboard Input Fix

## Problem
Xpra HTML5 client keyboard input was not working. Keys pressed in the browser were not being sent to the X11 applications running in the xpra session.

## Environment
- **Container**: lobster-xpra (Docker)
- **Xpra Version**: 6.5
- **Display**: :100
- **HTML5 Client**: Accessed via browser at http://...:9010
- **Test Application**: mousepad text editor

### Initial Investigation
1. Added extensive debug logging to track keyboard event flow
2. Identified that keyboard events were being received by the server
3. Found that `do_get_keycode()` was returning -1 (invalid keycode)

### Key Findings
1. **HTML5 clients vs Native clients**:
   - Native clients send `x11_keycodes` dictionary with pre-mapped keycodes
   - HTML5 clients only send `keyname` (e.g., "a", "Control_L") and `keyval` (JavaScript keyCode)
   - HTML5 clients set `x11_keycodes=False`

2. **The Problem**:
   - When `x11_keycodes=False`, the code fell back to `find_matching_keycode()`
   - `find_matching_keycode()` requires server-side keyboard mapping data
   - This data wasn't initialized for HTML5 clients, causing it to return -1
   - Without a valid keycode, `_handle_key()` was never called

3. **JavaScript keyCode vs X11 keysym**:
   - HTML5 client sends `keyval=74` for "j" (JavaScript keyCode)
   - X11 expects keysym=106 for "j" (ASCII value)
   - For modifier keys: JavaScript sends keyval=17 for Control, but X11 keysym is 65507

### Implementation
Modified `/usr/local/lib/python3.11/dist-packages/xpra-6.5-py3.11-linux-x86_64.egg/xpra/x11/server/keyboard_config.py`

Added a new code path in `do_get_keycode()` for HTML5 clients:

```python
# HTML5/Web clients: convert keyname to X11 keysym, then look up keycode
if not self.x11_keycodes and keyname:
    try:
        with xsync:
            # First try to parse keyname to get the correct X11 keysym
            x11_keysym = X11Keyboard.parse_keysym(keyname)
            if x11_keysym == 0:
                # If keyname parsing fails, try using keyval directly
                x11_keysym = keyval
            keysym_mappings = X11Keyboard.get_keysym_mappings()
            keycodes = keysym_mappings.get(x11_keysym, ())
        if keycodes:
            x11_keycode = keycodes[0]  # Use first keycode
            log.info("[KEY_DEBUG] Keysym lookup: keyname=%s, x11_keysym=%s -> keycode=%s",
                     keyname, x11_keysym, x11_keycode)
            return x11_keycode, group
        log.info("[KEY_DEBUG] No keycode found for keyname=%s, x11_keysym=%s",
                 keyname, x11_keysym)
    except Exception as e:
        log.info("[KEY_DEBUG] Keysym lookup failed: %s", e)
```

### How It Works
1. Use `X11Keyboard.parse_keysym(keyname)` to convert keyname to proper X11 keysym
   - "a" → 97
   - "Control_L" → 65507
   - "Return" → 65293
2. Use `X11Keyboard.get_keysym_mappings()` to get server's keyboard layout
3. Look up the X11 keycode for the keysym
4. Return the keycode so `_handle_key()` can process it

## Browser Keyboard Shortcut Interception (Resolved)

**Initial Issue**: Certain keyboard shortcuts (Ctrl+X, Ctrl+C, Ctrl+V, etc.) did not work in the HTML5 client. Web browsers intercept common keyboard shortcuts for their own operations, so only the key *release* event was being sent to the server, not the key *press* event.

**Solution Implemented**:
Modified `/usr/local/lib/python3.11/dist-packages/xpra-6.5-py3.11-linux-x86_64.egg/xpra/server/subsystem/keyboard.py` to synthesize missing key press events and properly inject modifiers.

Added logic to detect when a key is released but was never pressed. In this case, we first ensure that the required modifiers (like Control) are physically active on the X11 server using `ss.make_keymask_match()`, and then we synthesize the missing key press event before the release event is processed.

```python
# Fix for browser-intercepted shortcuts: synthesize missing key press events
if not pressed and keycode > 0:
    if keycode not in keyboard_config.keys_pressed:
        # First ensure modifiers are active on the server
        ss.make_keymask_match(modifiers, keycode, ignored_modifier_keynames=[keyname])
        # Then recursively call with pressed=True
        self.do_process_keyboard_event(proto, wid, keyname, True, kattrs)
```

This properly triggers the shortcuts (like Cut/Copy/Paste) in the remote X11 applications, even when initiated by a "fake" or intercepted shortcut event from the browser.

### DIFF
--- a/xpra/server/subsystem/keyboard.py
+++ b/xpra/server/subsystem/keyboard.py
@@ -268,6 +268,14 @@ class KeyboardServer(StubServerMixin):
             self.set_keyboard_layout_group(group)
         self._focus(ss, wid, None)
         ss.make_keymask_match(modifiers, keycode, ignored_modifier_keynames=[keyname])
+        # Fix for browser-intercepted shortcuts: synthesize missing key press events
+        if not pressed and keycode > 0:
+            if keycode not in self.keys_pressed:
+                # First ensure modifiers are active on the server
+                ss.make_keymask_match(modifiers, keycode, ignored_modifier_keynames=[keyname])
+                # Then recursively call with pressed=True
+                self.do_process_keyboard_event(proto, wid, keyname, True, kattrs)
+
         # negative keycodes are used for key events without a real keypress/unpress
         # for example, used by win32 to send Caps_Lock/Num_Lock changes
         if keycode >= 0:
--- a/xpra/x11/server/keyboard_config.py
+++ b/xpra/x11/server/keyboard_config.py
@@ -545,6 +545,28 @@ class KeyboardConfig(KeyboardConfigBase):
             return -1, group
         if self.raw:
             return client_keycode, group
+
+        # HTML5/Web clients: convert keyname to X11 keysym, then look up keycode
+        if not self.x11_keycodes and keyname:
+            try:
+                with xsync:
+                    # First try to parse keyname to get the correct X11 keysym
+                    x11_keysym = X11Keyboard.parse_keysym(keyname)
+                    if x11_keysym == 0:
+                        # If keyname parsing fails, try using keyval directly
+                        x11_keysym = keyval
+                    keysym_mappings = X11Keyboard.get_keysym_mappings()
+                    keycodes = keysym_mappings.get(x11_keysym, ())
+                if keycodes:
+                    x11_keycode = keycodes[0]  # Use first keycode
+                    log.info("[KEY_DEBUG] Keysym lookup: keyname=%s, x11_keysym=%s -> keycode=%s",
+                             keyname, x11_keysym, x11_keycode)
+                    return x11_keycode, group
+                log.info("[KEY_DEBUG] No keycode found for keyname=%s, x11_keysym=%s",
+                         keyname, x11_keysym)
+            except Exception as e:
+                log.info("[KEY_DEBUG] Keysym lookup failed: %s", e)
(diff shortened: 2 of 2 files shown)


diyism:
And the xpra-html5 also needs fix:
https://github.com/Xpra-org/xpra-html5/compare/master...diyism:xpra-html5:master

totaam:
> Xpra Version: 6.5

Yes, the keyboard mapping is in the process of being updated, for 2 major changes:
* asynchronous initialization
* generic keyboard events (moving away from X11 naming and behaviour conventions)

I will take a look at this fix ASAP.

> And the xpra-html5 also needs fix:

What does this do?

totaam:
This AI slop was just plain wrong.
The correct workaround is in aa0e76abb64240441ec51545d450f7ff84f59d97 - as for the xpra-html5 "fix", I have no idea why this would be needed at all.

Note: this is just a workaround for the html5 client, which should be updated to use asynchronous initialization: https://github.com/Xpra-org/xpra-html5/issues/420

PR #4815, opened 2026-03-17
Fix GL window cursor on Win32 via HWND subclassing

andrewachen:
## Problem

On Win32, GL windows have a separate child HWND for the DrawingArea (created by the GL backend, outside GDK's knowledge). GDK's `WM_SETCURSOR` handler stores the cursor on the `GdkWindow` but only handles `WM_SETCURSOR` for GDK-managed HWNDs. The GL child HWND is not one of them, so Windows falls back to the class cursor (arrow) regardless of what the server sends.

The symptom: xterm windows launched via xpra on Win32 always show an arrow cursor instead of the correct I-beam, because GDK never applies the server-provided cursor to the GL drawing surface.

## Fix

Subclass the GL child HWND via `comctl32.SetWindowSubclass` to intercept `WM_SETCURSOR` and call `SetCursor` with the correct `HCURSOR` directly.

The HCURSOR is extracted from the `GdkWin32Cursor` C struct by reading the pointer from `repr()` and offsetting to the `hcursor` field (offset 48, documented in `gdkprivate-win32.h`). This is ABI-fragile but GTK3 is in maintenance mode (3.24.x final) so the layout is frozen; the code degrades gracefully to the broken-cursor state if extraction fails rather than crashing.

**Lifecycle:**
- Subclass installed on first `enter-notify-event` on the DrawingArea widget
- HCURSOR holder updated by `set_windows_cursor` without re-subclassing
- On cursor reset, `gdkwin.set_cursor()` is also called to clear GDK's stored cursor (otherwise the parent HWND's `WM_SETCURSOR` handler restores the stale cursor when the subclass passes through to `DefWindowProc`)
- Subclass removed on `destroy()`, `new_backing()`, and self-removes on `WM_NCDESTROY`

**Changes:**
- `xpra/client/gtk3/opengl/client_window.py`: all Win32-specific subclassing logic, guarded by `if WIN32:`
- `xpra/client/gtk3/client_base.py`: cache `_last_cursor_data` for GL windows that miss a cursor packet; skip `gdkwin.set_cursor()` for subclassed windows to prevent GDK from overriding the subclass

## Notes

- No effect on non-Win32 platforms; the `if WIN32:` guard prevents any ctypes imports or signal connections on Linux/macOS
- `SetWindowSubclass` is the Microsoft-recommended approach for safe, chainable HWND subclassing (comctl32 v6+)
- `WindowFromPoint` is used to find the GL child HWND (not `gdk_win32_window_get_handle`) because the target HWND is created by the GL backend outside GDK's object graph; using the GdkWindow handle would subclass the wrong HWND and leave the bug in place

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-sonnet-4-6)

This PR was created by Claude in response to: _"xterm windows not showing the proper I-beam mouse cursor depending on what was in the foreground when launched"_ — followed by iterative code review (two Opus-model review passes), bug fixes, and squash.

### DIFF
--- a/xpra/client/gtk3/client_base.py
+++ b/xpra/client/gtk3/client_base.py
@@ -184,6 +184,11 @@ class GTKXpraClient(GObjectClientAdapter, UIXpraClient):
         self.gl_max_viewport_dims = 0, 0
         self.gl_texture_size_limit = 0
         self._cursors = weakref.WeakKeyDictionary()
+        # Global cursor cache: GL windows created after a cursor update (or whose
+        # backing is recreated) use this on enter-notify to recover the current
+        # cursor. Per-window cursor_data is set by set_cursor_data(); this cache
+        # covers windows that were absent when the last cursor packet arrived.
+        self._last_cursor_data: tuple = ()
         # frame request hidden window:
         self.frame_request_window = None
         # group leader bits:
@@ -930,6 +935,7 @@ class GTKXpraClient(GObjectClientAdapter, UIXpraClient):
             if cursor is None:
                 # use default:
                 cursor = get_default_cursor()
+        self._last_cursor_data = cursor_data
         for w in windows:
             w.set_cursor_data(cursor_data)
             # the cursor should only apply to the window contents (aka "drawingarea"),
@@ -939,7 +945,9 @@ class GTKXpraClient(GObjectClientAdapter, UIXpraClient):
             # trays don't have a gdk window
             if gdkwin:
                 self._cursors[w] = cursor_data
-                gdkwin.set_cursor(cursor)
+                update_fn = getattr(w, "_update_cursor_subclass", None)
+                if update_fn is None or not update_fn(cursor):
+                    gdkwin.set_cursor(cursor)
 
     def make_cursor(self, cursor_data: Sequence) -> Gdk.Cursor | None:
         # if present, try cursor ny name:
--- a/xpra/client/gtk3/opengl/client_window.py
+++ b/xpra/client/gtk3/opengl/client_window.py
@@ -6,13 +6,67 @@
 
 from xpra.client.gtk3.window.window import ClientWindow
 from xpra.gtk.window import set_visual
+from xpra.os_util import gi_import, WIN32
 from xpra.util.objects import typedict
 from xpra.util.env import envbool
 from xpra.log import Logger
 
+Gdk = gi_import("Gdk")
+
 log = Logger("opengl", "window")
+cursorlog = Logger("opengl", "cursor")
 
 MONITOR_REINIT = envbool("XPRA_OPENGL_MONITOR_REINIT", False)
+
+if WIN32:
+    import ctypes
+    import ctypes.wintypes
+    import re
+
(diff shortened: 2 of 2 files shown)


totaam:
First, `xterm` windows normally do not use opengl as they should be tagged as `text` - are you using `--opengl=force`?

Even assuming that it is needed, all the ctypes stuff does not belong there. It should go in `xpra.platform.win32.common`.
`_get_hcursor_from_gdk_cursor` also does not belong there.

`GLClientWindowBase` is clean and knows nothing about platforms - this butchers it.
I am unconvinced that this is actually needed: this solution seems unnecessarily complicated and intrusive.
Adding ~250 lines of code to set a cursor is not acceptable.

andrewachen:
I am not forcing OpenGL -- the problem is that new windows default to OpenGL. The server will eventually send over that the content-type is text, but by that point, the window has already been created and there's no attempt to change the backing at that point. I could implement something that flips the backing from GLAreaBacking to CairoBacking after the content-type becomes available, and that would negate the need for this patch on xterms. (It would still break on things like browsers, which get OpenGL and would be unable to display the I-beam in those cases.)

That said, I'm happy to refactor this into the xpra.platform.win32.common package -- just let me know.

totaam:
> I am not forcing OpenGL -- the problem is that new windows default to OpenGL

This will be fixed in #4818

> That said, I'm happy to refactor this into the xpra.platform.win32.common package -- just let me know.

Let's try to find a simpler solution.
Two things could be merged in preparation:
* "Global cursor cache" looks like a valid fix - no matter what
* "Win32 API function signatures" could be added - making the diff smaller once a suitable fix is found

andrewachen:
> > I am not forcing OpenGL -- the problem is that new windows default to OpenGL
> 
> This will be fixed in #4818

After a bit more testing, it looks like I'll be switching to `opengl=force`; it is substantially faster on my client, so the OpenGL cursor fix will be great for me. I'll open a separate PR for an OpenGL Catmull-Rom implementation and benchmark. The difference is pretty stunning.

> > That said, I'm happy to refactor this into the xpra.platform.win32.common package -- just let me know.
> 
> Let's try to find a simpler solution. Two things could be merged in preparation:
> 
> * "Global cursor cache" looks like a valid fix - no matter what
> * "Win32 API function signatures" could be added - making the diff smaller once a suitable fix is found

Yes, I can split this up -- I'll open two new PRs for that.

totaam:
> it looks like I'll be switching to `opengl=force`

We can probably change the logic to use opengl more often - these defaults were added when opengl was not stable on all platforms.
FYI: there is still the issue of opengl not working properly on win32 with transparency, something I am hoping to fix with replacing Gtk: #921

andrewachen:
> > it looks like I'll be switching to `opengl=force`
> 
> We can probably change the logic to use opengl more often - these defaults were added when opengl was not stable on all platforms. 

That would be great.

> FYI: there is still the issue of opengl not working properly on win32 with transparency, something I am hoping to fix with replacing Gtk: #921

Yes, Microsoft Edge is now my only window left that is using Cairo.

PR #4821, opened 2026-03-23
Allow OpenGL on Win32 for opaque-region-covered alpha windows

andrewachen:
## Summary

On Win32, `has-alpha=True` blocks OpenGL rendering in `can_use_opengl()` because GTK3 sets `WS_EX_LAYERED` on alpha windows, making OpenGL content invisible to the desktop compositor. Apps like Chromium and Edge create 32-bit ARGB windows for internal compositing but declare themselves fully opaque via `_NET_WM_OPAQUE_REGION`. Since these windows are opaque at the desktop level, OpenGL can render them normally.

This PR:
- Adds `is_covered_by_opaque_region()` helper to `xpra.common` (shared by client and server)
- Modifies `can_use_opengl()` on Win32 to allow OpenGL when opaque-region covers the full window despite `has-alpha=True`
- Scales opaque-region from server coordinates to client-scaled coordinates (to handle desktop scaling correctly)
- Refactors `compress.py` to use the shared helper (same behavior, less duplication)

Tested with Microsoft Edge (32-bit depth, `has-alpha=True`, `opaque-region=((0, 0, 1396, 1022),)`, `decorations=1`) on a Win32 client with 1.6x desktop scaling. Window renders a full-screen video playback via OpenGL at 22fps with nvenc encoding in BGRX format.

## Platform impact

- **Win32**: Chromium/Edge/Electron windows gain OpenGL rendering where they were previously forced to Cairo
- **Linux/macOS**: Unaffected — the check is inside the `if WIN32:` block; these platforms already allow OpenGL for alpha windows
- **HTML5 client**: Unaffected — change is only in the GTK3 Python client

## Limitations

- Only helps **decorated** alpha windows. Undecorated windows are blocked by a separate Win32 check (GTK3 `WS_EX_LAYERED` issue, #2514)
- If opaque-region shrinks after window creation, the GL renderer stays and transparent areas render opaque. This is no worse than Cairo on Win32 (which also can't do per-pixel transparency). In practice, Chromium sets opaque-region once at creation and never changes it.

## Port notice

This is a port to `master` of a version tested on `v6.4.x` (`andrewachen/fix/server-opaque-alpha-metadata` branch).

## AI attribution

This feature was debugged and implemented by [Anthropic Claude](https://claude.ai) (Claude Opus 4.6), pair-programming with @andrewachen. The prompt was roughly to debug why Microsoft Edge continued to use Cairo and what would be needed to make it render with OpenGL.

### DIFF
--- a/xpra/server/window/compress.py
+++ b/xpra/server/window/compress.py
@@ -23,6 +23,7 @@ from xpra.util.thread import check_main_thread
 from xpra.net.common import Packet, BACKWARDS_COMPATIBLE
 from xpra.net.packet_type import WINDOW_DRAW
 from xpra.constants import MAX_WINDOW_SIZE, WINDOW_DECODE_SKIPPED, WINDOW_DECODE_ERROR, WINDOW_NOT_FOUND
+from xpra.common import is_covered_by_opaque_region
 from xpra.server.window.windowicon import WindowIconSource
 from xpra.server.window.perfstats import WindowPerformanceStatistics
 from xpra.server.window.batch_delay_calculator import calculate_batch_delay, get_target_speed, get_target_quality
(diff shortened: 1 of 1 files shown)


totaam:
Very nice and not intrusive!
I do wonder if we could have resolved this completely server side by setting `alpha=False` there.

andrewachen:
I (and Claude) had considered that, but that might have unintended consequences on Linux and MacOS, which do seem to handle alpha ok.

totaam:
FYI: found an issue on the server side: cab52cf0ff9260188c39c432cf31781326f95a07
But from what I have seen of how applications use opaque region, not a huge problem.

PR #4822, opened 2026-03-23
Apply current cursor to newly-created windows

andrewachen:
## Problem

When a new window is created after the most recent cursor packet from the server, it shows the default arrow cursor. The server sends cursor data when the cursor changes and expects all windows to apply it, but windows that don't exist yet at that time have no way to recover the current cursor.

Since the server only sends cursor packets on actual cursor changes, the arrow persists indefinitely if the server-side cursor hasn't changed — typically until the client reconnects. This is most noticeable when opening new terminal or editor windows while an I-beam cursor is active.

## Solution

- Cache the last cursor data in `_last_cursor_data` on each `set_windows_cursor()` call
- In `show_window()`, after `window.show_all()` realizes the widget tree, apply the cached cursor to the newly-shown window

New windows immediately get the correct cursor with no dependency on future server packets.

The `show_window()` hook uses `getattr(self, "_last_cursor_data", ())` to access the cache defensively — if `CursorClient` isn't mixed in, it's a no-op. This follows the existing cross-mixin pattern (e.g., `CursorClient` already accesses `self._id_to_window` from `WindowClient`).

## Drawbacks

- Adds a `set_windows_cursor()` call per new window, which builds a `GdkCursor` from the cached data. This is a one-time cost at window creation and should be negligible.
- On cursor reset (empty cursor data), the cache is set to `()` and `show_window()` skips the apply, so new windows correctly get the system default.

### DIFF
--- a/xpra/client/gtk3/client_base.py
+++ b/xpra/client/gtk3/client_base.py
@@ -184,6 +184,9 @@ class GTKXpraClient(GObjectClientAdapter, UIXpraClient):
         self.gl_max_viewport_dims = 0, 0
         self.gl_texture_size_limit = 0
         self._cursors = weakref.WeakKeyDictionary()
+        # Last cursor data applied; show_window() uses this to apply the
+        # current cursor to newly-created windows.
+        self._last_cursor_data: tuple = ()
         # frame request hidden window:
         self.frame_request_window = None
         # group leader bits:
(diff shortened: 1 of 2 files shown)


totaam:
LGTM - sorry I broke it, can you please re-base on top of bd4ff849667a91cee0489084032248d6cd0dc7d4, `show_window` has just been moved here:
https://github.com/Xpra-org/xpra/blob/2b605215b4a361b92bce5a2a6da671f0c735ca90/xpra/client/subsystem/window/manager.py#L303

BTW, I was just looking at this massive 1500 lines base class thinking that some components should also be split out:
https://github.com/Xpra-org/xpra/blob/master/xpra/client/gtk3/client_base.py
Especially things like cursor - as they're quite well self-contained and already split up in the gtk-agnostic superclass under "subsystems":
https://github.com/Xpra-org/xpra/tree/master/xpra/client/subsystem
So let's merge it now, and I may split it up afterwards.

PR #4824, opened 2026-03-23
Fix GL window cursor on Win32 via platform HWND subclass

andrewachen:
## Problem

On Win32, GL windows have a child HWND for the DrawingArea whose WM_SETCURSOR handler resets the cursor to the default arrow on every mouse move. `gdkwin.set_cursor()` sets the cursor on the GDK window, but the child HWND's WM_SETCURSOR immediately overrides it.

This means the cursor cache fix (#4822) alone is not sufficient for GL windows — it applies the cursor via `gdkwin.set_cursor()`, which is overridden by the child HWND on the next mouse move. Cairo windows don't have this problem because GDK handles `WM_SETCURSOR` correctly on their HWNDs.

### Why subclassing is necessary

Two simpler alternatives were tested and both fail (see [test results comment](https://github.com/Xpra-org/xpra/pull/4824#issuecomment-4113255328)):

- **SetClassLongPtrW with actual HCURSOR** (Option 1): Class cursor IS updated (logs confirm), but the child HWND's WndProc handles WM_SETCURSOR directly before DefWindowProc ever sees it. Changing GCLP_HCURSOR has no effect.
- **Null class cursor to bubble WM_SETCURSOR** (Option 2): GCLP_HCURSOR is per-class and shared across all GL windows. Nulling it causes cursor to disappear entirely.

Both alternatives assume DefWindowProc handles WM_SETCURSOR, which testing proves is wrong. The comctl32 `SetWindowSubclass` approach works because it inserts our handler *before* the original WndProc.

## Solution

Subclass the GL DrawingArea's child HWND via comctl32 `SetWindowSubclass` to intercept `WM_SETCURSOR` and call `SetCursor` with the correct `HCURSOR`.

All Win32 code lives in `xpra.platform.win32.gui` — `GLClientWindowBase` remains completely platform-agnostic, calling only `setup_gl_drawing_area`/`cleanup_gl_drawing_area` from `xpra.platform.gui`.

### Changes

- `platform/gui.py`: no-op stubs + `platform_import`
- `platform/win32/gui.py`: HCURSOR extraction, HWND subclass, realize/enter-notify handlers
- `client/gtk3/opengl/client_window.py`: two platform calls (`setup_gl_drawing_area`/`cleanup_gl_drawing_area`)
- `client/gtk3/client_base.py`: `_update_cursor_subclass` duck-typing in `set_windows_cursor`

Supersedes #4815.

Sponsored-by: Netflix

### DIFF
--- a/tests/unittests/unit/client/gl_cursor_probe_test.py
+++ b/tests/unittests/unit/client/gl_cursor_probe_test.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python3
+# This file is part of Xpra.
+# Copyright (C) 2026 Netflix, Inc.
+# Xpra is released under the terms of the GNU GPL v2, or, at your option, any
+# later version. See the file COPYING for details.
+# ABOUTME: Tests for the runtime HCURSOR offset probe used by the Win32 GL cursor fix.
+# ABOUTME: Verifies the probe discovers a valid offset and extracts working HCURSOR handles.
+
+import struct
+import unittest
+from ctypes import addressof, c_void_p, sizeof, create_string_buffer
+
+from xpra.os_util import WIN32
+
+
+def _make_fake_struct(fields):
+    """Create a fake C struct in memory with pointer-sized fields.
+    Returns (buffer, base_address)."""
+    ptr_size = sizeof(c_void_p)
+    buf = create_string_buffer(len(fields) * ptr_size)
+    fmt = "P" * len(fields)
+    struct.pack_into(fmt, buf, 0, *fields)
+    return buf
+
+
+class TestScanLogic(unittest.TestCase):
+    """Test the offset-scanning algorithm independent of Win32 APIs."""
+
+    def _scan_for_differing_valid_handles(self, ptr_a, ptr_b, scan_limit, is_valid_handle):
+        """Reimplementation of the core scan logic from _probe_hcursor_offset
+        for testing without Win32 dependencies."""
+        ptr_size = sizeof(c_void_p)
+        for offset in range(0, scan_limit, ptr_size):
+            val_a = c_void_p.from_address(ptr_a + offset).value or 0
+            val_b = c_void_p.from_address(ptr_b + offset).value or 0
+            if not val_a or not val_b or val_a == val_b:
+                continue
+            if is_valid_handle(val_a) and is_valid_handle(val_b):
+                return offset
+        return -1
+
+    def test_finds_differing_field(self):
+        """Two structs with same layout, different values at one field."""
+        ptr_size = sizeof(c_void_p)
+        # Struct: [same, same, DIFFERENT, same]
+        buf_a = _make_fake_struct([100, 200, 0xABCD, 400])
+        buf_b = _make_fake_struct([100, 200, 0xDCBA, 400])
+        valid = {0xABCD, 0xDCBA}
+        offset = self._scan_for_differing_valid_handles(
+            addressof(buf_a), addressof(buf_b),
+            len(buf_a), lambda v: v in valid,
+        )
+        self.assertEqual(offset, 2 * ptr_size)
+
+    def test_skips_same_values(self):
+        """Fields with identical values are skipped."""
+        buf_a = _make_fake_struct([100, 200, 300])
+        buf_b = _make_fake_struct([100, 200, 300])
+        offset = self._scan_for_differing_valid_handles(
+            addressof(buf_a), addressof(buf_b),
+            len(buf_a), lambda v: True,
+        )
+        self.assertEqual(offset, -1)
+
+    def test_skips_zero_values(self):
+        """Fields with zero values are skipped even if they differ."""
+        buf_a = _make_fake_struct([0, 200])
+        buf_b = _make_fake_struct([100, 200])
+        offset = self._scan_for_differing_valid_handles(
+            addressof(buf_a), addressof(buf_b),
+            len(buf_a), lambda v: True,
+        )
+        # First field: val_a=0, skipped. Second field: same. No match.
+        self.assertEqual(offset, -1)
+
+    def test_skips_invalid_handles(self):
+        """Fields that differ but fail validation are skipped."""
+        buf_a = _make_fake_struct([0xAAAA, 0xBBBB])
+        buf_b = _make_fake_struct([0xCCCC, 0xDDDD])
+        valid = {0xBBBB, 0xDDDD}
+        ptr_size = sizeof(c_void_p)
+        offset = self._scan_for_differing_valid_handles(
+            addressof(buf_a), addressof(buf_b),
+            len(buf_a), lambda v: v in valid,
+        )
+        self.assertEqual(offset, 1 * ptr_size)
+
+    def test_first_valid_match_wins(self):
+        """When multiple fields differ and validate, the first offset wins."""
+        buf_a = _make_fake_struct([0x1111, 0x3333])
+        buf_b = _make_fake_struct([0x2222, 0x4444])
+        ptr_size = sizeof(c_void_p)
+        offset = self._scan_for_differing_valid_handles(
+            addressof(buf_a), addressof(buf_b),
+            len(buf_a), lambda v: True,
+        )
+        self.assertEqual(offset, 0)
+
+
+@unittest.skipUnless(WIN32, "Win32-only: requires GDK Win32 backend")
+class TestHCursorProbeWin32(unittest.TestCase):
+    """End-to-end tests that run on Win32 with a real GDK display."""
+
+    def test_probe_finds_offset(self):
+        """The probe should discover a non-negative offset."""
+        from xpra.platform.win32.gui import _probe_hcursor_offset
+        offset = _probe_hcursor_offset()
+        self.assertGreaterEqual(offset, 0, "probe failed to find HCURSOR offset")
+        # Offset should be reasonable (within typical GObject struct sizes)
+        self.assertLess(offset, 256)
+
+    def test_extract_hcursor_from_arrow(self):
+        """Extracting HCURSOR from an arrow cursor should return non-zero."""
+        from xpra.platform.win32.gui import _get_hcursor_from_gdk_cursor
+        from xpra.os_util import gi_import
+        Gdk = gi_import("Gdk")
+        display = Gdk.Display.get_default()
+        if not display:
+            self.skipTest("no display")
+        cursor = Gdk.Cursor.new_for_display(display, Gdk.CursorType.ARROW)
+        hcursor = _get_hcursor_from_gdk_cursor(cursor)
+        self.assertNotEqual(hcursor, 0, "failed to extract HCURSOR from arrow cursor")
+
+    def test_extract_hcursor_different_types(self):
+        """Different cursor types should yield different HCURSOR values."""
+        from xpra.platform.win32.gui import _get_hcursor_from_gdk_cursor
+        from xpra.os_util import gi_import
+        Gdk = gi_import("Gdk")
+        display = Gdk.Display.get_default()
+        if not display:
+            self.skipTest("no display")
+        arrow = Gdk.Cursor.new_for_display(display, Gdk.CursorType.ARROW)
+        cross = Gdk.Cursor.new_for_display(display, Gdk.CursorType.CROSSHAIR)
+        h_arrow = _get_hcursor_from_gdk_cursor(arrow)
+        h_cross = _get_hcursor_from_gdk_cursor(cross)
+        self.assertNotEqual(h_arrow, 0)
+        self.assertNotEqual(h_cross, 0)
+        self.assertNotEqual(h_arrow, h_cross,
+                            "arrow and crosshair should have different HCURSOR handles")
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/xpra/client/gtk3/client_base.py
+++ b/xpra/client/gtk3/client_base.py
@@ -943,7 +943,9 @@ class GTKXpraClient(GObjectClientAdapter, UIXpraClient):
             # trays don't have a gdk window
             if gdkwin:
                 self._cursors[w] = cursor_data
-                gdkwin.set_cursor(cursor)
+                update_fn = getattr(w, "_update_cursor_subclass", None)
+                if update_fn is None or not update_fn(cursor):
+                    gdkwin.set_cursor(cursor)
 
     def make_cursor(self, cursor_data: Sequence) -> Gdk.Cursor | None:
         # if present, try cursor ny name:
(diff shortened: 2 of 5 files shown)


totaam:
I asked Claude for simpler alternatives.

I am unable to test any of these solutions at the moment, as I need to revive a broken MS Windows system to test natively - my virtual machines don't have a working opengl driver.

---

A few alternatives that avoid subclassing entirely, roughly ordered by invasiveness:

1. SetClassLongPtr(GCLP_HCURSOR) — one call, no hooking
When DefWindowProc handles WM_SETCURSOR for a child window, it calls SetCursor with the class cursor (GCLP_HCURSOR). So:
pythonfrom xpra.platform.win32.common import GetClassLongPtr, SetClassLongPtr
GCLP_HCURSOR = -12
SetClassLongPtr(child_hwnd, GCLP_HCURSOR, hcursor)
No subclassing. No message interception. Just one call whenever your logical cursor changes, patching the class cursor to whatever you want the GL area to show. DefWindowProc's WM_SETCURSOR handler will then call SetCursor with your cursor.
Caveat: GCLP_HCURSOR is global to the window class, not the instance. If the GL child HWND class is shared with other top-level windows that expect the arrow, you'd clobber them. In practice, the GL surface child HWND is registered with its own unique WNDCLASS, so this is almost certainly safe.

2. Set GCLP_HCURSOR to NULL and let WM_SETCURSOR bubble
If you set the class cursor to NULL:
cSetClassLongPtr(child_hwnd, GCLP_HCURSOR, 0);
DefWindowProc will not call SetCursor and will instead send WM_SETCURSOR up to the parent — the GDK parent window — which does have the right cursor registered. You're effectively opting the child HWND out of cursor management entirely. No cursor-value tracking needed in your code at all.
This only works if the GDK parent's WM_SETCURSOR correctly sets the cursor from its own GCLP_HCURSOR / SetCursor state, which it should since that's exactly what gdkwin.set_cursor() modifies.

3. GDK Win32 message filter on the DrawingArea
If the child HWND is actually wrapped by a GDK window (sometimes it is at the GTK layer), GDK's existing filter mechanism works:
pythondrawing_area.get_window().add_filter(wm_setcursor_filter)
Where the filter just calls SetCursor(current_hcursor) and returns GDK_FILTER_REMOVE for WM_SETCURSOR. This is GDK-idiomatic, no ctypes subclassing, and stays within the layer you're already using. The catch is that add_filter only fires for HWNDs GDK itself owns; if the GL child HWND is below GDK's abstraction, the filter won't see it.

Recommendation
Option 2 (null class cursor, let it bubble) is the most elegant if it works — zero cursor tracking, zero subclassing, pure delegation. A quick test would confirm whether GDK's parent HWND WM_SETCURSOR path picks it up correctly.
If bubbling is unreliable (e.g. GDK eats the message before re-dispatching), option 1 (SetClassLongPtr with the actual cursor value) is the next simplest — it's literally one extra call alongside wherever you already call SetCursor or gdkwin.set_cursor().

andrewachen:
## Testing results for the simpler alternatives

I tested both Option 2 (null class cursor) and Option 1 (SetClassLongPtr with actual HCURSOR) on Win32 ARM64 (Snapdragon X Elite, Adreno X1-85, GTK3 3.24.x). Both fail for the same fundamental reason.

### Option 2: Set GCLP_HCURSOR to NULL — FAILED

Setting `SetClassLongPtrW(child_hwnd, GCLP_HCURSOR, 0)` causes two problems:

1. **GCLP_HCURSOR is per-class, not per-instance.** All GL child HWNDs share one window class. The first window's null propagated immediately to all subsequent windows (logs showed `old GCLP_HCURSOR=0` for every window after the first).

2. **WM_SETCURSOR did not bubble to the GDK parent.** Instead of the parent applying the correct cursor, the cursor disappeared entirely on typing (consistent with `SetCursor(NULL)` being called). The parent's WM_SETCURSOR handler was not picking up the bubbled message.

### Option 1: SetClassLongPtr with actual HCURSOR — FAILED

Setting `SetClassLongPtrW(child_hwnd, GCLP_HCURSOR, hcursor)` with the correct HCURSOR value — verified by logs showing the class cursor IS being updated on every cursor change:

```
SetClassLongPtrW GCLP_HCURSOR=0x2e924e5 (prev=0x323ca) on child_hwnd=0x10255a
SetClassLongPtrW GCLP_HCURSOR=0x1b6819bd (prev=0x2e924e5) on child_hwnd=0x10255a
```

The HCURSOR extraction works, SetClassLongPtrW succeeds (prev values change correctly), but **the cursor still shows as arrow**. This proves that the GL child HWND's WndProc does **not** use DefWindowProc's default WM_SETCURSOR handler. The WndProc handles WM_SETCURSOR directly, calling SetCursor with the arrow cursor before DefWindowProc ever sees the message. Changing GCLP_HCURSOR has no effect because DefWindowProc's WM_SETCURSOR code path is never reached.

### Why subclassing is the minimum viable fix

Both alternatives assume DefWindowProc handles WM_SETCURSOR for the GL child HWND. Testing proves this assumption is wrong — the child's WndProc intercepts WM_SETCURSOR before DefWindowProc.

The comctl32 `SetWindowSubclass` approach works because it inserts our handler into the subclass chain *before* the original WndProc. Our handler calls `SetCursor(our_hcursor)` and returns 1 (handled), preventing the original WndProc from ever seeing WM_SETCURSOR.

The updated PR moves all Win32 subclassing code to `xpra.platform.win32.gui` — `GLClientWindowBase` remains completely platform-agnostic, calling only `setup_gl_drawing_area`/`cleanup_gl_drawing_area` from `xpra.platform.gui`.

totaam:
I still don't have the hardware to test, and I am confused by this diff. Adding hooks in:
https://github.com/andrewachen/xpra/blob/9dfd6d198c72c350955310cd60241b557cd4191a/xpra/client/gtk3/opengl/client_window.py#L98
Means that we have a `drawing_area` gtk widget, can't we set the cursor on that?
I can't remember which of `drawing_area.py` or `glarea_backing.py` is used on win32 by default. (`--opengl=native` vs `--opengl=gtk`) but this looks like the right place to forward `set_cursor` calls - and leave it as a no-op on all other platforms.

---

And even if that doesn't work:
> All GL child HWNDs share one window class

Wouldn't it be easier to just change this instead?

---

I can't merge things like `_get_hcursor_from_gdk_cursor` that use regular expressions to find object addresses with hard-coded offsets.
And the whole change is just too big for what it does, the maintenance cost would be way too high.

andrewachen:
Thanks for the feedback. Here's where each concern stands:

**"Can't we set the cursor on the drawing_area?"**

`set_windows_cursor` already calls `gdkwin.set_cursor(cursor)` on the drawing_area's GDK window. This works for the parent HWND but not the GL child HWND — the child has its own WndProc that handles WM_SETCURSOR directly and overrides whatever GDK set. We now also ensure `gdkwin.set_cursor()` is always called (not skipped), so both the parent and child HWNDs have cursors set — the parent via GDK, the child via the subclass. This fixed a cursor-disappearing-during-drag issue.

**"Wouldn't it be easier to change the shared window class?"**

We tested both class-level approaches on Win32 ARM64 (see [test results](https://github.com/Xpra-org/xpra/pull/4824#issuecomment-4113255328)):
- `SetClassLongPtrW(GCLP_HCURSOR, hcursor)`: class cursor IS updated (logs confirm), but the child HWND's WndProc handles WM_SETCURSOR before DefWindowProc reads it. No effect.
- `SetClassLongPtrW(GCLP_HCURSOR, 0)`: GCLP_HCURSOR is per-class and shared across all GL windows. Cursor disappeared entirely.

**"Can't merge regex + hardcoded offsets"**

Replaced with a runtime self-calibrating probe. No regex, no hardcoded offsets. `_probe_hcursor_offset()`:
1. Creates two GDK cursors (ARROW, CROSSHAIR) and gets their C struct pointers via `PyCapsule_GetPointer` (same pattern as `get_window_handle`)
2. Queries `GType.instance_size` for the exact struct size
3. Scans pointer-aligned offsets for values that differ between the two structs and are valid cursor handles (verified by `GetIconInfo`)
4. If the probe fails for any reason, the cursor fix is safely disabled

Unit tests cover the scan logic (platform-independent, runs in Linux CI) and end-to-end probe + extraction (Win32-only, confirmed passing on ARM64 CI).

**Code size**

The diff is ~260 lines in `win32/gui.py`, of which ~65 are the probe, ~80 the subclass, and the rest is setup/cleanup/logging. `GLClientWindowBase` remains completely platform-agnostic — it just calls `setup_gl_drawing_area`/`cleanup_gl_drawing_area` from `xpra.platform.gui`.

**Windows CI**

I've also added a basic Windows ARM64 unit test workflow ([7b82de3](https://github.com/andrewachen/xpra/commit/7b82de3261bc188830631272f90da8d738802577)) that runs a curated set of client tests on `windows-11-arm` with MSYS2 CLANGARM64. It only covers specific test files at this point, but would be good to expand to broader coverage.

---

*This response was drafted with Claude Code and reviewed by @andrewachen before posting.*

totaam:
> `set_windows_cursor` already calls `gdkwin.set_cursor(cursor)` on the drawing_area's GDK window.

Specifically, here:
https://github.com/Xpra-org/xpra/blob/70b51cd90e2fa3654c56e66cc8bc6720b51a31ac/xpra/client/gtk3/client_base.py#L937-L946

More notes regarding `native` vs `glarea`, I believe that the selection is done here:
https://github.com/Xpra-org/xpra/blob/70b51cd90e2fa3654c56e66cc8bc6720b51a31ac/xpra/opengl/window.py#L25-L39
And that MS Windows will use `native`, which loads:
https://github.com/Xpra-org/xpra/blob/70b51cd90e2fa3654c56e66cc8bc6720b51a31ac/xpra/opengl/window.py#L71
The gtk native window implementation, which uses the `GLDrawingArea` backend:
https://github.com/Xpra-org/xpra/blob/70b51cd90e2fa3654c56e66cc8bc6720b51a31ac/xpra/client/gtk3/opengl/native_window.py#L29-L30
This is just a plain `Gtk.DrawingArea` widget:
https://github.com/Xpra-org/xpra/blob/70b51cd90e2fa3654c56e66cc8bc6720b51a31ac/xpra/client/gtk3/opengl/drawing_area.py#L38-L48
Have you tried setting the cursor on this widget directly for testing purposes?

---

> This works for the parent HWND but not the GL child HWND

What is the `GL child HWND`?
Isn't this the HWND of the `Gtk.DrawingArea`?
The paint context gets the HWND:
https://github.com/Xpra-org/xpra/blob/70b51cd90e2fa3654c56e66cc8bc6720b51a31ac/xpra/platform/win32/gl_context.py#L157-L165
By calling 
https://github.com/Xpra-org/xpra/blob/70b51cd90e2fa3654c56e66cc8bc6720b51a31ac/xpra/platform/win32/gui.py#L189-L201

So I don't understand:
* why we can't just use Gtk to set the cursor on the `DrawingArea`
* this statement:
> the child has its own WndProc that handles WM_SETCURSOR directly and overrides whatever GDK set.

Which WndProc? Set by whom? How is Gtk bypassed?

andrewachen:
> Have you tried setting the cursor on this widget directly for testing purposes?

Your code tracing is correct — [line 941](https://github.com/Xpra-org/xpra/blob/70b51cd90e2fa3654c56e66cc8bc6720b51a31ac/xpra/client/gtk3/client_base.py#L941) gets the `Gtk.DrawingArea` widget via `getattr(w, "drawing_area", w)`, and [line 946](https://github.com/Xpra-org/xpra/blob/70b51cd90e2fa3654c56e66cc8bc6720b51a31ac/xpra/client/gtk3/client_base.py#L946) calls `gdkwin.set_cursor(cursor)` on its GDK window. The WndProc diagnostic below shows why this call has no effect on GL windows.

> What is the GL child HWND? Isn't this the HWND of the Gtk.DrawingArea?

Yes. The `Gtk.DrawingArea` starts as a client-side GDK window. When `get_window_handle()` is called by [gl_context.py line 158](https://github.com/Xpra-org/xpra/blob/70b51cd90e2fa3654c56e66cc8bc6720b51a31ac/xpra/platform/win32/gl_context.py#L158), `gdk_win32_window_get_handle()` promotes it to native via `gdk_window_ensure_native()` (`gdk/win32/gdkwindow-win32.c:6319` in gtk-3-24). So the parent window and the DrawingArea have separate native HWNDs.

> Which WndProc? Set by whom? How is Gtk bypassed?

The OpenGL driver. I added diagnostic logging that reads `GetWindowLongPtrW(hwnd, GWL_WNDPROC)` at two points: at DrawingArea realize (before WGL context creation) and at first cursor update (after WGL). Results from Adreno X1-85 on Snapdragon X Elite:

```
WndProc at realize: child=0x7ffd4e980aa4, parent=0x7ffd4e980aa4, match=True
WndProc at cursor update: 0x7ffcfc0600f0 (was 0x7ffd4e980aa4 at realize, changed=True)
```

At realize, the DrawingArea HWND has the same WndProc as the parent — GDK's `_gdk_win32_window_procedure`. After WGL context creation (`SetPixelFormat`/`wglCreateContext`), the driver replaces it with `0x7ffcfc0600f0`. Parent HWNDs are unaffected.

GDK's WM_SETCURSOR handler (`gdk/win32/gdkevents-win32.c:3287` in gtk-3-24) checks `impl->cursor` and calls `SetCursor` — the logic is correct. But it never executes because the driver's WndProc handles WM_SETCURSOR without forwarding to GDK.

This also explains why `SetClassLongPtrW(GCLP_HCURSOR)` had no effect in earlier tests — DefWindowProc reads the class cursor, but the driver's WndProc handles WM_SETCURSOR without reaching DefWindowProc.

Diagnostic commit: [363bc39](https://github.com/andrewachen/xpra/commit/363bc39184) on `fix/win32-gl-cursor-v3`.

---

*Drafted with Claude Code and reviewed by @andrewachen before posting.*

totaam:
This still does not feel right to me yet. Please bear with me, I need to find some hardware I can install Microsoft Windows on to test it.
Is the problem specific to this driver or are all opengl drivers going to override `WndProc`?
As in, will I be able to reproduce the issue with an Intel GPU?

andrewachen:
> This still does not feel right to me yet. Please bear with me, I need to find some hardware I can install Microsoft Windows on to test it. 

No worries; I have my own builds going and this meets my needs, so I'm not in a rush or blocked by this.

> Is the problem specific to this driver or are all opengl drivers going to override `WndProc`? As in, will I be able to reproduce the issue with an Intel GPU?

I'm not sure; I have only tested this on my Adreno-based machine.

andrewachen:
I now have an x64 machine with an Intel GPU and can confirm that without this fix, I-beams are not showing up on xterms, so this is not an Adreno-specific thing. The fix does work on x64. The primary reason I haven't been testing on x64 is it's not my daily driver and I didn't have builds for it.

I now have x64 and arm64 Windows builds via GH Actions. It leverages the cache functionality, so iteration cycles only take about 5 minutes now. https://github.com/andrewachen/xpra/blob/v6.4.3-achen/.github/workflows/windows.yml

totaam:
I found a simpler version of this fix and applied it in f8e6aa68c28d6d21f5436271256eb6bfb46cb3fc.

As much as I still dislike the extra hook, at least now the work it does is fairly small.

PR #4826, opened 2026-03-23
Add sigmoid upscaling and anti-ringing to OpenGL upscale shader

andrewachen:
## Summary

Enhances the Catmull-Rom upscale shader with two techniques from the video player world (mpv/libplacebo) that reduce ringing artifacts on text and UI edges at non-integer scale factors:

- **Sigmoid upscaling**: transforms each texel through a sigmoid curve before filtering, compressing contrast at sharp edges so the filter produces less overshoot. Uses mpv/libplacebo defaults (center=0.75, slope=6.5). Reference: [libplacebo colorspace.c](https://github.com/haasn/libplacebo/blob/master/src/shaders/colorspace.c)

- **Anti-ringing clamp**: constrains the filtered output to the min/max range of the 4 nearest source texels (strength=0.8). Reference: [Artoriuz's evaluation](https://artoriuz.github.io/blog/mpv_upscaling.html)

Both enhancements are controlled by uniforms (`use_sigmoid`, `ar_strength`) so they can be toggled at draw time without recompilation. The shader always uses 16-tap individual texel fetches (`GL_NEAREST`) since the sigmoid transform is nonlinear and cannot use the bilinear trick.

## Benchmark results (1280x720 @ 1.6x)

| Method | PSNR | Overshoot | SSIM | ms/frame (RTX A2000) | ms/frame (Adreno X1-85) |
|--------|------|-----------|------|---------------------|------------------------|
| Plain Catmull-Rom | 22.4 dB | 0.017 | 0.919 | 1.3ms | 3.8ms |
| **CR + sigmoid + AR** | **23.5 dB** | **0.003** | **0.939** | **1.1ms** | **3.9ms** |

+1.1 dB PSNR, 5x less ringing, +0.020 SSIM — well within the 16.67ms frame budget on both desktop and mobile GPUs.

Also includes an AMD FidelityFX RCAS (contrast-adaptive sharpening) benchmark variant ported from [agyild's mpv GLSL port](https://gist.github.com/agyild/82219c545228d70c5604f865ce0b0ce5) (MIT license), but the +0.005 SSIM improvement from the second pass is imperceptible and not worth the complexity for a compositor.

## Changes

- `xpra/opengl/shaders.py`: Replace 9-tap bilinear-trick CR shader with 16-tap uniform-controlled CR + sigmoid + AR shader
- `xpra/opengl/backing.py`: Set `use_sigmoid` and `ar_strength` uniforms, use `GL_NEAREST`
- `tests/scripts/benchmark_scaling.py`: Add sigmoid, AR, RCAS benchmarks with PSNR, overshoot, and SSIM metrics; parallelized CPU references; AMD RCAS port

### DIFF
--- a/xpra/opengl/backing.py
+++ b/xpra/opengl/backing.py
@@ -934,16 +934,18 @@ class GLWindowBackingBase(WindowBackingBase):
         program = self.programs["upscale"]
         glUseProgram(program)
 
-        # Bind FBO texture with GL_LINEAR for the bilinear trick on the center fetch
+        # GL_NEAREST: the 16-tap shader fetches individual texels (no bilinear trick)
         glActiveTexture(GL_TEXTURE0)
         glBindTexture(target, self.textures[TEX_FBO])
-        glTexParameteri(target, GL_TEXTURE_MAG_FILTER, GL_LINEAR)
-        glTexParameteri(target, GL_TEXTURE_MIN_FILTER, GL_LINEAR)
+        glTexParameteri(target, GL_TEXTURE_MAG_FILTER, GL_NEAREST)
+        glTexParameteri(target, GL_TEXTURE_MIN_FILTER, GL_NEAREST)
         glUniform1i(glGetUniformLocation(program, "fbo"), 0)
 
         glUniform2f(glGetUniformLocation(program, "viewport_pos"),
                     left * xscale, top * yscale)
         glUniform2f(glGetUniformLocation(program, "scaling"), xscale, yscale)
+        glUniform1i(glGetUniformLocation(program, "use_sigmoid"), 1)
+        glUniform1f(glGetUniformLocation(program, "ar_strength"), 0.8)
 
         pos_buffer = self.set_vao(0)
         glDrawArrays(GL_TRIANGLE_STRIP, 0, 4)
--- a/xpra/opengl/shaders.py
+++ b/xpra/opengl/shaders.py
@@ -28,9 +28,23 @@
 # G = Y - (a * e / b) * Cr - (c * d / b) * Cb
 # B = Y + d * Cb
 
+import math
+
 from xpra.codecs.constants import get_subsampling_divs
 
 GLSL_VERSION = "330 core"
+
+# Sigmoid upscaling constants (mpv/libplacebo defaults).
+# Compresses contrast before filtering to reduce ringing on sharp edges.
+# Reference: libplacebo src/shaders/colorspace.c, mpv video/out/gpu/video.c
+_SIG_CENTER = 0.75
+_SIG_SLOPE = 6.5
+_SIG_OFFSET = 1.0 / (1.0 + math.exp(_SIG_SLOPE * _SIG_CENTER))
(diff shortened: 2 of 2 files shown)


totaam:
Good stuff. Seeing how long it took me to get these shaders working, it's amazing to get this kind of improvement so quickly!

andrewachen:
Benchmark results on x64 Windows (Intel Meteor Lake):

**System:** Core Ultra 7 165U, Intel Graphics (Xe-LPG, 4 Xe-cores / 128 EU, up to 2.0 GHz)

**1280x720 @ 1.6x:**

| Method | RTX A2000 | Adreno X1-85 | Intel Xe-LPG |
|--------|-----------|-------------|--------------|
| Plain Catmull-Rom | 1.3ms | 3.8ms | 2.51ms |
| **CR + sigmoid + AR** | **1.1ms** | **3.9ms** | **4.21ms** |

All variants comfortably within the 16.67ms frame budget. The Intel iGPU shows more overhead from sigmoid+AR (+1.7ms) than Adreno (+0.1ms), likely due to the 16-tap `GL_NEAREST` fetch pattern being more costly on the smaller Xe-LPG shader array.

totaam:
Is this going to scale?
> 1280x720 @ 1.6x:
> (..)
> Intel Xe-LPG (..) 4.21ms

4k is 9 times the number of pixels!
Assuming linear scaling (which is probably wrong), that would be 40ms on Intel!?

andrewachen:
It is linear. On my A2000 (because I didn't want to wait for the Intel...), 3840x2160 is 29ms/frame, which is 34fps (640x480 was 1.1ms on this card). Given it's only doing the work for damage, I don't think it'll actually be all that bad most of the time. There's always `XPRA_OPENGL_SCALING_FILTER=bilinear` which is fast at all resolutions.

totaam:
In the future, it might be worth considering an automatic switch back to `bilinear` for resolutions greater than 4k and / or with slower GPUs. (which can be measured in realtime rather than guesstimated)
Going completely over the top, we could also take into account the user's quality vs speed preference.

ie here's the metadata for a `glxgears` window:
```
compress:   1.5ms for  300x298  pixels at    0,0    for wid=12582914 using      h264 with ratio   0.7%  (  350KB to     3KB), sequence    56, \
    client_options={'frame': 33, 'pts': 1569, 'csc': 'YUV420P', 'full-range': True, 'type': 'B', 'delayed': 4, 'csc-type': 'libyuv', 'encoder': 'x264', 'flush': 0}, \
    options=typedict({'quality': 69, 'speed': 16, 'rgb_formats': ('BGRX', 'RGBX', 'BGR', 'RGB', 'r210', 'BGR565'), 'lz4': True, 'alpha': False, 'content-types': ('video',), 'av-delay': 0, 'window-size': (300, 298), 'damage-time': 314540.00169921, 'process-damage-time': 314540.038880415, 'depth': 24, 'b-frames': True})
```
A `YUV420P` subsampled frame from a `video` content-type window can use bilinear if needed at larger sizes to maximize the framerate, a `browser` content-type window should not.

Issue #4515, opened 2025-02-24
Resized windows shrink and migrate


matthewpersico:
**Describe the bug**
After resizing a window, it will redraw itself, shrinking in size and migrating to the lower right hand corner of the display.

**To Reproduce**
Resize a window. It only happens intermittently.

**System Information (please complete the following information):**
 - Server OS: Red Hat Enterprise Linux release 8.10 (Ootpa)
 - Client OS: Windows 10.0.19041.5486
 - Xpra Server Version: v6.1-r0
 - Xpra Client Version: xpra v6.3-r37166 (g17113d0213) beta

**Additional context**
Sorry, cannot reproduce. Can you give me some information as to getting server logs when this happens so I can send them along?

totaam:
This kind of bug is almost impossible to debug from the log samples.
What application can I use to reproduce it?
`-d geometry,metadata` may help, if you include it at both ends.

totaam:
This is very likely to be caused by the 250% scaling.
Microsoft Windows then usually lies to the applications about the true resolution of the display, despite the manifest file specifically stating that it should not: 0c0f7e26aa095f1ab6282fc0fd1c4339838a42f2
You can see what is detected from the output of `Xpra_cmd.exe`, or from the server log.

totaam:
See #4831 - which was meant to be making things better... and may have made things worse!

PR #4831, opened 2026-03-25
snap window resize to increment grid

andrewachen:
## Summary
- `motion_moveresize` only clamped to min/max bounds but ignored `width_inc`/`height_inc` from `WM_NORMAL_HINTS`, leaving terminal windows (xterm, etc) with a white border where the window size didn't land on a character-cell boundary
- Extract `snap_to_increment()` to floor dimensions to the resize increment grid (`base + increment * N`) and apply it during alt-drag resize, with position adjustment for left/top edge drags to keep the opposite edge fixed
- Flush pending `moveresize_data` on button release so the final snapped size is always applied
- Snap size in `send_configure_event` before sending to the server, preventing server-side `move_resize_window` corrections that caused the window to bounce back to a previous size

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Anthropic Claude). Prompt: "when I resize xterm windows with alt-drag, it creates a white border; as if it's not snapping correctly to xterm's actual size"

Sponsored-by: Netflix

### DIFF
--- a/tests/unittests/unit/client/moveresize_test.py
+++ b/tests/unittests/unit/client/moveresize_test.py
@@ -113,6 +113,17 @@ class TestSnapToIncrement(unittest.TestCase):
             assert snapped_cw == snapped_cw2, \
                 f"unstable at client_w={client_w}: {snapped_cw} → {snapped_cw2}"
 
+    def test_snap_shrinks_maximized_size(self):
+        # Snapping a maximized/fullscreen window would shrink it below screen size.
+        # This validates WHY we skip snapping for maximized/fullscreen windows.
+        hints = {"width_inc": 10, "height_inc": 20, "base_width": 4, "base_height": 2}
+        screen_w, screen_h = 1920, 1080
+        snapped_w, snapped_h = snap_to_increment(screen_w, screen_h, hints)
+        # screen size is almost never on the grid — snap shrinks it
+        assert snapped_w <= screen_w
+        assert snapped_h <= screen_h
+        assert (snapped_w, snapped_h) != (screen_w, screen_h)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/xpra/client/gtk3/window/base.py
+++ b/xpra/client/gtk3/window/base.py
@@ -1116,6 +1116,8 @@ class GTKClientWindowBase(ClientWindowBase, Gtk.Window):
 
     def do_increment_snap(self) -> None:
         self.increment_snap_timer = 0
+        if self._fullscreen or self._maximized:
+            return
         target = self.increment_snap_target
         if not target:
             return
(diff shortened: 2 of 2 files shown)


totaam:
Makes sense.
Did you try this with desktop scaling?

Looks like this does not - but maybe it should?
https://github.com/Xpra-org/xpra/blob/99d5c7919438cf07d94bf072c2801c673c1764e1/xpra/client/gtk3/window/base.py#L1004-L1011

Also, since the actual changes happen via the move-resize timer:
https://github.com/Xpra-org/xpra/blob/99d5c7919438cf07d94bf072c2801c673c1764e1/xpra/client/gtk3/window/base.py#L1052-L1057
Wouldn't it make sense to only apply the `snap_to_increment` in `do_moveresize`?

andrewachen:
> Looks like this does not - but maybe it should?

The snap is applied after the min/max clamping. Since `snap_to_increment` only rounds sizes down (subtracts the modulus), it can't push past the max. In theory it could go below min, but `base_width`/`base_height` act as a floor so that shouldn't happen in practice. The two constraints are independent and both applied.

> Wouldn't it make sense to only apply the `snap_to_increment` in `do_moveresize`?

Good point — I considered this placement. The complication is the position adjustment for left/top edge drags (SIZE_TOPLEFT, SIZE_LEFT, etc.): when snapping reduces the width, the position needs to shift right by the same amount to keep the opposite edge fixed. That adjustment needs `direction` from `moveresize_event`, which is set to `None` before the final `do_moveresize()` flush on button release, so it wouldn't be available there.

There's also a secondary benefit to the current placement: `moveresize_data` always holds grid-aligned values, so the window is always in a consistent state regardless of when the timer fires.

Happy to restructure if you feel strongly, but I think keeping the snap and position adjustment together immediately after the delta calculation is the cleanest approach.

totaam:
> The two constraints are independent and both applied.

I'm talking about the desktop-scaling option, which scales all coordinates by a fixed amount - this is notoriously difficult to handle with size-constraints, as the rounding errors can cause resize loops.
It looks to me like the coordinates are not scaled here - and I believe they should be.

> I think keeping the snap and position adjustment together immediately after the delta calculation is the cleanest approach

That's fine.

andrewachen:
Updated the PR with a significantly reworked approach based on the scaling discussion. The original version snapped in client coordinates using `geometry_hints`, which breaks at non-integer desktop scales (e.g. 1.6x) because `set_size_constraints` drops `base_width` when it doesn't scale to an integer, causing snapping to a wrong-offset grid.

Changes in this update (3 commits):

**1. `fix move_resize losing resize_counter`** — standalone bugfix: `move_resize()` called `self.resize(w, h)` without passing `resize_counter`, clobbering it to 0.

**2. `snap window resize to increment grid`** — the core fix, reworked for correct behavior at any desktop scale:
- `snap_to_server_grid()` snaps in server coordinates using the unscaled hints from `size_constraints` (always has correct base+increment regardless of scale)
- `motion_moveresize` and `send_configure_event` use `snap_to_server_grid` instead of client-coord snap
- `resize(force=True)` uses GDK directly to bypass WM geometry hint enforcement for server-authoritative sizes
- `apply_geometry_hints` clears `RESIZE_INC` from the WM hint mask when `BASE_SIZE` is not also set (prevents WM from snapping to wrong-offset grid)

**3. `add nearest rounding and debounced client snap timer`**:
- `send_configure_event` uses nearest rounding (not floor) to keep the `cx→snap→sx` round-trip stable at any scale — a ±1 rounding error from double conversion can never cross the midpoint of a grid cell (`inc/2 ≥ 1`), so nearest always converges in one step
- 50ms debounced timer in `do_configure_event` corrects the client window size after WM handle resize settles; checks staleness before applying (ignores itself if the window was resized since scheduling)
- Timer only schedules when `size_constraints` has increment hints, avoiding pointless resizes on non-terminal windows

**Why this is not fragile:**

Every component degrades gracefully to baseline xpra behavior. If `size_constraints` is empty, `snap_to_server_grid` is a no-op. If the server changes its snap formula and disagrees with ours, server corrections resume via `move_resize(force=True)` — the same path that existed before. If the timer fires stale or not at all, the only effect is a small cosmetic gap on WM handle resize. Nothing is worse than pre-existing behavior.

The core snap uses the same formula as the server's `calc_constrained_size` (`w -= (w - base) % inc`), so client and server always agree on grid positions. Since the client sends grid-aligned values, the server has nothing to correct — this eliminates server round-trips entirely for the alt-drag path, which is a strict improvement under high-latency connections (no delayed bounce-back from server corrections arriving 500ms+ later).

The timer's staleness check (`current_size != recorded_size`) naturally handles mid-drag cancellation without needing explicit drag-state tracking. During active WM drag, configure events keep arriving, so any timer that fires sees a stale target and does nothing. Only after the drag stops does the timer see a matching size and apply the correction.

Tested on v6.4.x at 1x and 1.6x desktop scale; this is a port to master. Analyzed for correctness at sub-1x scales (nearest rounding prevents the oscillation that floor rounding causes at the `cx`/`sx` round-trip boundary).

🤖 This comment was drafted by [Claude Code](https://claude.ai/claude-code) (Anthropic Claude).

totaam:
I haven't reviewed this as thoroughly as I would like, but it looks like the correct approach and should solve the issue where we go back and forth between client and server, rounding the window dimensions down and entering a loop where the window just shrinks until it reaches its minimum size - this has plagued us for years.
Thanks!

totaam:
I think I have seen #4515 with a MS Windows client - which means that I may have to revert this or find a better fix.

PR #4839, opened 2026-03-28
Backup subcommand writes to truncated filename instead of config path

### LINKED ISSUE #4838: fix: backup subcommand writes to truncated filename instead of config path
## Description

The backup filename is computed as `conf[-5:] + ".bak"`, which uses only the last 5 characters of the config path. This causes backups to be written to an unrelated relative file (often in current working directory), potentially overwriting wrong files and silently failing to back up the real config.

**Severity**: `high`
**File**: `xpra/gtk/configure/main.py`

## Expected Behavior

The code should handle this case properly to avoid unexpected errors or degraded quality.

huyhoang171106:
## Summary

The backup filename is computed as `conf[-5:] + ".bak"`, which uses only the last 5 characters of the config path. This causes backups to be written to an unrelated relative file (often in current working directory), potentially overwriting wrong files and silently failing to back up the real config.

## Files changed

- `xpra/gtk/configure/main.py` (modified)

## Testing

Closes #4838

### DIFF
--- a/xpra/gtk/configure/main.py
+++ b/xpra/gtk/configure/main.py
@@ -30,7 +30,7 @@ def main(args=()) -> ExitValue:
             if not os.path.exists(conf):
                 print(f"# {conf!r} does not exist yet")
                 return ExitCode.FILE_NOT_FOUND
-            bak = conf[-5:]+".bak"
+            bak = conf+".bak"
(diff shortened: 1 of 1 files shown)


totaam:
528ffe14571d959a168be3a221b38b52775a8324 is better.

PR #4841, opened 2026-03-28
Make XShm segment lifecycle explicit and idempotent to prevent leaked server-side buffers

### LINKED ISSUE #4751: Using existing X11 display causing GPU memory leak
**Describe the bug**
When using `xpra seamless --use-display` option, GPU memory use of x11 process grows when resizing window. 

**To Reproduce**
1. `xpra start :100 --bind-tcp=0.0.0.0:10000 --video-encoders=nvenc,x264 --encoding=nvenc --use-display`
2. `.\Xpra_cmd.exe attach tcp://localhost:10000`
3. Start `vkcube`, resize the window for some time.
4. GPU memory of `xorg` process keeps increasing when checked with `nvidia-smi`, and is not freed when `vkcube` or `xpra` are stopped.

 - Server OS: Ubuntu24.04 (containerized)
 - Client OS: Windows11
 - server: xpra v6.5-r39704
 - client: Xpra-Light-x86_64_6.5-r39621

**Additional context**
I'm running xpra to develop gpu accelerated gui apps inside docker containers. The Xorg process is started specifically for the remote connection and not used for anything else. This was the only setup that resulted in acceptable performance. Would be happy to hear if there is a better way to do this.

huyhoang171106:
## Summary

The current leak pattern (growth during resize only while a client is attached) strongly points to shared-memory image churn where old XShm segments are not released promptly on buffer replacement. This file should be updated so segment teardown is explicit, repeat-safe, and independent of Python GC timing.

## Files changed

- `xpra/x11/bindings/shm.pyx` (modified)

## Testing

Closes #4751

### DIFF
--- a/xpra/x11/bindings/shm.pyx
+++ b/xpra/x11/bindings/shm.pyx
@@ -114,12 +114,31 @@ cdef class XShmWrapper:
     def __repr__(self):
         return "XShmWrapper(%#x - %ix%i)" % (self.window, self.width, self.height)
 
+    cdef void release(self):
+        if self.closed:
+            return
+        self.closed = True
+        self.cleanup()
+        self.shminfo.shmaddr = NULL
+        self.shminfo.shmid = -1
+        self.shminfo.readOnly = 0
+        self.image = NULL
+        self.display = NULL
+        self.visual = NULL
+        self.ref_count = 0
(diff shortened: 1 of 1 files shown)


totaam:
Please address the copilot review comments.

PR #4842, opened 2026-03-28
Missing PolicyKit authentication agent in `xpra desktop` sessions

### LINKED ISSUE #4771: Super user (root) pop-up do not appear with xpra desktop
**Describe the bug**
The "Super user" pop-up do not appear when a GUI program need super user (root) 
with `xpra desktop ...`

-dead link removed-

**To Reproduce**
Server:
`xpra desktop --start=xfce4-session ...` ( regular user )

Client:
`xpra attach ...`

 - Server OS: MX Linux 23.3 ( Debian bookworm )
 - Client OS: MX Linux 23.3 ( Debian bookworm )
 - Xpra Server Version:   V6.4.1-R0
 - Xpra Client Version:  V6.4.1-R0

**Additional context**
[Original Discussion](https://github.com/orgs/Xpra-org/discussions/4770)

huyhoang171106:
## Summary

In `xpra desktop --start=xfce4-session`, privilege-escalation dialogs (PolicyKit / “super user” pop-ups) fail to appear because the desktop session may start without a registered polkit authentication agent (or with startup ordering that leaves no active agent on the session bus). As a result, GUI apps that request root elevation silently fail to prompt. This is a session integration bug in desktop-mode startup rather than an app bug.

## Files changed

- `xpra/x11/server/seamless.py` (modified)

## Testing

Closes #4771

### DIFF
--- a/xpra/x11/server/seamless.py
+++ b/xpra/x11/server/seamless.py
@@ -7,6 +7,7 @@
 
 import os
 import signal
+import subprocess
 import sys
 from time import monotonic, sleep
 from collections import deque
@@ -123,6 +124,7 @@ class SeamlessServer(GObject.GObject, ServerBase):
         # for handling resize synchronization between client and server (this is not xsync!):
         self.last_client_configure_event = 0.0
         self.snc_timer = 0
+        self.polkit_agent_command: tuple[str, ...] = ()
 
     def init(self, opts) -> None:
         self.wm_name = opts.wm_name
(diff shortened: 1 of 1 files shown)


totaam:
This needs work: the copilot review makes some valid points, but most importantly, this is totally the wrong place to be adding this: `xpra/x11/server/seamless.py` is for seamless servers only.
Either polkit should be started for all non-shadow session types (and why not), or if it should only be started for desktop / monitor only - either way, this is not the place.

PR #4835, opened 2026-03-28
Fix NV12 OpenGL shader (4 bugs) and add self-test with CSC fallback

andrewachen:
## Summary
- Fix four bugs in the NV12 shader path that prevented it from ever working:
  `.a` swizzle, uniform binding, UV coordinate scaling, and texture allocation mismatch
- Add runtime GL self-test to detect broken `GL_LUMINANCE_ALPHA` drivers, with automatic
  CSC fallback to YUV420P (using #4833's `csc_libyuv` NV12→YUV420P)
- Add headless EGL GPU validation test (`tests/scripts/test_nv12_shader.py`)

## Context
@totaam suggested in #4829 that we add a GL self-test to detect broken NV12 support
and remove it from `RGB_MODES`. Implementing that test forced a deeper investigation
of the NV12 render pipeline, revealing that the shader had never actually worked —
it had four independent bugs (see commit message for details). All four validated on
GPU via headless EGL test.

A nice side effect: the MediaFoundation hardware decoder (#4828) can now output NV12
directly on Windows ARM64 (Adreno X1-85 on GLon12/D3D12 compat layer). We previously
believed `.a` was broken on that driver, but the failure was caused by bugs 2-4, not
a driver defect. Correct colors confirmed on the device.

## Test plan
- [x] GPU validation: `PYTHONPATH=. python3 tests/scripts/test_nv12_shader.py`
  (headless EGL, validates all 4 bugs on NVIDIA RTX A2000)
- [x] Docker unit tests pass (`tests/docker/run-tests.sh`)
- [x] Windows ARM64 (Adreno/GLon12): MF decoder → NV12 → GL shader, correct colors
- [ ] Linux with nvdec NV12 output: `XPRA_FORCE_VIDEO_PIXEL_FORMAT=NV12 xpra attach`
- [ ] Test `XPRA_OPENGL_NV12=0` forces CSC fallback path

Fixes #4829

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>
Sponsored-By: Netflix

Issue #4829, opened 2026-03-25
NV12 GL shader reads wrong UV channel due to GL_LUMINANCE_ALPHA


andrewachen:
*This issue was drafted by Anthropic's Claude (Opus 4.6).*

## Bug

The NV12-to-RGB fragment shader in `xpra/opengl/shaders.py` reads `.g` for the V channel:

```glsl
highp float u = (texture(UV, pos).r - 0.5);
highp float v = (texture(UV, pos).g - 0.5);
```

But the UV texture uses `GL_LUMINANCE_ALPHA` as both the internal format and data format (`xpra/opengl/backing.py` lines 120, 131). Per the OpenGL spec, `GL_LUMINANCE_ALPHA` maps:
- `.r = .g = .b = luminance` (byte 0 = U)
- `.a = alpha` (byte 1 = V)

So `.g` returns the luminance (U), not alpha (V). The shader gets U for both channels.

## Confirmed

Tested via EGL texture readback on Mesa/llvmpipe:

```
GL_LUMINANCE_ALPHA: [0,0] r= 10 g=  0 b=  0 a= 20
```

`.r = 10` (luminance = byte 0), `.g = 0` (not byte 1), `.a = 20` (alpha = byte 1).

The shader should read `.a` for V, not `.g`.

## Impact

Currently no visible impact because the only NV12 producer (nvdec) uses CUDA-GL interop and bypasses the texture upload path entirely. Any future CPU-based NV12 consumer would hit this bug.

Discovered while building a MediaFoundation hardware decoder (#4828) which outputs NV12 from the MFT.

## Potential Fix

Replace `GL_LUMINANCE_ALPHA` with `GL_RG8`/`GL_RG` (core OpenGL 3.0+), which unambiguously maps `.r = byte 0` and `.g = byte 1`. This would also require updating the NV12 subsampling divs and the shader's UV sampling coordinates.

Note: on the Windows D3D12 OpenGL compatibility layer (OpenGLOn12/Mesa), `GL_RG` with `GL_UNPACK_ROW_LENGTH` produced horizontal ghosting, likely due to the row length being interpreted as bytes instead of texels. This needs further investigation before deploying `GL_RG` on that platform.

totaam:
Adding some links to the source:
* shader:
https://github.com/Xpra-org/xpra/blob/2632a8f8b923e55d6f4f27c512c1ab6b571e1ee3/xpra/opengl/shaders.py#L121-L151
* internal format:
[...]

PR #4843, opened 2026-03-29
Use GL_RG for NV12 UV texture instead of GL_LUMINANCE_ALPHA (#4829)

andrewachen:
## Summary

Fixes the `GL_INVALID_VALUE` error reported in https://github.com/Xpra-org/xpra/pull/4835#issuecomment-4147743233 when painting NV12 on core GL 3.3+ profiles.

## Problem

After PR #4835 merged, @totaam reported `err = 1281` (`GL_INVALID_VALUE`) from `glTexSubImage2D` when painting NV12 frames on his NVIDIA Linux system:

```
glTexSubImage2D(GL_TEXTURE_RECTANGLE, 0, 0, 0, 300, 298,
                GL_LUMINANCE, GL_UNSIGNED_BYTE, ...)
```

The root cause: `GL_LUMINANCE` and `GL_LUMINANCE_ALPHA` were **removed from the OpenGL core profile in GL 3.3**. They only work on compatibility profiles.

This went undetected because our test environments all use compat profiles:
- **Windows** (WGL/GTK3) always gets a compatibility profile
- **Our Linux EGL tests** explicitly requested compat profile
- **totaam's Linux** (GLX/GTK3) — GTK3 requests a **core profile** first on Linux, which is why only his setup failed

## Fix

- Replace NV12 format constants: `GL_LUMINANCE`/`GL_LUMINANCE_ALPHA` → `GL_R8`/`GL_RG8` (internal) + `GL_RED`/`GL_RG` (data format). These are valid on both core and compat profiles.
- Change the NV12 shader V channel read from `.a` to `.g`. On `GL_RG`, byte 0 maps to `.r` (U) and byte 1 maps to `.g` (V). The previous `.a` swizzle was correct for `GL_LUMINANCE_ALPHA` (where byte 1 = alpha), but on `GL_RG`, `.a` is always 1.0.

Note: #4829 was originally filed about `.g` vs `.a` on `GL_LUMINANCE_ALPHA`. PR #4835 fixed it by using `.a`. This PR changes to `GL_RG` where `.g` is the correct swizzle — so the shader goes back to `.g`, but for a different format and a correct reason.

## Changes

- **`xpra/opengl/shaders.py`**: `.a` → `.g` in `gen_NV12_to_RGB()` V channel read
- **`xpra/opengl/backing.py`**: NV12 entries in `PIXEL_INTERNAL_FORMAT` and `PIXEL_DATA_FORMAT` changed to `GL_R8`/`GL_RG8` and `GL_RED`/`GL_RG`; texture width checks updated for `GL_RG`
- **`tests/scripts/gl_helpers.py`**: new shared GL test utilities (EGL context, shader compilation, NV12 rendering)
- **`tests/scripts/test_nv12_shader.py`**: updated to use `GL_R8`/`GL_RG` and core profile
- **`tests/scripts/test_nv12_core_profile.py`**: new regression test validating NV12 on both core and compat GL 3.3

## Test plan

- [x] `PYTHONPATH=. python3 tests/scripts/test_nv12_shader.py` — 4-bug regression (core profile, NVIDIA RTX A2000)
- [x] `PYTHONPATH=. python3 tests/scripts/test_nv12_core_profile.py` — core + compat profile validation
- [ ] `XPRA_FORCE_VIDEO_PIXEL_FORMAT=NV12 xpra attach --opengl=force` — integration test on core profile

Sponsored-By: Netflix

### DIFF
--- a/tests/scripts/gl_helpers.py
+++ b/tests/scripts/gl_helpers.py
@@ -1,0 +1,211 @@
+#!/usr/bin/env python3
+# This file is part of Xpra.
+# Copyright (C) 2026 Netflix, Inc.
+# Xpra is released under the terms of the GNU GPL v2, or, at your option, any
+# later version. See the file COPYING for details.
+# ABOUTME: Shared GL test utilities for headless EGL context creation,
+# ABOUTME: shader compilation, and NV12 texture rendering with readback.
+
+import os
+
+os.environ["PYOPENGL_PLATFORM"] = "egl"
+
+
+def create_egl_context(core=True):
+    """Headless GL 3.3 context via EGL device platform.
+
+    core=True requests a core profile (GL_LUMINANCE invalid).
+    core=False requests a compat profile (GL_LUMINANCE works).
+    """
+    import ctypes
+    from ctypes import pointer, c_int, c_void_p, POINTER, CFUNCTYPE
+    from OpenGL.EGL import (
+        eglGetProcAddress, eglInitialize, eglChooseConfig,
+        eglCreatePbufferSurface, eglBindAPI, eglCreateContext, eglMakeCurrent,
+        EGL_SURFACE_TYPE, EGL_PBUFFER_BIT, EGL_RED_SIZE, EGL_GREEN_SIZE,
+        EGL_BLUE_SIZE, EGL_ALPHA_SIZE, EGL_RENDERABLE_TYPE, EGL_OPENGL_BIT,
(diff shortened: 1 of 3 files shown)


totaam:
Excellent diagnostic and fix.

I've added a commit to remove `GL_LUMINANCE_ALPHA` from the backing class since it's no longer used, and should not be used with core profiles. (not pushed yet)

Issue #2466, opened 2019-10-29
GTK3 OpenGL win32 window is transparent


totaam:
Like it's not getting painted at all.

Tried 2.5.3 and 3.0.x on a win7 system with a Quadro M4000.

The code was originally added in [r16993](../commit/eed27ba9296562da861b5b7626bfe54244b74c8c) + [r16964](../commit/01a868a88f199916c2e6600f5c2244f7cbf848b7), see #1569#comment:5.

See also:
* #1569 python3 opengl client
* #1682 transparency with the native opengl backend on win32

totaam:
Turns out that we may be able to use [GTK.GLArea](https://lazka.github.io/pgi-docs/Gtk-3.0/classes/GLArea.html) with legacy contexts: [GTK Bug 756142 - Bring back support for legacy OpenGL contexts](https://bugzilla.gnome.org/show_bug.cgi?id=756142) using the undocumented env var `GDK_GL=legacy`. (patch hooks the widget)

As per [gdkglcontext-win32.c](https://github.com/GNOME/gtk/blob/master/gdk/win32/gdkglcontext-win32.c): OpenGL does not work with WS_EX_LAYERED enabled, so we need to disable WS_EX_LAYERED when we acquire a valid HGLRC
I've tried doing the same thing in the window style fixup code (removing `WS_EX_LAYERED`), but that did not help.

totaam:
The fix is incorrect, it ends up disabling opengl because this now triggers a rendering error:
```
Error presenting FBO
Traceback (most recent call last):
  File "E:\Xpra\trunk\src/xpra/client/gl/gl_window_backing_base.py", line 666, in present_fbo
    self.do_present_fbo()
  File "E:\Xpra\trunk\src/xpra/client/gl/gl_window_backing_base.py", line 691, in do_present_fbo
    glBindFramebuffer(GL_FRAMEBUFFER, 0)
  File "E:\Xpra\trunk\src\dist\lib\library.zip\OpenGL\error.py", line 223, in glCheckError
    raise GLError(
OpenGL.error.GLError: GLError(
        err = 1282,
        description = b'invalid operation',
        baseOperation = glBindFramebuffer,
        cArguments = (GL_FRAMEBUFFER, 0)
)
EndPaint
```

PR #4845, opened 2026-03-30
Fix GL window bleed-through on Win32 with Intel GPUs

andrewachen:
### Problem

On Win32 with Intel integrated GPUs, OpenGL windows show content from windows behind them bleeding through character/text backgrounds. Affects both native WGL and GtkGLArea rendering paths.

**Root cause**: Intel's `ChoosePixelFormat` returns a pixel format with 8 alpha bits even when `cAlphaBits=0` is requested. DWM reads undefined alpha from the window surface as transparent ([freedesktop.org bug #14165](https://bugs.freedesktop.org/show_bug.cgi?id=14165)).

Confirmed on Intel(R) Graphics, driver build 32.0.101.8508. Does not affect NVIDIA, AMD, or Adreno (ARM64/GLon12).

### Fix

The workaround is Intel+Win32-specific, gated on `GL_VENDOR` containing "Intel". The bulk of the workaround logic lives in `platform/win32/gl_context.py` to keep platform-specific hacks out of the shared OpenGL code. The shared `backing.py` only adds a no-op `_fixup_present_alpha()` hook that the Win32 code overrides.

**Intel+Win32 workarounds:**
- Force RGBA visual on GL windows so the GDI surface has a proper alpha channel (`client_window.py`)
- Force alpha in GtkGLArea readback so the surface gets alpha=255 (`glarea_backing.py`)
- Force alpha=1 via `glColorMask`+`glClear` before each present, with a one-time `glReadPixels` to prime the Intel driver — Intel's `glClear` doesn't properly respect `glColorMask` without a prior readback (`gl_context.py:setup_intel_workarounds`)

**Platform-independent:**
- Clear non-alpha FBOs with alpha=1 instead of alpha=0 (`backing.py`)
- Add `_fixup_present_alpha` hook in `do_present_fbo` (`backing.py`)
- Set `app_paintable` on GL DrawingArea — matches the Cairo path (`drawing_area.py`)

### Testing

Tested on Win32 x64 with Intel(R) Graphics (driver 32.0.101.8508):
- `opengl=force` (native WGL path) — no bleedthrough
- `opengl=force:glarea` (GtkGLArea path) — no bleedthrough
- `opengl=off` (Cairo) — unaffected, still works

Tested on Win32 ARM64 with Adreno X1-85 (GLon12) — no regression, workaround not activated.

Confirmed `WS_EX_LAYERED` is NOT set on any decorated GL window after the RGBA visual change.

### References
- [freedesktop.org #14165](https://bugs.freedesktop.org/show_bug.cgi?id=14165) — i965 glClear + glColorMask
- [Mozilla #1523030](https://bugzilla.mozilla.org/show_bug.cgi?id=1523030) — glBlitFramebuffer ignores glColorMask
- [Xpra #2466](https://github.com/Xpra-org/xpra/issues/2466) — related GTK3 OpenGL Win32 transparency

Sponsored-By: Netflix

### DIFF
--- a/xpra/client/gtk3/opengl/client_window.py
+++ b/xpra/client/gtk3/opengl/client_window.py
@@ -97,7 +97,14 @@ class GLClientWindowBase(ClientWindow):
         widget = super().new_backing(bw, bh)
         if self.drawing_area:
             self.remove(self.drawing_area)
-        set_visual(widget, self._has_alpha)
+        # On Intel Win32, force RGBA visual so the GDI surface has a proper
+        # alpha channel (Intel's pixel format has alpha bits even when none
+        # are requested, and DWM reads undefined alpha as transparent).
+        # _is_intel is None before detection, so `or False` ensures a clean bool.
+        from xpra.os_util import WIN32
+        from xpra.opengl.backing import GLWindowBackingBase
+        intel = WIN32 and (GLWindowBackingBase._is_intel or False)
+        set_visual(widget, self._has_alpha or intel)
         widget.show()
         self.init_widget_events(widget)
         if self.drawing_area and self.size_constraints:
--- a/xpra/client/gtk3/opengl/drawing_area.py
+++ b/xpra/client/gtk3/opengl/drawing_area.py
@@ -37,6 +37,7 @@ class GLDrawingArea(GLWindowBackingBase):
 
     def init_backing(self) -> None:
         da = Gtk.DrawingArea()
+        da.set_app_paintable(True)
         da.connect_after("realize", self.on_realize)
         # da.connect('configure_event', self.on_configure_event)
         # da.connect('draw', self.on_draw)
--- a/xpra/client/gtk3/opengl/glarea_backing.py
+++ b/xpra/client/gtk3/opengl/glarea_backing.py
@@ -22,7 +22,13 @@ def GLArea(alpha: bool) -> Gtk.GLArea:
     glarea = Gtk.GLArea()
     glarea.set_use_es(True)
     glarea.set_auto_render(False)
-    glarea.set_has_alpha(alpha)
+    # On Intel Win32, force alpha in readback so the GDI surface gets
+    # alpha=255 (freedesktop.org bug #14165).
+    # _is_intel is None before detection, so `or False` ensures a clean bool.
+    from xpra.os_util import WIN32
+    from xpra.opengl.backing import GLWindowBackingBase
+    intel = WIN32 and (GLWindowBackingBase._is_intel or False)
+    glarea.set_has_alpha(alpha or intel)
     glarea.set_has_depth_buffer(False)
     glarea.set_has_stencil_buffer(False)
     glarea.set_required_version(3, 2)
--- a/xpra/opengl/backing.py
+++ b/xpra/opengl/backing.py
@@ -53,7 +53,7 @@ from OpenGL.GL.ARB.framebuffer_object import (
     glGenFramebuffers, glDeleteFramebuffers, glBindFramebuffer, glFramebufferTexture2D, glBlitFramebuffer,
 )
 
-from xpra.os_util import gi_import
+from xpra.os_util import WIN32, gi_import
 from xpra.util.str_fn import repr_ellipsized, hexstr, csv
 from xpra.util.env import envint, envbool, first_time
 from xpra.util.objects import typedict
@@ -247,6 +247,7 @@ class GLWindowBackingBase(WindowBackingBase):
         "RGB", "BGR",
     )
     HAS_ALPHA: bool = GL_ALPHA_SUPPORTED
+    _is_intel: bool | None = None  # cached across all windows
 
     def __init__(self, wid: int, window_alpha: bool, pixel_depth: int = 0):
         self.wid: int = wid
@@ -411,7 +412,7 @@ class GLWindowBackingBase(WindowBackingBase):
         if self._alpha_enabled:
             glClearColor(0, 0, 0, 1)
         else:
-            glClearColor(1, 1, 1, 0)
+            glClearColor(1, 1, 1, 1)
         glClear(GL_COLOR_BUFFER_BIT)
         # copy offscreen to new tmp:
         self.copy_fbo(w, h, sx, sy, dx, dy)
@@ -569,6 +570,13 @@ class GLWindowBackingBase(WindowBackingBase):
 
         if self.gl_setup:
             return
+        if GLWindowBackingBase._is_intel is None:
+            from OpenGL.GL import glGetString, GL_VENDOR
+            vendor = (glGetString(GL_VENDOR) or b"").decode("utf-8", "replace").lower()
+            GLWindowBackingBase._is_intel = "intel" in vendor
+            if WIN32 and GLWindowBackingBase._is_intel:
+                from xpra.platform.win32.gl_context import setup_intel_workarounds
+                setup_intel_workarounds(GLWindowBackingBase)
         mt = get_max_texture_size()
         w, h = self.size
         if w > mt or h > mt:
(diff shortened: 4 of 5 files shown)


andrewachen:
This feels a little invasive and the Intel machine is not my daily driver, so I won't feel terrible if you decide not to merge this -- it's unfortunate that the Intel integrated GPU needs such things.

totaam:
Claude "max effort" review confirms some of my concerns:
> Freedesktop `#14165` is real but different. It's a 2009 Mesa/i965 DRI bug — glClear() with glColorMask() caused assertion failures in intel_clear_tris() on Linux. Fixed by restricting that path to front/back-left buffers only. It confirms Intel has a history of glClear+glColorMask misbehaviour, but it's a Linux Mesa driver fix, not the Windows ICD.                                                                                                                                                              

2009 on Linux...

> The "readback priming" claim has no documented basis found. No Mesa commits, no bug reports, no vendor documentation describes needing a glReadPixels to prime driver state before glColorMask+glClear works correctly on Windows Intel drivers. It appears to be empirically discovered — someone found it worked and shipped it.

Is this absolutely needed?

> The ChoosePixelFormat alpha-bits issue is well-known (the DWM bleed-through mechanism is documented), but the specific glReadPixels-as-primer workaround for Windows Intel isn't.

If it is well known, I would expect workarounds to be more common and easier than readback!

> Worth noting in the PR review: the code comment/docstring says the primer is needed because "Intel's glClear doesn't properly respect glColorMask without a prior readback", but no reference is cited for that specific claim on Windows. Given the 2009 Mesa fix, it's plausible there's a related Windows driver bug, but the PR would be stronger with a test case or driver version that demonstrates the primer is actually necessary — otherwise it's expensive insurance for an unverified assumption.

Claude was making good progress, but then:
> ` ⎿  You've hit your limit · resets 2am`

totaam:
@andrewachen can you try running 88d8abcddadf136cdec4ea9216f915e1f6553666 with `XPRA_OPENGL_ALWAYS_RGBA=1`?

Claude said:
_So `ALWAYS_RGBA=1` paired with the PR's FBO clear changes (alpha=0 → alpha=1) could in principle fix the bleed-through without any per-frame workaround — if Intel's `glBlitFramebuffer` correctly copies the alpha channel from the RGBA8 FBO to the window framebuffer. That's the unknown. If the blit doesn't propagate alpha (which is
  plausible for the same class of Intel driver bug), `ALWAYS_RGBA` alone still leaves the window framebuffer alpha undefined after the blit, and you're back to needing the `glColorMask`+`glClear` trick._

_Worth testing: it's a much lighter change than the per-frame workaround, but it may or may not be sufficient depending on whether Intel's blit honours the alpha source._

---

I found an Intel card, but it stopped working after a reboot!

totaam:
The Intel card is working again, turns out that this GPU is so big that it bent the SATA connectors out of their sockets, which prevented the OS from loading...

Anyway, I can test with Intel now, but I'm not seeing any issues. Which app can I used to reproduce the problem and how?

totaam:
@andrewachen please see https://github.com/Xpra-org/xpra/pull/4845#issuecomment-4294735836
This may already be fixed, and if not then I need help to reproduce this.

andrewachen:
> @andrewachen please see [#4845 (comment)](https://github.com/Xpra-org/xpra/pull/4845#issuecomment-4294735836) This may already be fixed, and if not then I need help to reproduce this.

I'll try to repo this over the next few weeks to see what next steps are. Thanks for the ping.

totaam:
I assume that the problem is fixed.

Simba98:
I found I can reproduce it on Xpra 6.5.3, the environment is:
- GPU: Intel UHD Graphics 770.
- OpenGL / driver: 4.6.0 - Build 32.0.101.7085.
- Backend: native Win32, double buffered, zerocopy available.
I tried to apply commit 5231ea2a98 as a patch on Xpra 6.5.3 Client, looks like still not working on Intel GPU. Maybe I need to check if the patch is logically correct.

And I can confirm Nvidia RTX3000 Series GPU on same server will not have same issue. This is should be related Intel (Again!).

However, we (my AI agents) noticed that there is [xpra/codecs/argb/argb.pyx](https://github.com/Xpra-org/xpra/blob/master/xpra/codecs/argb/argb.pyx) newly introduced, I guess totaam is working on it.
Hence I will not try to PR on it.

Some info:

Selected OpenGL log lines (startup plus the final local updates before ending the reproduction):
```text
2026-09-15 15:06:49,747 init_opengl(auto)
2026-09-15 15:06:50,399 OpenGL test rendering succeeded: {'success': True}
2026-09-15 15:06:50,399 OpenGL 4.6 enabled on 'Intel(R) UHD Graphics 770'
2026-09-15 15:06:50,399  using native win32 backend
2026-09-15 15:07:19,243 BGRX update at (1768,133) size 8x8 (256 bytes) to 8x8, using GL zerocopy:bytes-as-memoryview format=BGRA / UNSIGNED_BYTE to internal=RGB8
2026-09-15 15:07:19,247 swap_buffers: calling SwapBuffers(0x201106e)
2026-09-15 15:07:19,264 BGRX update at (1758,143) size 8x8 (256 bytes) to 8x8, using GL zerocopy:bytes-as-memoryview format=BGRA / UNSIGNED_BYTE to internal=RGB8
2026-09-15 15:07:19,266 swap_buffers: calling SwapBuffers(0x201106e)
2026-09-15 15:07:19,284 BGRX update at (1758,153) size 8x8 (256 bytes) to 8x8, using GL zerocopy:bytes-as-memoryview format=BGRA / UNSIGNED_BYTE to internal=RGB8
2026-09-15 15:07:19,287 swap_buffers: calling SwapBuffers(0x201106e)
2026-09-15 15:07:19,305 BGRX update at (1748,163) size 8x18 (576 bytes) to 8x18, using GL zerocopy:memoryview format=BGRA / UNSIGNED_BYTE to internal=RGB8
2026-09-15 15:07:19,314 swap_buffers: calling SwapBuffers(0x201106e)
2026-09-15 15:07:19,331 BGRX update at (1739,174) size 17x17 (1156 bytes) to 17x17, using GL zerocopy:memoryview format=BGRA / UNSIGNED_BYTE to internal=RGB8
2026-09-15 15:07:19,334 swap_buffers: calling SwapBuffers(0x201106e)
2026-09-15 15:07:19,351 BGRX update at (1719,184) size 26x37 (3848 bytes) to 26x37, using GL zerocopy:memoryview format=BGRA / UNSIGNED_BYTE to internal=RGB8
2026-09-15 15:07:19,354 swap_buffers: calling SwapBuffers(0x201106e)
```

`Xpra_cmd.exe opengl` output:
```text
GLU=1.2.2.0 Microsoft Corporation
accelerate=3.1.10
array-handlers=
backend=win32
core-profile=False
display_mode=DOUBLE
flags=
max-viewport-dims=16384, 16384
module=native
opengl=4.6
platform=win32
pyopengl=3.1.10
renderer=Intel(R) UHD Graphics 770
safe=True
shading-language-version=4.60 - Build 32.0.101.7085
success=True
texture-size-limit=16384
vendor=Intel
version=4.6.0 - Build 32.0.101.7085
zerocopy=True

```

totaam:
@Simba98 did you try 
* a recent beta build
* `XPRA_OPENGL_ALWAYS_RGBA=1`
* `opengl=force` and `opengl=no`

Simba98:
Quick answer, only tried: opengl=no with Xpra 6.5.3, and rendering correctly.
Let me see others.

Simba98:
@totaam As you requestsed:

`6.5.3 + opengl=no`: Good.

`6.5.3 + XPRA_OPENGL_ALWAYS_RGBA=1 + --opengl=auto` : Bad.
`6.5.3 + --opengl=force`: Bad.

`6.6-r42200 Client + 6.5.3 Server `: Good. 
`6.6-r42200 --opengl=auto --backend=gtk + 6.5.3 Server `: Bad, 
`6.6-r42200 --opengl=force --backend=gtk + 6.5.3 Server `: Bad, 
( so, is it due to GTK3->Win32 Native? )

totaam:
Yes, it looks that way. 
The gtk backend was always a fudge to bring gl rendering into the window. 
I guess we can leave this as fixed in 6.6

PR #4846, opened 2026-03-30
remove resize increment snap timer (#4831 follow-up)

andrewachen:
## Summary

Inevitably, as soon as you mentioned the history around death-spiral collapsing windows, I managed to experience it myself. It's not fun. Claude tried a bunch of options to make the timer stuff more resilient to this, but after a bunch of iterations, I gave up. This PR removes the timer, which is the root of all of this evil.

- Remove the debounced client-side snap timer added in #4831
- The timer tried to correct the client window size after WM handle resize, but caused a death spiral during OS DPI/scale changes: the timer's resize fought the OS adjustment, shrinking the window one increment per cycle
- The important fixes from #4831 remain: `snap_to_server_grid` in `send_configure_event` (prevents server bounce-back), `motion_moveresize` server-coord snap (alt-drag), `resize(force=True)` (server-authoritative sizes), `apply_geometry_hints` guard (wrong-grid WM snapping)
- The timer was a cosmetic fix (few-pixel gap on WM handle resize at non-integer desktop scales) that caused worse problems than it solved

🤖 Generated with [Claude Code](https://claude.ai/claude-code) (Anthropic Claude)

Sponsored-By: Netflix

### DIFF
--- a/xpra/client/gtk3/window/base.py
+++ b/xpra/client/gtk3/window/base.py
@@ -194,8 +194,6 @@ class GTKClientWindowBase(ClientWindowBase, Gtk.Window):
         self.moveresize_timer: int = 0
         self.moveresize_event: tuple[int, int, int, int, tuple[int], int, int, int, int] | tuple = ()
         self.moveresize_data: tuple[tuple, tuple] | tuple = ()
-        self.increment_snap_timer: int = 0
-        self.increment_snap_target: tuple[int, int, int, int] | None = None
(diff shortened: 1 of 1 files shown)


totaam:
Ah, I was curious why this timer was needed.
Asynchronous resizing is hard!

andrewachen:
Yeah, the idea was to enable reconciliation, but it ended up being a bunch of degenerate edge cases.

PR #4848, opened 2026-03-30
Fix nvenc encoder context leak from CUDA lock contention

andrewachen:
## Summary

- Fix nvenc encoder context leak caused by CUDA device lock contention during threaded cleanup
- `do_clean()` was `cdef void` — Cython silently swallowed `TransientCodecException` when `cuda_device_context.__enter__`'s non-blocking `lock.acquire(False)` failed during concurrent `compress_image()`, leaving `context_counter` permanently elevated
- Over time, leaked contexts accumulate (27 observed on a production server, using 6.7 GB VRAM), causing `get_runtime_factor()` to degrade and nvenc to lose pipeline scoring to x264 despite lower encode latency

### Changes

- Convert `do_clean()` from `cdef void` to `def` so exceptions propagate instead of being silently swallowed
- Use blocking lock acquisition (`timeout=5s`) instead of relying on `cuda_device_context.__enter__`'s non-blocking acquire, since cleanup runs in a daemon thread and can afford to wait
- Add `_force_context_release()` fallback if lock times out, ensuring `context_counter` is always decremented
- Wrap NVENC API calls in `cuda_clean()` with `try/finally` to guarantee counter decrement even if `nvEncDestroy*` calls fail

## Test plan

- [x] Unit test verifies counter is always decremented under lock contention (`test_nvenc_context_cleanup.py`)
- [x] Verified on production server: GPU memory dropped from 6.7 GB to 290 MB after fix, `runtime_factor` stays at 1.0 across multiple client reconnects, nvenc wins pipeline scoring (461 vs x264's 182)
- [ ] Needs Cython compilation to fully validate (tested via `cuda_device_context` override for local deployment)

Sponsored-By: Netflix

### DIFF
--- a/tests/unittests/unit/codecs/test_nvenc_context_cleanup.py
+++ b/tests/unittests/unit/codecs/test_nvenc_context_cleanup.py
@@ -1,0 +1,337 @@
+#!/usr/bin/env python3
+# This file is part of Xpra.
+# Copyright (C) 2026 Netflix, Inc.
+# Xpra is released under the terms of the GNU GPL v2, or, at your option, any
+# later version. See the file COPYING for details.
+# ABOUTME: Tests that nvenc encoder context cleanup always decrements the context
+# ABOUTME: counter, even when the CUDA device lock is held by another thread.
+
+"""
+Reproduces the nvenc context leak caused by lock contention during cleanup.
+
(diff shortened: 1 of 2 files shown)


totaam:
LGTM. Thanks!

Issue #4618, opened 2025-07-07
aom av1 decoder


totaam:
As an alternative to the dav1d decoder from #4465 which causes crashes with some streams..
This will help with #4612

[aom](https://aomedia.googlesource.com/aom)

totaam:
The current version of the code in git master can decode the sample frames we use for testing the decoders as long as we ignored the fact that it decodes to `YUV420P16` and not `YUV420P`, unlike dav1d (#4465) it doesn't have any problems decoding the 320x240 test data.
But it does choke on the nvenc data:
```
aom.init_context('av1', 300, 300, 'YUV420P')
codec: AOMedia Project AV1 Decoder v3.12.0
decompress_image(14498 bytes, typedict({'csc': 'YUV420P', 'frame': 0, 'pts': 0, 'full-range': True, \
    'csc-type': 'cuda:BGRX_to_NV12', 'type': 'IDR', 'quality': 99, 'encoder': 'nvenc', 'flush': 0, \
    'window-size': (300, 300), 'encoding': 'av1'}))
aom_codec_decode(..)=7
RuntimeError: AOM codec corrupt frame
```

totaam:
Working well!
Still TODO:
* avoid expensive memory copy of plane data - aom owns it, probably safe to reference until the next call to decode, but the painting is done in the UI thread whereas the decoding is done from the decode thread... may need to add a lock to the decoder
* decode to 16-bit for high bit depth displays / #4621

totaam:
b610d0e2354d54a00b1fcfa61b10fb1efbbd8549 does zero copy opengl upload from the aom buffers, but if the previous image has not been freed by the time we try to decode the next one... we busy wait.
This was easier than adding locking around access to the image wrapper's buffers.
And in any case, it seems that we consistently wait for about 5ms for the UI thread to run before we can decode, so the locking would have had to either:
* wait for the lock
* copy the pixel buffers

And neither of these solutions would be an improvement on the busy wait.

totaam:
FTB on Ubuntu Jammy:
```
xpra/codecs/aom/api.cpp: In function ‘int __pyx_pymod_exec_api(PyObject*)’:
xpra/codecs/aom/api.cpp:2788:47: error: ‘AOM_IMG_FMT_NV12’ was not declared in this scope; did you mean ‘AOM_IMG_FMT_YV12’?
[...]

Issue #4464, opened 2025-01-08
`libde265` decoder with YUV444 support


totaam:
This would help with artifacts with subsampling.
  
[`libde265`](https://github.com/strukturag/libde265) supports the following formats:
https://github.com/strukturag/libde265/blob/01bb88b7a518bedec28b9c9229ee7ee83d436832/libde265/de265.h#L316-L321
```c
enum de265_image_format {
  de265_image_format_mono8    = 1,
  de265_image_format_YUV420P8 = 2,
  de265_image_format_YUV422P8 = 3,
  de265_image_format_YUV444P8 = 4
};
```

[Decoder API Tutorial](https://github.com/strukturag/libde265/wiki/Decoder-API-Tutorial)

PR #4850, opened 2026-04-01
Fix nvenc h265 lossless support and prefer h265 over h264

andrewachen:
## Summary
- Fix `has_lossless_mode` in nvenc encoder spec to use probed `LOSSLESS_CODEC_SUPPORT` instead of hardcoding h264-only. This was causing h265 to get `quality=60` vs h264's `quality=100`, leading to unstable encoder flip-flopping as dynamic target quality fluctuated.
- Move h265 ahead of h264 in `PREFERRED_ENCODING_ORDER`. h265 was left at the bottom (after png, jpeg, rgb) as a legacy artifact from when it was in `PROBLEMATIC_ENCODINGS` (removed 2020, commit b0b1d12024).

## Details
The nvenc `_get_spec()` function (encoder.pyx:250) hardcoded `has_lossless_mode = encoding == "h264"`, overwriting the correctly probed value from line 247. This meant h265 got `quality=60+0=60` while h264 got `quality=60+40=100` in the VideoSpec, even though nvenc hardware supports lossless h265 (verified via `NV_ENC_CAPS_SUPPORT_LOSSLESS_ENCODE` probe at init time).

The quality spec difference caused the scoring to flip-flop: at target_quality ~67, h265's quality=60 was closer to target (+26 quality score points), but h264 had +17 from its higher position in `PREFERRED_ENCODING_ORDER`. As target quality fluctuated, the winner alternated, defeating the edge resistance mechanism.

### Why h265 ahead of h264
- Better compression at same quality on hardware encode paths
- No practical latency or power difference with hardware encode/decode
- No software h265 encoder exists in xpra — no footgun risk
- Clients without h265 hw decode don't advertise it, auto-fallback to h264
- av1 intentionally left low — has slow sw encoder (av1enc via gstreamer)

Sponsored-By: Netflix

### DIFF
--- a/xpra/codecs/constants.py
+++ b/xpra/codecs/constants.py
@@ -28,11 +28,11 @@ FAST_DECODE_MIN_SPEED: int = envint("XPRA_FAST_DECODE_MIN_SPEED", 70)
 # note: this is just for defining the order of encodings,
 # so we have both core encodings (rgb24/rgb32) and regular encodings (rgb) in here:
 PREFERRED_ENCODING_ORDER: Sequence[str] = (
-    "h264", "vp9", "vp8", "mpeg4",
+    "h265", "h264", "vp9", "vp8", "mpeg4",
     "mpeg4+mp4", "h264+mp4", "vp8+webm", "vp9+webm",
     "png", "png/P", "png/L", "webp", "avif",
     "rgb", "rgb24", "rgb32", "jpeg", "jpega",
-    "h265", "av1",
+    "av1",
     "scroll",
     "grayscale",
     "stream",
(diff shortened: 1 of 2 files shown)


totaam:
The `has_lossless_mode` part is correct for sure, the second part I have doubts - let's merge it and see what happens.

totaam:
FYI: claude said in 71ef8c80f3f7bcb7a0ed55e2d70d234b23690759 that "_av1 is intentionally left low in the order because it has a slow sw encoder (av1enc via gstreamer) that could be selected on systems without hw av1 encode_".
This is incorrect, we try hard not to use gstreamer video codecs, and we do have a module for av1 decoding now: #4618

totaam:
We could add a software `HEVC` decoder: #4464

PR #4852, opened 2026-04-01
Fix session info graphs not populating

andrewachen:
## Summary

- Fix typo in `getattr(self._protocol, "._conn", None)` → `"_conn"` (no dot)
- The erroneous dot meant the attribute was never found, so `conn` was always `None`
- `populate()` bailed out immediately at `if not conn: return False`, so no bandwidth or latency data was ever collected for the Graphs tab
- Introduced in commit 5e056cd32c ("simplify", May 2024)

Sponsored-By: Netflix


totaam:
Ooops!

PR #4855, opened 2026-04-01
Fix About dialog empty on second open

andrewachen:
## Summary
- `close_about()` only called `hide()`, but closing via the window manager's X button fires `delete-event` which destroys the widget
- The stale global reference then showed an empty window on re-open
- Now `destroy()` and clear the reference so a fresh dialog is created each time

Sponsored-By: Netflix


totaam:
I've gone for a slightly different approach in fe2cfca81fd256acd46d3293bc20a59c9955c4b8: the point using `close_about` is that we keep the about dialog around, preserving the link to the parent window if it has one.

Issue #3514, opened 2022-04-03
GTK widgets are too large on high resolution MS Windows displays


totaam:
See https://github.com/Xpra-org/xpra/issues/3482#issuecomment-1061336439
ie:
![157152678-68063d00-d415-4f93-bd26-5049e006817f](https://user-images.githubusercontent.com/1049737/161431578-864dc16e-553e-4f6a-b8bc-a4c66d2ff178.png)

totaam:
@bdklahn does this still happen with the beta builds and the patched dummy driver version 0.4.0?

bdklahn:
@totaam I don't know . . . Is this the latest beta?


[Xpra_Setup.exe](https://xpra.org/beta/windows/Xpra_Setup.exe)	2022-04-17 12:59	105M


![Capture](https://user-images.githubusercontent.com/920400/165659317-adce561a-5627-4ae6-a5b4-eaca314a0545.PNG)

totaam:
Oops, that's now fixed. Sorry about that.
In any case, newer builds are unlikely to fix this since the server is not involved, right? You're just running `Xpra_cmd` and getting this gui?

bdklahn:
I just tried again with 4.4

https://xpra.org/beta/windows/Xpra-x86_64_Setup.exe

2022-05-03 12:41 110M
![Xpra4 4](https://user-images.githubusercontent.com/920400/166747729-e73ebb7c-2f16-433c-ad89-61334059f098.PNG)
![AboutXpraCapture](https://user-images.githubusercontent.com/920400/166748086-f9f2d074-85c7-4c87-8912-5c0d01797a13.PNG)

totaam:
According to [GDK-Win32/4.0: Enable HiDPI support for Windows](https://anzwix.com/a/GTK/GDKWin3240EnableHiDPISupportForWindows), the support for HiDPI displays is only available in GTK4.
And since we're unlikely to ever port to GTK4 #1977 (too much Waylandization and levelling down)

We explicitly disable `GDK_SCALING` to ensure that GTK gives us the actual window dimensions. I tried allowing us to override this environment variable, but since it only supports whole integers, I can only use it to make things bigger, not smaller.

And there are no plans to support fractional scaling beyond fonts: https://gitlab.gnome.org/GNOME/gtk/-/issues/4345#note_1344625
[...]

PR #4856, opened 2026-04-01
Fix session info graph labels for HiDPI displays

andrewachen:
## Summary
- Replace Cairo toy text API with PangoCairo for DPI-aware graph text rendering
- Pango uses point-based font sizes scaled by the display's DPI resolution, so graph labels (title, Y axis, legend) render at the correct size on HiDPI displays
- Compute DPI scale from `Gdk.Screen.get_resolution()`, falling back to `get_scale_factor()`
- Measure actual Y label widths for left margin instead of fixed character-width estimate
- Skip Y axis labels that would overlap based on text height vs grid spacing
- Add `DEFAULT_DPI` constant (96) to `xpra.common`

Note: X11 GLX and macOS may need similar DPI detection changes but are not included since we cannot test them.

Sponsored-By: Netflix

### DIFF
--- a/xpra/common.py
+++ b/xpra/common.py
@@ -33,6 +33,9 @@ def gravity_str(v) -> str:
         return str(v)
 
 
+DEFAULT_DPI: int = 96
+
+
 def noop(*_args, **_kwargs) -> None:
     """ do nothing """
 
--- a/xpra/gtk/dialogs/session_info.py
+++ b/xpra/gtk/dialogs/session_info.py
@@ -1265,6 +1265,10 @@ class SessionInfo(Gtk.Window):
         if h <= 0:
             return True
         start_x_offset = min(1.0, (monotonic() - self.last_populate_time) * 0.95)
+        from xpra.common import DEFAULT_DPI
+        screen = Gdk.Screen.get_default()
+        dpi = screen.get_resolution() if screen else -1
+        scale = max(1, dpi / DEFAULT_DPI) if dpi > 0 else self.get_scale_factor()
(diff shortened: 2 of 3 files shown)


totaam:
LGTM!
What prompt did you use?
Can the same solution be used for #3514 ?

andrewachen:
*[Reply from Claude Code]*

The prompt was iterative — started with "the labels on the session info graph are really tiny at 1.6x desktop scaling, should the scaling also apply to the chart labels?" Claude Code explored the graph rendering code, identified the hardcoded Cairo font sizes, and we went through several approaches:

1. `set_device_scale` on the Cairo surface (broke GTK widget sizing on Win32)
2. `context.scale()` on the entire context (zoomed the graph data too, not just text)
3. Manual `*scale` on each font size (worked but ugly)
4. PangoCairo (the final approach — DPI-aware text rendering, proper text measurement)

Each iteration was tested on a Win32 client at 150% DPI scaling. The key discovery was that `Gtk.Widget.get_scale_factor()` returns 1 on Win32 with fractional DPI, so we had to use `Gdk.Screen.get_resolution()` instead.

Regarding #3514 — the same PangoCairo approach wouldn't fix that. #3514 is about GTK's own widgets (buttons, labels, layout) being too large on high-res Win32, which is GDK3/GTK3 over-scaling its own widget rendering — possibly double-applying DPI (Windows DPI awareness + GTK's own scaling). Our fix only affects custom Cairo-rendered text in `graph.py`; GTK widget sizing is handled by GTK's layout engine, which we can't control from xpra's side.

Issue #3104, opened 2021-04-30
windows 10 supports full size cursors


totaam:
According to [Make the windows cursor bigger on windows 8](https://stackoverflow.com/a/66050685/428751): _So if you pass in 0 for width and height and do not pass in LR_DEFAULTSIZE it will use the size of the resource. So if your cursor has a 64x64 image in it, it will display at that size even if the system pointer is 32x32._

So we may be able to remove the win32 specific workarounds I am referring to in https://github.com/Xpra-org/xpra/issues/2623#issuecomment-829765777

PR #4857, opened 2026-04-01
Fix cursor scaling on Win32 HiDPI displays

andrewachen:
## Summary
- Scale cursor image by desktop-scale (`xscale`/`yscale`) before fixed-size padding, so cursors like I-bars match the scaled remote content
- When the OS requires a fixed cursor size (currently Win32), the existing code only pads the cursor into the fixed canvas without scaling — so cursors appear too small when desktop-scale > 1
- The non-fixed-size path already handles desktop-scale in its else branch; the guard on `get_fixed_cursor_size` prevents double-scaling

Sponsored-By: Netflix

### DIFF
--- a/xpra/client/gtk3/client_base.py
+++ b/xpra/client/gtk3/client_base.py
@@ -985,6 +985,16 @@ class GTKXpraClient(GObjectClientAdapter, UIXpraClient):
                   encoding, xhot, yhot, serial, w, h, len(pixels))
         cursorlog("default cursor size is %s, maximum=%s", csize, (cmaxw, cmaxh))
         fw, fh = get_fixed_cursor_size()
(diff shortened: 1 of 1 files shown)


totaam:
I didn't like the way the code looked (duplicated scaling code paths) so I refactored it: 6a8a2aa2b1d02f04b9ba474876378a765f11369e

Does that still work for you?

andrewachen:
> I didn't like the way the code looked (duplicated scaling code paths) so I refactored it: [6a8a2aa](https://github.com/Xpra-org/xpra/commit/6a8a2aa2b1d02f04b9ba474876378a765f11369e)
> 
> Does that still work for you?

Yup, this works -- thank you cleaning this up!

totaam:
Not sure how that relates to #3104

Issue #3608, opened 2022-08-28
oneVPL encoder


totaam:
[oneAPI Video Processing Library](https://github.com/oneapi-src/oneVPL) is yet another encoding API that supports both CPUs and Intel GPUs.
Of particular interest is the performance of the AV1 encoder on the new ARC series of GPUs.
It supports older "Media SDK" Intel GPUs too, so this could even replace the `libva` via `ffmpeg` option.

totaam:
Much less of a priority now that we have #3706, so (temporarily?) removed in 3dbd4bfca13cb256ab6420432adfef2324e6e097

totaam:
Removed the stubs in 3dbd4bfca13cb256ab6420432adfef2324e6e097.
#3706 is making better progress.
Maybe we will re-visit this in the future.

PR #4860, opened 2026-04-07
add MediaFoundation hardware video decoder for Windows

### LINKED ISSUE #346: Hardware-accelerated video decoding on Windows using DxVA
*Issue migrated from trac ticket # 346*

**component:** client | **priority:** minor

#### 2013-05-27 16:45:29: ahuillet created the issue

___
> See https://lists.ffmpeg.org/pipermail/ffmpeg-user/2012-May/006600.html
> 
> More generic hardware decoding ticket: #202

### LINKED ISSUE #4828: MediaFoundation hardware video decoder for Windows
## Problem

The xpra Windows client decodes all video in software (openh264). On ARM64 laptops (Snapdragon X Elite), this creates a significant bottleneck:

- `pixels_decoded_per_second=80M` vs server's `pixels_encoded_per_second=201M` — client is 2.5x slower than the server
- `client-latency=158ms` — nearly 4 frames of pipeline lag at 23fps
- High CPU usage and battery drain on a battery-constrained device

Hardware video decode is available on these devices (confirmed via `edge://gpu`: H.264/HEVC/VP9/AV1 all supported via DXVA). However, GStreamer video is disabled on Windows builds, NVDEC requires an NVIDIA GPU, and no MediaFoundation decoder exists in xpra.

## Prototype

Anthropic's Claude (Opus 4.6) built a working prototype of a MediaFoundation DXVA hardware decoder module, starting from a detailed implementation plan that provided the architecture (thin C wrapper + Cython layer), data flow (NV12 → GL shader), file list, fragility assessment, and TDD implementation order. Claude implemented the code, debugged COM/D3D11 integration, discovered and worked around GL driver issues on the D3D12 OpenGL compatibility layer, and added multi-codec support through iterative testing cycles on Windows ARM64 hardware.

**Compare link:** https://github.com/andrewachen/xpra/compare/fix/win32-gl-cursor-v2...andrewachen:xpra:feature/mf-hw-decode-v3?expand=1

Note: this is based on a 6.4.x branch (`fix/win32-gl-cursor-v2`), not master. Opening as an issue for discussion before investing in a master port.

**Architecture:**
- `mf_decode.c` (~830 lines): C wrapper managing COM, D3D11, and MFT lifecycle. Creates a D3D11 device and DXGI device manager to enable DXVA hardware decode. Uses IMF2DBuffer for direct GPU staging texture access with ConvertToContiguousBuffer as fallback.
- `decoder.pyx` (~280 lines): Cython VideoDecoder protocol implementation. De-interleaves NV12 UV plane into separate U/V for YUV420P output (see "NV12 GL issue" below).

**Codecs supported:** H.264, HEVC, VP9, AV1 (any codec the platform's MFT supports with NV12 output)

## Measured Results (Snapdragon X Elite ARM64)

All measurements taken from `xpra info` and per-frame timing instrumentation in the prototype.

**Over wifi (1800x1060):**

| Metric | Before (openh264) | After (MF + DXVA) |
|---|---|---|
| Client latency | 158ms | 20ms |
| Per-frame decode | 25-85ms | 8-19ms |
| Server encoder selection | webp/x264 (penalized by high client latency) | nvenc H.265 at 50fps, full resolution |

**Over ethernet (2560x1440):**

| Metric | Value |
|---|---|
| Client latency | 7-8ms |
| Encoder | nvenc H.264 at 30fps, quality 99 |
| Resolution | Full (scaling 1:1) |
| `pixels_decoded_per_second` | 168M (up from 80M) |

Note: ~~the 30fps cap is due to Chromium's vsync behavior on the Xdummy virtual display~~ (**Claude hallucination** — the Chromium vsync explanation was fabricated). The 30fps cap appears to be a limitation of the Xdummy virtual display or xpra's frame pipeline — glxgears also caps around 24-25fps on the same Xdummy (xrandr shows 59.96Hz). This is not a client decode limitation. At 7ms client latency there is 26ms of headroom per frame at 60fps.

**Per-frame timing breakdown at 1808x1024 (from instrumented decode path):**
- `input` (ProcessInput): 20-50us
- `output` (ProcessOutput + GPU decode): 1-3ms
- `extract` (IMF2DBuffer Lock2D, GPU→CPU): 4-17ms (dominates)
- `copy` (NV12→YUV420P de-interleave + Python bytes): 0.4-1.8ms

The `extract` time dominates because the decoded D3D11 texture must be mapped to CPU memory for the GL texture upload. True zero-copy would require D3D11↔OpenGL texture sharing, which is not possible on ARM64 Windows where OpenGL runs on a D3D12 translation layer (Microsoft's OpenGLOn12/Mesa).

## Design Decisions

**MFT enumeration uses SYNCMFT, not HARDWARE:**
The inbox `MSH264DecoderMFT` is a sync MFT that uses DXVA internally — it does not register as `MFT_ENUM_FLAG_HARDWARE`. Using `MFT_ENUM_FLAG_HARDWARE` returns zero results on the Snapdragon X Elite. `MFT_ENUM_FLAG_SYNCMFT | MFT_ENUM_FLAG_SORTANDFILTER` finds it correctly.

**YUV420P output instead of NV12:**
The MFT outputs NV12 (Y plane + interleaved UV). When DXVA hardware decode is active, NV12 is the only 8-bit output format available — requesting I420/IYUV would force the MFT back to software decode (per [Microsoft's documentation](https://learn.microsoft.com/en-us/windows/win32/medfound/mft-decoder-expose-output-types-in-native-order)). The Cython layer de-interleaves the UV plane into separate U and V planes and outputs YUV420P. This works around two issues with the existing NV12 GL upload path:

1. `GL_LUMINANCE_ALPHA` (used for the NV12 UV texture) maps `.r = .g = .b = luminance, .a = alpha` per the OpenGL spec. The NV12 shader reads `.g` for V but gets luminance (= U). Confirmed via EGL texture readback test — `.g` always equals `.r` for `GL_LUMINANCE_ALPHA`. This is a pre-existing bug that doesn't affect nvdec since nvdec uses CUDA-GL interop and bypasses the texture upload path entirely.

2. Switching to `GL_RG`/`GL_RG8` (which correctly maps `.r` and `.g` to the two bytes) produced horizontal ghosting on the Windows D3D12 compat layer. The likely cause is `GL_UNPACK_ROW_LENGTH` being interpreted as bytes instead of texels for multi-byte formats — confirmed mathematically (stride 1856, `row_length=928` texels is correct per GL spec, but if treated as 928 bytes the effective stride is halved) but not yet confirmed in the Mesa D3D12 driver source.

The YUV420P path with `GL_RED` textures (1 byte per texel) has no ambiguity and works correctly on all drivers tested. The de-interleave costs ~1ms, negligible vs the 4-17ms extraction time.

**Graceful fallback:**
1. D3D11 device creation fails → software MFT decode (no DXVA)
2. MFT not found for a codec → `MF_DEC_NOT_AVAILABLE` → selftest removes that codec
3. MF module init fails entirely → `ImportError` → openh264 continues

**MF preferred over openh264 on Windows:**
openh264 only officially supports Constrained Baseline Profile (per [openh264 issue #1407](https://github.com/cisco/openh264/issues/1407)), while nvenc/x264 typically encode High Profile. MF supports Baseline/Main/High up to Level 5.1. The prototype registers MF first in `ALL_VIDEO_DECODER_OPTIONS` and `DECODER_VIDEO_CODECS`.

## AV1 Test Data Bug

While testing, we discovered the existing AV1 320x240 test sample in `checks.py` is inconsistent with the other AV1 test samples. Confirmed by parsing the OBU sequence headers:

- 64x64: `seq_profile=0` (4:2:0) ✓
- 320x240: `seq_profile=1` (4:4:4) ✗ — labeled as YUV420P in the test dict but the bitstream is actually 4:4:4
- 1920x1080 (from nvenc): `seq_profile=0` (4:2:0) ✓

The 320x240 sample was generated with `gst-launch-1.0 ... ! av1enc` without specifying `format=I420`, causing av1enc to default to 4:4:4 output. Since NV12 can only represent 4:2:0, decoders that output NV12 fail the selftest on this sample (stream change renegotiation finds no NV12 type available) while the other two samples pass — which removes AV1 from the active codec list entirely.

The prototype includes a regenerated 320x240 sample using `format=I420` to produce Profile 0 (4:2:0), consistent with the other samples. This fix benefits all AV1 decoders, not just the MF module.

## Known Limitations

- **GPU→CPU copy overhead:** The dominant cost (~4-17ms at 1080p) is `IMF2DBuffer::Lock2D` which maps the D3D11 staging texture to CPU memory. ANGLE with EGL D3D11 interop is the potential path forward for zero-copy, but would require migrating xpra's GL backing from desktop OpenGL to GLES.

- **Occasional minor color artifacts:** Small icons within larger video frames occasionally show brief color corruption, possibly due to GPU synchronization timing on the Adreno driver. The artifacts are transient and self-correcting.

- **NV12 GL upload is broken:** As noted above, the existing `GL_LUMINANCE_ALPHA` NV12 upload path has a pre-existing bug per the GL spec. This doesn't affect nvdec (CUDA-GL interop) but would affect any future CPU-based NV12 consumer.

## Questions for Discussion

1. **Is this the right approach?** The thin C wrapper + Cython layer follows the pattern of existing codecs (openh264, nvdec). Is there a preference for a different architecture?

2. **MF decoder priority:** Should MF always be preferred over openh264 on Windows, or should it only be preferred when DXVA hardware decode is available?

3. **Session info pipeline display:** The prototype also adds codec/decoder/renderer info to the Window Encoders section (e.g., "h265 (nvenc) → mf → OpenGL"). Should this be included or submitted separately?

4. **NV12 GL_LUMINANCE_ALPHA:** The existing NV12 shader has a pre-existing bug where `.g` reads luminance instead of V. This doesn't affect nvdec (CUDA-GL interop) but would affect any future CPU-based NV12 consumer. Should this be tracked separately?

5. **Build dependencies:** The module links against `-lmfplat -lmfuuid -lole32 -ld3d11 -ldxguid`. These are standard Windows SDK libraries available in MinGW/MSYS2. Any concerns?

andrewachen:
## Summary

New `xpra.codecs.mf` codec module: Windows Media Foundation DXVA hardware decoder for H.264, HEVC, VP9, and AV1.

Closes #4828
Closes #346

## Details

- C wrapper (`mf_decode.c`) manages COM, D3D11, and MFT lifecycle with DXVA hardware acceleration
- Cython `decoder.pyx` implements the VideoDecoder protocol, outputs NV12 directly
- Graceful fallback: D3D11 failure → software MF decode; MF failure → openh264 continues
- Registered first in decoder priority (High Profile support, native ARM64, DXVA hardware acceleration)
- Minimum 64x64 frame size to avoid D3D11 overhead for tiny images

Tested on Windows ARM64 (Snapdragon X Elite, Adreno X1-85):
- DXVA hardware decode active
- 2560x1440 decode in 6-13ms (sub-1ms decode, 5-12ms GPU→CPU extract)
- Client latency: 158ms → 20ms (8x improvement over wifi)

Session info and deferred decoder fix split into separate PRs (#4861, #4862).

Sponsored-By: Netflix

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,6 +132,7 @@ tests/unittests/test-file*
 /xpra/codecs/nvidia/*/*.cpp
 /xpra/codecs/*/*.cpp
 /xpra/codecs/*/*.c
+!/xpra/codecs/mf/mf_decode.c
 /xpra/codecs/*.c
 /xpra/gtk/*.c
 /xpra/gtk/dialogs/*.c
--- a/setup.py
+++ b/setup.py
@@ -317,6 +317,7 @@ vpx_encoder_ENABLED     = vpx_ENABLED
 vpx_decoder_ENABLED     = vpx_ENABLED
 amf_ENABLED             = pkg_config_version("1.0", "amf") or has_header_file("AMF/components/VideoEncoderVCE.h")
 amf_encoder_ENABLED     = amf_ENABLED
+mf_decoder_ENABLED      = DEFAULT and WIN32
 remote_encoder_ENABLED  = DEFAULT
 # opencv currently broken on 32-bit windows (crashes on load):
 webcam_ENABLED          = DEFAULT and not OSX and not WIN32
@@ -369,6 +370,7 @@ ENCODER_SWITCHES = [
 DECODER_SWITCHES = [
     "openh264_decoder",
     "nvdec", "nvjpeg_decoder",
+    "mf_decoder",
     "vpx_decoder", "webp_decoder", "pillow_decoder",
     "jpeg_decoder", "avif_decoder",
 ]
@@ -1722,6 +1724,7 @@ def clean() -> None:
     # these files would match the pattern of generated files, but they are not, so protect them:
     PROTECTED = [
         "xpra/buffers/memalign.c",
+        "xpra/codecs/mf/mf_decode.c",
         "xpra/platform/win32/setappid.cpp",
         "xpra/x11/gtk/gdk_x11_macros.c",
     ]
@@ -3008,6 +3011,10 @@ if amf_ENABLED:
     tace(amf_encoder_ENABLED, "xpra.codecs.amf.encoder", **amf_kwargs)
     toggle_packages(WIN32, "xpra.platform.win32.d3d11")
     tace(WIN32, "xpra.platform.win32.d3d11.device")
+toggle_packages(mf_decoder_ENABLED, "xpra.codecs.mf")
+if mf_decoder_ENABLED:
+    ace("xpra.codecs.mf.decoder,xpra/codecs/mf/mf_decode.c",
+        extra_link_args=("-lmfplat", "-lmfuuid", "-lole32", "-ld3d11", "-ldxguid"))
 toggle_packages(gstreamer_ENABLED, "xpra.gstreamer")
 toggle_packages(gstreamer_video_ENABLED, "xpra.codecs.gstreamer")
 toggle_packages(remote_encoder_ENABLED, "xpra.codecs.remote")
--- a/xpra/codecs/loader.py
+++ b/xpra/codecs/loader.py
@@ -36,6 +36,7 @@ NOWARN = [
     "filter_pillow",
     "dec_avif", "enc_avif",
     "enc_amf",
+    "dec_mf",
 ]
 
 SELFTEST = envbool("XPRA_CODEC_SELFTEST", True)
@@ -81,7 +82,7 @@ DECODER_CODECS: Sequence[str] = gfilt(f"dec_{x}" for x in (
     "pillow", "webp", "jpeg", "nvjpeg", "avif", "gstreamer",
 ))
 DECODER_VIDEO_CODECS: Sequence[str] = gfilt(autoprefix("dec", x) for x in (
-    "vpx", "openh264", "nvdec", "aom",
+    "mf", "nvdec", "vpx", "openh264", "aom",
 ))
 SOURCES: Sequence[str] = filt("v4l2", "evdi", "drm", "nvfbc")
 
@@ -280,6 +281,7 @@ CODEC_OPTIONS: dict[str, tuple[str, str, str, str]] = {
     "nvdec"         : ("nvdec decoder",     "nvidia.nvdec", "decoder", "Decoder"),
     "dec_gstreamer" : ("gstreamer decoder", "gstreamer",    "decoder", "Decoder"),
     "dec_aom"       : ("aom decoder",       "aom",          "decoder", "Decoder"),
+    "dec_mf"        : ("mf decoder",        "mf",           "decoder", "Decoder"),
     # sources:
     "v4l2"          : ("v4l2 source",       "v4l2",         "virtual", "VirtualWebcam"),
     "evdi"          : ("evdi source",       "evdi",         "capture", "EvdiDevice"),
--- a/xpra/codecs/mf/__init__.py
+++ b/xpra/codecs/mf/__init__.py
@@ -1,0 +1,6 @@
+# This file is part of Xpra.
+# Copyright (C) 2026 Netflix, Inc.
+# Xpra is released under the terms of the GNU GPL v2, or, at your option, any
+# later version. See the file COPYING for details.
+# ABOUTME: MediaFoundation codec package for Windows.
+# ABOUTME: Provides hardware-accelerated H.264 decoding via DXVA.
--- a/xpra/codecs/mf/decoder.pyx
+++ b/xpra/codecs/mf/decoder.pyx
@@ -1,0 +1,279 @@
+# This file is part of Xpra.
+# Copyright (C) 2026 Netflix, Inc.
+# Xpra is released under the terms of the GNU GPL v2, or, at your option, any
+# later version. See the file COPYING for details.
+
+# ABOUTME: MediaFoundation hardware video decoder for Windows (H.264, HEVC, VP9, AV1).
+# ABOUTME: Wraps the C mf_decode API to implement xpra's VideoDecoder protocol.
+
+#cython: wraparound=False
+
+from time import monotonic
+from typing import Any, Dict, Tuple
+from collections.abc import Sequence
+
+from xpra.codecs.constants import VideoSpec
+from xpra.util.objects import typedict
+from xpra.codecs.image import ImageWrapper
+from xpra.log import Logger
+log = Logger("decoder", "mf")
+
+from libc.stdint cimport uint8_t, uintptr_t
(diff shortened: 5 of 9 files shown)


xpra/client/gui/window/backing.py
```diff
@@ -86,14 +86,14 @@ def load_video() -> None:
...
         return
-    _loaded_video = True
     vh = getVideoHelper()
     for csc_in in vh.get_csc_inputs():
         CSC_OPTIONS[csc_in] = vh.get_csc_specs(csc_in)
     log("csc options: %s", CSC_OPTIONS)
     for encoding in vh.get_decodings():
         VIDEO_DECODERS[encoding] = vh.get_decoder_specs(encoding)
     log("video decoders: %s", VIDEO_DECODERS)
+    _loaded_video = bool(VIDEO_DECODERS)
```
totaam: Please submit separately, this is not related to the codec.

totaam (changes requested):
Minor edits + split and this should be good to go!

xpra/client/gui/window/backing.py
andrewachen: Makes sense, will split it out.

totaam:
FYI:
* 934ffb173e62e215931e01c56b3eb9e58925da58 fixes codecs errors when support is not available
* bded581b930ece6dc40304c52db2941f7a496965 provides better tuning

andrewachen:
Thanks for the selftest fix in 934ffb17 — that's a good defensive improvement, we'll pick it up.

I think bded581b might not be working as intended: `HARDWARE_DECODERS` is initialized as an empty list and nothing populates it, so `encoding in HARDWARE_DECODERS` is always `False`. All codecs get the software path values (`speed=50`, `setup_cost=0`, `cpu_cost=100`, `gpu_cost=0`) even when DXVA hardware decode is active. Is there a follow-up commit planned to populate it, or should the selftest be detecting hardware status via `mf_decoder_is_hardware()` and adding to the list?

Separately — we're also working on an Intel oneVPL HEVC decoder that handles YUV 4:4:4, which would let Intel GPU clients decode h265 without chroma subsampling. The server already encodes YUV444P at high quality via nvenc; this would preserve full chroma resolution on the client side, which is a visible improvement for text-heavy content like browsers. For code editors, it could mean that switching from webp/rgb24 to h265 4:4:4 during high-motion periods (outside of scrolling) becomes viable, since we wouldn't be compromising text quality.

totaam:
> I think bded581b930ece6dc40304c52db2941f7a496965 might not be working as intended..

Right, I was hoping to do something a bit cleaner, but 8cf35d5145c29705d82d8462f92f5310beddf8bd will do.

---

wrt oneVPL, I had started one: #3608
IIRC, there were threading issues at the time.

PR #4863, opened 2026-04-10
Detect WASAPI device changes and auto-restart audio pipeline

andrewachen:
## Summary

- Detect WASAPI audio device removal/switch via `IMMNotificationClient` COM interface (pure ctypes, no comtypes dependency)
- Auto-restart audio pipeline on device change with 200ms delay and exponential backoff
- Skip WASAPI unmute ramp (shared mode doesn't pop)
- Add low-latency sink attributes for wasapisink/wasapi2sink

Unlike PulseAudio/CoreAudio, WASAPI does not manage device routing — when the default output changes or headphones are unplugged, the `IAudioClient` handle is permanently invalidated and the GStreamer pipeline crashes. This adds detection and recovery.

This is a port from our v6.4.x fork where it has been running in production. It has been adapted to master's code structure but has not been integration-tested on master beyond unit tests.

Detection runs in the audio subprocess using a pure ctypes `IMMNotificationClient` COM vtable (`_win32_device_monitor.py`). We avoided `comtypes.COMObject` because it crashes in frozen builds due to GIL/COM thread conflicts. COM callbacks signal a Windows Event; a 100ms GLib timer polls it and emits an error to the client. The client restarts audio with a 200ms delay and exponential backoff on failure.

### New platform files

`platform/audio.py` and `platform/win32/audio.py` add `is_recoverable_audio_error()` using the existing `platform_import` pattern (same as `gui.py`, `keyboard.py`, `features.py`, etc.). This classifies WASAPI `DEVICE_INVALIDATED` (0x88890004) errors as recoverable so the client can restart instead of treating them as fatal.

We chose the platform module approach because:
- It follows xpra's established pattern for platform-specific behavior
- It's extensible — PulseAudio and CoreAudio have their own device removal semantics that could be added later
- It keeps platform-specific error strings out of the generic audio client code

If you'd prefer not to establish a new platform module for this, the Win32 check can be inlined directly in `audio_sink_error()` instead.

### Tests

The 3 new `DeviceInvalidationTest` tests pass (`is_recoverable_audio_error`, `device_changed_restarts`, `non_device_error_does_not_restart`). The existing `AudioClientReceiveTest` in the same file has a pre-existing import error (`No module named 'xpra.net.protocol.constants'` — that module moved to `xpra.net.packet_type`) unrelated to this PR.

Sponsored-By: Netflix

### DIFF
--- a/xpra/client/subsystem/audio.py
+++ b/xpra/client/subsystem/audio.py
@@ -1,5 +1,6 @@
 # This file is part of Xpra.
 # Copyright (C) 2010 Antoine Martin <antoine@xpra.org>
+# Copyright (C) 2026 Netflix, Inc.
 # Xpra is released under the terms of the GNU GPL v2, or, at your option, any
 # later version. See the file COPYING for details.
 
@@ -15,7 +16,7 @@ from xpra.net.compression import Compressed
 from xpra.net.packet_type import CONNECTION_LOST
 from xpra.common import noop, SizedBuffer, may_notify_client
 from xpra.constants import NotificationID
-from xpra.os_util import get_machine_id, get_user_uuid, OSX, POSIX
+from xpra.os_util import get_machine_id, get_user_uuid, WIN32, OSX, POSIX
 from xpra.util.objects import typedict
 from xpra.util.str_fn import csv, bytestostr
 from xpra.util.env import envint
@@ -27,6 +28,13 @@ avsynclog = Logger("av-sync")
 log = Logger("client", "audio")
 
 QUERY_SLEEP = envint("XPRA_AUDIO_QUERY_SLEEP", 0)
+
+
+def _is_recoverable_audio_error(error_str: str) -> bool:
+    if not WIN32:
+        return False
+    upper = error_str.upper()
+    return "DEVICE_INVALIDATED" in upper or "88890004" in upper
 
 AV_SYNC_DELTA = envint("XPRA_AV_SYNC_DELTA")
 DELTA_THRESHOLD = envint("XPRA_AV_SYNC_DELTA_THRESHOLD", 40)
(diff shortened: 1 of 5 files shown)


xpra/platform/win32/audio_device_monitor.py
```diff
@@ -0,0 +1,270 @@
...
+
+    if _event:
+        ctypes.windll.kernel32.CloseHandle(_event)
+        _event = None
+
+    _client = None
+    _vtbl = None
+    _on_change = None
+    _prevent_gc.clear()
+    log("audio device monitor stopped")
```
totaam, line 270: This file belongs in `xpra/platform/win32`.

xpra/client/subsystem/audio.py
```diff
@@ -549,11 +550,35 @@ def audio_sink_error(self, audio_sink, error) -> None:
...
-        self.may_notify_audio("Speaker forwarding error", estr)
-        log.warn("Error: stopping speaker:")
-        log.warn(" %s", estr)
+        if "AUDIO_DEVICE_CHANGED" in estr:
+            # audio subprocess detected a device change — restart quickly:
+            log.info("audio output device changed, restarting speaker")
+            self.stop_receiving_audio()
+            self.idle_add(self._restart_audio_after_device_change)
+            return
+        from xpra.platform.audio import is_recoverable_audio_error
```
totaam: It's overkill to add new modules just for this one function.
Just define it it at the top of this module and return `False` for non-`WIN32`.

totaam (changes requested):

andrewachen:
Addressed both review comments:

1. **`_win32_device_monitor.py`** → moved to `xpra/platform/win32/audio_device_monitor.py`
2. **Platform module pair removed** — deleted `platform/audio.py` + `platform/win32/audio.py`. Inlined `_is_recoverable_audio_error()` directly in `client/subsystem/audio.py` with a `WIN32` guard.

The `device_monitor.py` thin abstraction (in `xpra/audio/`) still imports from `xpra.platform.win32.audio_device_monitor` — this is the layer that provides the no-op on non-Win32.

Sponsored-By: Netflix

totaam:
Thanks!

Issue #4663, opened 2025-09-06
ssl connections drop, sometimes


totaam:
Hard to reproduce - I can't, but I've seen it.

It could be related to #4660 or even something similar to https://github.com/python/cpython/issues/137583

The server log and client output don't show any useful information, just the usual `Connection lost`.

totaam:
Changes that may have an impact:
* [gtk-osx-build: python 3.12](https://github.com/Xpra-org/gtk-osx-build/issues/32)
* [gtk-osx-build: openssl 3.5.2](https://github.com/Xpra-org/gtk-osx-build/commit/c5a52561d2ed910f5321df149cea39a0461b6f14)
* [gtk-osx-build: libffi 3.5.2](https://github.com/Xpra-org/gtk-osx-build/commit/5a678b97f0a3f687415bb0edd691c7b3e19286b3) for https://github.com/Xpra-org/gtk-osx-build/issues/27
* [gtk-osx-build: python 3.12.11](https://github.com/Xpra-org/gtk-osx-build/commit/0a1998c539c420a4a7d5e3367990738e637e02a6)
* [gtk-osx-build: pyobjc 11.1](https://github.com/Xpra-org/gtk-osx-build/commit/a2ee729f0fadef5e880a89d0f1ec8b25c7211c68)
* fcb29d01d643dd73e75a65ce991e3e40c6f689c4
* #4662
* #4651

Worth trying Python 3.11.13 instead: https://www.python.org/downloads/release/python-31113/

totaam:
The [6.3.3](https://github.com/Xpra-org/xpra/releases/tag/v6.3.3) was made after downgrading to [Python 3.11.13](https://github.com/Xpra-org/gtk-osx-build/commit/4e969259a91287555f2a9a71efdef98832762e9d)
So all the latest beta builds and v6.3.3 release builds are based on 3.11.13

totaam:
I have another similar issue reported privately. Also using ssl. Server is RHEL8.9, xpra v5.

The symptom is the same, the client's `-d network` log just shows:
```
read thread: eof
```

For debugging, see https://github.com/Xpra-org/xpra/issues/4553#issuecomment-2788797736


---

Some links on the websocket read calls below.
The websocket protocol adapter does not interfere with the read call:
​https://github.com/Xpra-org/xpra/blob/master/xpra/net/websockets/handler.py
As it only comes into play once the data has been read and needs parsing.
The read function is here:
[...]

Issue #4202, opened 2024-04-27
quic congestion control


totaam:
Improving upon #3376, see also #3854.

My understanding of [aioquic](https://github.com/aiortc/aioquic) packet loss is that [`QuicConnection`](https://github.com/aiortc/aioquic/blob/a6d91fbcb5d7660adb1597bb711f6f963c598e03/src/aioquic/quic/connection.py#L234)
Instantiates a [`QuicPacketRecovery`](https://github.com/aiortc/aioquic/blob/a6d91fbcb5d7660adb1597bb711f6f963c598e03/src/aioquic/quic/connection.py#L388-L397), which is a _Packet loss and congestion controller._
There is no easy way to override this class.
Loss detection happens in [`_detect_loss`](https://github.com/aiortc/aioquic/blob/a6d91fbcb5d7660adb1597bb711f6f963c598e03/src/aioquic/quic/recovery.py#L300)
Which calls [`_on_packets_lost`](https://github.com/aiortc/aioquic/blob/a6d91fbcb5d7660adb1597bb711f6f963c598e03/src/aioquic/quic/recovery.py#L353)
In turn, this calls [`delivery_handlers`](https://github.com/aiortc/aioquic/blob/a6d91fbcb5d7660adb1597bb711f6f963c598e03/src/aioquic/quic/recovery.py#L378-L379) (see below) and the [congestion controller](https://github.com/aiortc/aioquic/blob/a6d91fbcb5d7660adb1597bb711f6f963c598e03/src/aioquic/quic/recovery.py#L383)

The `congestion_control_algorithm` string comes from the [`QuicConfiguration`](https://github.com/aiortc/aioquic/blob/a6d91fbcb5d7660adb1597bb711f6f963c598e03/src/aioquic/quic/configuration.py#L19) and defaults to [`reno`](https://github.com/aiortc/aioquic/blob/a6d91fbcb5d7660adb1597bb711f6f963c598e03/src/aioquic/quic/configuration.py#L29).
The congestion control algorithm is instantiated by [`create_congestion_control`](https://github.com/aiortc/aioquic/blob/a6d91fbcb5d7660adb1597bb711f6f963c598e03/src/aioquic/quic/congestion/base.py#L109)
New implementations can be registered via [`register_congestion_control`](https://github.com/aiortc/aioquic/blob/a6d91fbcb5d7660adb1597bb711f6f963c598e03/src/aioquic/quic/congestion/base.py#L122)

[...]

Issue #3854, opened 2023-05-15
quic substreams add datagrams for some packet types


totaam:
Split from #3376.

Let's use substreams for some packet types, the stub is here:
https://github.com/Xpra-org/xpra/blob/d98af749bd0b88a0657a6055e39d56455b428901/xpra/net/quic/websocket.py#L20

For datagrams, there's already a stub in:
https://github.com/Xpra-org/xpra/blob/d98af749bd0b88a0657a6055e39d56455b428901/xpra/net/quic/connection.py#L26

This will probably require more changes to the packet format: #1942
Ideally, we want each subsystem to use its own packet prefix.

Issues:
* sometimes we can hit a `aioquic.h3.exceptions.NoAvailablePushIDError`: https://github.com/Xpra-org/xpra/issues/3850#issuecomment-1545433576 - or what this caused by a bug?
* ensure we only skip the packet types we can afford to miss (ie: pointer position)
* could we use a dedicated stream for video? (only when fullscreen / full window?)

Issue #3850, opened 2023-05-11
QUIC connection fails to start with "Idle timeout" from Windows to Ubuntu


jzyrobert:
**Describe the bug**
Trying to connect to an Xpra server over QUIC UDP is failing with `connection failed: Idle timeout`.
In the debug logs, the last lines before the error are:
```
CreatePipeHandle()=0x1f24
ConnectNamedPipe()=0
GetLastError()=Overlapped I/O operation is in progress. (997)
Traceback (most recent call last):
  File "E:/xpra/xpra/scripts/main.py", line 933, in connect_or_fail
  File "E:/xpra/xpra/scripts/main.py", line 1126, in connect_to
  File "E:/xpra/xpra/net/quic/client.py", line 226, in quic_connect
  File "E:/xpra/xpra/net/quic/asyncio_thread.py", line 100, in sync
Exception: Idle timeout
```
I've verified that the UDP connection should work by running `nc -ul 4000` on the server side and sending over data from the client.
I've also tested this over a wireguard VPN connection that is port-forwarding UDP 4000 and the logs show that it detects the test packets I send, but when Xpra is starting up no UDP packets are detected by wireguard.

**To Reproduce**
1. `xpra start --bind-quic=0.0.0.0:4000 --ssl-cert ssl-cert.pem --ssl-key key.pem --no-daemon --debug quic`
2. `.\Xpra_cmd.exe attach quic://IP_ADDRESS:4000/ --ssl-server-verify-mode=none --compress=0 --bandwidth-detection=no
 --speed=100 --debug quic --debug network`
3. Client cannot connect

**System Information (please complete the following information):**
 - Server OS: Ubuntu 22.04
 - Client OS: Windows 10
 - Xpra Server Version v5.0-r33336 (gd0fbf257e)
 - Xpra Client Version xpra v5.0-r33337 (ga7f97a3ad)

jzyrobert:
> @jzyrobert did you try to connect locally on the server? Even just `xpra info quic://127.0.0.1:4000/`?

Of the two servers I tried:
First one reports
```
no uvloop
Warning: failed to connect:
 connection failed: Idle timeout
```

Second reports
```
no uvloop
Exception in callback _SelectorDatagramTransport._read_ready()
handle: <Handle _SelectorDatagramTransport._read_ready()>
Traceback (most recent call last):
  File "/usr/lib/python3.10/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
[...]

PR #4864, opened 2026-04-10
QUIC substreams: independent streams for audio, video, and input

andrewachen:
## Summary

Route audio, video, and input to independent QUIC streams to prevent
head-of-line blocking, replacing the broken push-promise approach that was
disabled in May 2023 due to `NoAvailablePushIDError` (#3850).

Ported from my v6.4.x daily-driver branch where this has been running for several days as a daily driver via module overrides.

**Cellular benchmark results** (no audio pipe bypass — substreams + priority only):

| Sound network jitter | TCP | QUIC single | QUIC multi |
|---|---|---|---|
| stdev (run 1) | 80ms | 109ms | **20ms** |
| stdev (run 2) | 98ms | 157ms | **17ms** |
| max (run 1) | 817ms | 1249ms | **193ms** |
| max (run 2) | 708ms | 1906ms | **147ms** |

Single-stream QUIC is *worse* than TCP on cellular — H3's single stream has
stricter in-order delivery than TCP's kernel-level buffering. Multi-stream
eliminates this entirely: 4-5x better than TCP, 6-8x better than QUIC-single.

On WiFi/LAN with sufficient bandwidth, all modes perform equivalently since the
congestion window never fills.

- **Raw QUIC streams replace push promises** — each substream is a raw QUIC
  bidirectional stream identified by a `"xpra:<type>\n"` header prefix,
  allocated via `_quic.get_next_available_stream_id()` and `send_stream_data()`.
  Push promises were fundamentally limited by the H3 push ID pool; raw streams
  have no such limit.
- **Per-stream parse state isolation** — `ParseState` class (with `__slots__`)
  holds independent reassembly state per stream in the parse thread, preventing
  interleaving corruption when draw and sound data arrive simultaneously on
  different streams.
- **Stream priority reordering** — aioquic iterates `_streams` in dict insertion
  order with no priority mechanism. When video bandwidth spikes (e.g. window
  resize, 2→100 Mbps), the video stream consumes the entire congestion window,
  starving audio and input. `_prioritize_streams()` reorders the dict after each
  allocation: control streams first, then input (key, pointer), audio (sound),
  and bulk data (draw) last.
- **Client→server substreams** — keystrokes and pointer events bypass the main
  WebSocket write queue via dedicated QUIC streams, preventing input latency
  spikes when the queue is congested with clipboard or ack packets.
- **TLS error surfacing** — override `datagram_received()` to catch TLS/OpenSSL
  exceptions and resolve the connected waiter immediately instead of hanging
  until timeout with no error message. Both fast-open and non-fast-open paths
  use `asyncio.wait_for()` with the connect timeout.
- **certifi CA fallback** — for frozen Windows builds where the OS cert store
  isn't accessible to pyOpenSSL, falls back to `certifi.where()` and checks
  relative to the executable directory.
- **Zombie connection cleanup** — `ConnectionTerminated` event handler,
  `_close_dead_connection()` on write failure, and 60s read timeout loop ensure
  the xpra disconnect chain fires reliably when a peer dies.
- **UDP buffer sizing** — 4 MiB send/receive buffers reduce packet drops during
  Python GIL pauses.
- **HOL blocking benchmark** — `tests/scripts/test_quic_hol.py` connects as a
  real QUIC client and measures inter-arrival jitter with and without substreams
  enabled. Compares TCP, QUIC-single, and QUIC-multi in sequential or parallel
  (`--parallel`) mode.

### Why raw QUIC streams instead of H3 push promises

The existing substream stub used `send_push_promise()`, which maps to HTTP/3
server push. This has two problems: (1) the push ID space is limited and
exhausted quickly, causing `NoAvailablePushIDError` after a few streams, and
(2) push promises are semantically wrong — they're meant for speculative
resource delivery, not bidirectional data channels. Raw QUIC streams have none
of these limitations and are what the QUIC transport was designed for.

### Why WS framing is stripped in do_write() instead of conditionally skipped

The WebSocket frame header (`make_wsframe_header`) is added by the format thread,
but the stream routing decision happens on the asyncio thread. A conditional
"don't add framing for substreams" check in the format thread would race with
the stream allocation on the asyncio thread. Instead, framing is always added
(no cross-thread coordination needed), and `_strip_ws_header()` removes it in
`do_write()` for substream writes, which is already on the asyncio thread.

### Why accessing aioquic private attributes

`_prioritize_streams()` reorders `_quic._streams`, and the TLS error path reads
`_connected_waiter` and `_close_event`. aioquic provides no public API for stream
scheduling or error reporting during handshake. aioquic itself uses these
attributes the same way internally. The access is wrapped in try/except so a
future aioquic refactor degrades gracefully rather than crashing.

Configurable via environment variables:
- `XPRA_QUIC_SUBSTREAM_PACKET_TYPES` — server→client (default: `sound,draw`)
- `XPRA_QUIC_CLIENT_SUBSTREAM_PACKET_TYPES` — client→server (default: `key-,pointer`)

Addresses #3854. Partially addresses #4202 (stream priority is a first step
toward congestion-aware scheduling), #4663 (zombie cleanup and TLS error
surfacing fix silent connection drops), #3850 (TLS errors now surface
immediately instead of idle timeout).

Sponsored-By: Netflix

### DIFF
--- a/tests/scripts/test_quic_hol.py
+++ b/tests/scripts/test_quic_hol.py
@@ -1,0 +1,720 @@
+#!/usr/bin/env python3
+# This file is part of Xpra.
+# Copyright (C) 2026 Netflix, Inc.
+# Xpra is released under the terms of the GNU GPL v2, or, at your option, any
+# later version. See the file COPYING for details.
+# ABOUTME: Measures QUIC head-of-line blocking by connecting to an xpra server
+# ABOUTME: and timestamping packet arrivals with and without substreams enabled.
+
+"""
+QUIC Head-of-Line Blocking Test
+
+Connects to a live xpra server via QUIC and measures inter-arrival jitter
+of transport-level data chunks. Runs two passes:
+  1. Single-stream (no quic.substreams capability)
+  2. Multi-stream (with quic.substreams capability)
+
+Generate traffic on the xpra display first (e.g., glxgears), then run:
+
+    python3 test_quic_hol.py quic://HOST:PORT/ --password PASS
+
+Options:
+    --duration N      seconds per test pass (default: 10)
+    --single-only     only run single-stream test
+    --multi-only      only run multi-stream test
+"""
+
+import os
+import sys
+import time
+import argparse
+import statistics
+from collections import defaultdict
+
+# Add xpra source tree to path
+script_dir = os.path.dirname(os.path.abspath(__file__))
+src_root = os.path.abspath(os.path.join(script_dir, "..", ".."))
+if os.path.isdir(os.path.join(src_root, "xpra")):
+    sys.path.insert(0, src_root)
+
+
+class PacketTimingCollector:
+    """Collects timestamps for each packet category."""
+
+    def __init__(self):
+        self.arrivals: dict[str, list[float]] = defaultdict(list)
+        self.sizes: dict[str, list[int]] = defaultdict(list)
+        self.network_jitter: list[float] = []
+        self.start_time = 0.0
+        self.collecting = False
+
+    def start(self):
+        self.arrivals.clear()
+        self.sizes.clear()
+        self.network_jitter.clear()
+        self.start_time = time.monotonic()
+        self.collecting = True
+
+    def stop(self):
+        self.collecting = False
+
+    def record(self, category: str, size: int):
+        if self.collecting:
+            self.arrivals[category].append(time.monotonic())
+            self.sizes[category].append(size)
+
+    def inter_arrival_ms(self, category: str) -> list[float]:
+        times = self.arrivals.get(category, [])
+        if len(times) < 2:
+            return []
+        return [(times[i] - times[i - 1]) * 1000 for i in range(1, len(times))]
+
+    def report(self, label: str):
+        elapsed = time.monotonic() - self.start_time if self.start_time else 0
+        print(f"\n{'=' * 60}")
+        print(f"  {label}")
+        print(f"  Duration: {elapsed:.1f}s")
+        print(f"{'=' * 60}")
+
+        for cat in sorted(self.arrivals.keys()):
+            count = len(self.arrivals[cat])
+            total_bytes = sum(self.sizes[cat])
+            ia = self.inter_arrival_ms(cat)
+            if not ia:
+                print(f"\n  {cat}: {count} chunks ({total_bytes:,} bytes)")
+                continue
+            sorted_ia = sorted(ia)
+            rate = count / elapsed if elapsed > 0 else 0
+            print(f"\n  {cat}: {count} chunks ({total_bytes:,} bytes, {rate:.1f}/s)")
+            print(f"    inter-arrival (ms): "
+                  f"mean={statistics.mean(ia):.1f}  "
+                  f"p50={sorted_ia[int(len(ia) * 0.50)]:.1f}  "
+                  f"p95={sorted_ia[int(len(ia) * 0.95)]:.1f}  "
+                  f"p99={sorted_ia[min(int(len(ia) * 0.99), len(ia) - 1)]:.1f}  "
+                  f"max={max(ia):.1f}")
+            if len(ia) > 1:
+                print(f"    jitter (stdev): {statistics.stdev(ia):.1f} ms")
+
+        if self.network_jitter and len(self.network_jitter) > 1:
+            nj = sorted(self.network_jitter)
+            print(f"\n  sound network jitter (arrival_diff - send_diff):")
+            print(f"    samples={len(nj)}  "
+                  f"mean={statistics.mean(nj):.1f}ms  "
+                  f"p50={nj[int(len(nj) * 0.50)]:.1f}ms  "
+                  f"p95={nj[int(len(nj) * 0.95)]:.1f}ms  "
+                  f"max={max(nj):.1f}ms  "
+                  f"stdev={statistics.stdev(nj):.1f}ms")
+
+
+def _ensure_substream_support():
+    """Patch WebSocketClient with substream receive support if missing.
+
+    The server overrides add substream send/receive to connection.py, listener.py,
+    etc., but xpra.net.quic.client is client-only and may not be overridden.
+    Without this patch, the server routes draws to raw QUIC streams that the
+    client silently drops — resulting in zero draw traffic to measure.
+    """
+    from xpra.net.quic.client import WebSocketClient, ClientWebSocketConnection
+    if hasattr(WebSocketClient, '_handle_substream_data'):
+        return  # already patched or running from source with full support
+
+    # ClientWebSocketConnection may have inherited from the system
+    # XpraQuicConnection (without substream methods) due to import ordering.
+    # Patch the missing methods directly onto the class.
+    if not hasattr(ClientWebSocketConnection, '_raw_read_cb'):
+        ClientWebSocketConnection._raw_read_cb = None
+
+    if not hasattr(ClientWebSocketConnection, 'put_raw_substream_data'):
+        def put_raw_substream_data(self, data, stream_id=1):
+            if self._raw_read_cb:
+                self._raw_read_cb(data, stream_id)
+        ClientWebSocketConnection.put_raw_substream_data = put_raw_substream_data
+
+    from aioquic.quic.events import StreamDataReceived
+
+    def _is_server_bidirectional(self, stream_id):
+        from aioquic.quic.connection import stream_is_client_initiated, stream_is_unidirectional
+        return not stream_is_client_initiated(stream_id) and not stream_is_unidirectional(stream_id)
+
+    def _deliver_substream(self, websocket, data, stream_id):
+        """Deliver substream data to the connection's read path."""
+        websocket.put_raw_substream_data(data, stream_id)
+
+    def _handle_substream_data(self, event):
+        stream_id = event.stream_id
+        websocket = self._substream_map.get(stream_id)
+        if websocket:
+            self._deliver_substream(websocket, event.data, stream_id)
+            return True
+        if not self._is_server_bidirectional(stream_id):
+            return False
+        buf = self._substream_buffers.get(stream_id, b"") + event.data
+        if b"\n" not in buf:
+            self._substream_buffers[stream_id] = buf
+            return True
+        header, remainder = buf.split(b"\n", 1)
+        self._substream_buffers.pop(stream_id, None)
+        header_str = header.decode()
+        if not header_str.startswith("xpra:"):
+            return False
+        stream_type = header_str[5:]
+        websocket = next(iter(self._websockets.values()), None)
+        if not websocket:
+            return True
+        self._substream_map[stream_id] = websocket
+        print(f"  substream {stream_id} registered for {stream_type!r}")
+        if remainder:
+            self._deliver_substream(websocket, remainder, stream_id)
+        return True
+
+    _orig_quic_event = WebSocketClient.quic_event_received
+
+    def quic_event_received(self, event):
+        if isinstance(event, StreamDataReceived) and self._handle_substream_data(event):
+            return
+        _orig_quic_event(self, event)
+
+    _orig_init = WebSocketClient.__init__
+
+    def patched_init(self, *args, **kwargs):
+        _orig_init(self, *args, **kwargs)
+        if not hasattr(self, '_substream_map'):
+            self._substream_map = {}
+        if not hasattr(self, '_substream_buffers'):
+            self._substream_buffers = {}
+
+    WebSocketClient.__init__ = patched_init
+    WebSocketClient._is_server_bidirectional = _is_server_bidirectional
+    WebSocketClient._deliver_substream = _deliver_substream
+    WebSocketClient._handle_substream_data = _handle_substream_data
+    WebSocketClient.quic_event_received = quic_event_received
+    print("  (patched WebSocketClient with substream receive support)")
+
+
+def run_test(url: str, password: str, duration: float, substreams: bool, sync_interval: int = 0) -> tuple:
+    """Connect to the server, collect transport-level timing.
+
+    Returns (parse_collector, quic_collector) — the parse collector timestamps
+    packets after the parse thread processes them, the quic collector timestamps
+    raw substream data as it arrives from the QUIC layer (before parse thread).
+    """
+    from xpra.net.packet_encoding import init_all as init_encoders
+    from xpra.net.compression import init_all as init_compressors
+    init_encoders()
+    init_compressors()
+    if substreams:
+        _ensure_substream_support()
+
+    from gi.repository import GLib
+    from xpra.client.base import features
+    features.file = features.printer = features.control = features.debug = False
+
+    from xpra.client.base.command import MonitorXpraClient
+    from xpra.util.objects import typedict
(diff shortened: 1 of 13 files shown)


totaam:
I'm going to need time to review this one as this is quite involved.

I was actually in the process of doing some work on QUIC: #4306 and #3854 - I added 2439782796201ee76377ceafe49e8d198f529604 as a stepping stone towards handling audio using datagrams rather than separate streams.

andrewachen:
No worries — this ended up being more involved than I expected to get good performance on cellular.

Makes sense on datagrams for audio — the stream infrastructure in this PR is still needed for draw and input prioritization which require reliable, ordered delivery.

One thing worth considering on the datagrams-for-audio tradeoff: we also have an adaptive jitter buffer with WSOLA time-stretching (libsonic/SoundTouch) that absorbs retransmit delays gracefully — it slightly speeds up or slows down playback (±2.5%) to smooth jitter instead of underrunning. With that in place, stream retransmit delay becomes a non-issue for audio, while datagram loss produces gaps that can't be recovered without PLC. On LAN it doesn't matter, but on WiFi/cellular where the congestion window fills and datagrams get dropped, streams + tempo may actually produce better audio than datagrams + silence gaps.

Should I open a PR for the time-stretching? For my use case, slightly stretched audio is less distracting than dropouts.

totaam:
> we also have an adaptive jitter buffer with WSOLA time-stretching (libsonic/SoundTouch) that absorbs retransmit delays gracefully

That's exactly what I've always wanted to implement but couldn't figure out how with gstreamer.

> Should I open a PR for the time-stretching?

Oh! Yes please.

> For my use case, slightly stretched audio is less distracting than dropouts.

I think that this is true for most use-cases. And will definitely help keep audio running smoothly with low latency datagrams.

totaam (commented):
I've merged some minor parts of this huge change: f902d7c8faa2181bc94d436e29533b22132e037b + 302d25684a261ea1d377d1395ee078f409b5e47f

---

See also #4873 which may interfere with packet re-assembly.

FYI: I was actually thinking of going a different route with quic substreams and perhaps using a dedicated socket parser, without transport layer compression, without "raw-packets" support, without a lot of the cruft that has accumulated in there...
But maybe this is a better approach, the `ParseState` abstraction looks like it would be worth having.

---

https://github.com/Xpra-org/xpra/pull/4864/changes/0225911232c61ce5e05f675b15dacb678263777d looked worthwhile so I tried to fix the merge conflicts, but it doesn't work for me and I still hit the timeout with a self-signed cert:
```
2026-04-22 23:09:26,890 [3b3501892c6fc0b1] Error: 298, reason: self-signed certificate, frame_type: 6
2026-04-22 23:09:28,769 created unix domain sockets:
2026-04-22 23:09:28,769  '/run/user/1000/xpra/clients/fedora-251604'

Warning: failed to connect:
 failed to connect: self-signed certificate, connection timed out
```

What am I doing wrong?

Here is the patch I was using:
```patch
diff --git a/xpra/net/quic/client.py b/xpra/net/quic/client.py
index ac7e8f453a..093909dec8 100644
--- a/xpra/net/quic/client.py
+++ b/xpra/net/quic/client.py
@@ -1,10 +1,12 @@
 # This file is part of Xpra.
 # Copyright (C) 2022 Antoine Martin <antoine@xpra.org>
+# Copyright (C) 2026 Netflix, Inc.
 # Xpra is released under the terms of the GNU GPL v2, or, at your option, any
 # later version. See the file COPYING for details.
 
 import re
 import os
+import asyncio
 import socket
 import ipaddress
 from queue import SimpleQueue
@@ -44,6 +46,29 @@ log = Logger("quic")
 
 HttpConnection = H0Connection | H3Connection
 
+
+def format_tls_error(exc: Exception) -> tuple[int, str]:
+    """Map a TLS/OpenSSL exception to (ExitCode, user-friendly message)."""
+    try:
+        from OpenSSL.crypto import Error as OpenSSLError
+    except ImportError:
+        OpenSSLError = None
+    if OpenSSLError and isinstance(exc, OpenSSLError):
+        # OpenSSL errors are lists of (lib, func, reason) tuples
+        reasons = " ".join(r[2] for r in exc.args[0] if r[2]).strip()
+        if "no such file" in reasons:
+            return ExitCode.SSL_FAILURE, f"SSL CA certificate file not found: {reasons}"
+        if "hostname mismatch" in reasons:
+            return ExitCode.SSL_CERTIFICATE_VERIFY_FAILURE, f"SSL certificate hostname mismatch: {reasons}"
+        if "certificate verify failed" in reasons:
+            return ExitCode.SSL_CERTIFICATE_VERIFY_FAILURE, f"SSL certificate verify failed: {reasons}"
+        return ExitCode.SSL_FAILURE, f"SSL error: {reasons}"
+    msg = str(exc)
+    if not msg or msg == str(type(exc)):
+        return ExitCode.SSL_FAILURE, "SSL handshake failed"
+    return ExitCode.SSL_FAILURE, f"SSL handshake failed: {msg}"
+
+
 IPV6 = socket.has_ipv6 and envbool("XPRA_IPV6", True)
 PREFER_IPV6 = IPV6 and envbool("XPRA_PREFER_IPV6", POSIX)
 HOSTS_PREFER_IPV4 = os.environ.get("XPRA_HOSTS_PREFER_IPV4", "localhost,127.0.0.1").split(",")
@@ -189,13 +214,25 @@ class WebSocketClient(QuicConnectionProtocol):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._http: HttpConnection | None = None
-        self._push_types: dict[str, int] = {}
         self._websockets: dict[int, ClientWebSocketConnection] = {}
+        self._tls_error: Exception | None = None
         if self._quic.configuration.alpn_protocols[0].startswith("hq-"):
             self._http = H0Connection(self._quic)
         else:
             self._http = H3Connection(self._quic)
 
+    def datagram_received(self, data, addr) -> None:
+        try:
+            super().datagram_received(data, addr)
+        except Exception as e:
+            log("datagram_received TLS error", exc_info=True)
+            self._tls_error = e
+            # resolve the connected waiter so we don't hang until timeout
+            if self._connected_waiter is not None:
+                waiter = self._connected_waiter
+                self._connected_waiter = None
+                waiter.set_exception(e)
+
     def open(self, host: str, port: int, path: str) -> ClientWebSocketConnection:
         log(f"open({host}, {port}, {path})")
         stream_id = self._quic.get_next_available_stream_id()
@@ -303,6 +340,31 @@ def _make_quic_configuration(ssl_cert: str, ssl_key: str, ssl_key_password: str,
         max_datagram_frame_size=MAX_DATAGRAM_FRAME_SIZE,
     )
     configuration.verify_mode = parse_ssl_verify_mode(ssl_server_verify_mode)
+    if not ssl_ca_certs or ssl_ca_certs == "default":
+        # "default" means "use OS cert store" — handled by ssl.load_default_certs()
+        # for TCP+SSL, but aioquic uses pyOpenSSL which has no equivalent.
+        # Find the CA bundle ourselves instead.
+        # certifi.where() may return a stale build-time path in cx_Freeze frozen
+        # builds, so also check relative to the executable.
+        ssl_ca_certs = ""
+        ca_candidates = []
+        try:
+            import certifi
+            ca_candidates.append(certifi.where())
+        except ImportError:
+            pass
+        # frozen build: cacert.pem is next to the executable in lib/certifi/
+        import sys
+        exe_dir = os.path.dirname(getattr(sys, "executable", ""))
+        if exe_dir:
+            ca_candidates.append(os.path.join(exe_dir, "lib", "certifi", "cacert.pem"))
+        for ca_path in ca_candidates:
+            if ca_path and os.path.exists(ca_path):
+                ssl_ca_certs = ca_path
+                log(f"using CA bundle: {ssl_ca_certs}")
+                break
+        else:
+            log("no CA bundle found, certificate verification may fail")
     if ssl_ca_certs:
         configuration.load_verify_locations(ssl_ca_certs)
     if ssl_cert:
@@ -338,42 +400,49 @@ def _quic_connect(create_protocol, host: str, port: int, path: str, fast_open: b
         addr = addr_info[4]  # ie: ('192.168.0.10', 10000)
         log(f"connecting from {pretty_socket(local_addr)} to {pretty_socket(addr)} with {fast_open=}")
         if fast_open:
+            # fast-open: queue handshake, then open() triggers transmit
+            # sending ClientHello + websocket headers together
             protocol.connect(addr, transmit=False)
+            conn = protocol.open(host, port, path)
         else:
             protocol.connect(addr)
+        from xpra.scripts.picker import CONNECT_TIMEOUT
         log("awaiting connected state")
         try:
-            if not fast_open:
-                await protocol.wait_connected()
-            else:
-                from xpra.scripts.picker import CONNECT_TIMEOUT
-                tl.call_later(CONNECT_TIMEOUT, verify_connected, protocol)
-            log(f"{protocol}.open({host!r}, {port}, {path!r})")
-            conn = protocol.open(host, port, path)
-            log(f"quic connection {conn}")
-            return conn
+            # check for TLS errors that arrived before we started waiting
+            if protocol._tls_error:
+                raise protocol._tls_error
+            await asyncio.wait_for(protocol.wait_connected(), timeout=CONNECT_TIMEOUT)
+        except asyncio.TimeoutError:
+            # check if a TLS error was captured before the timeout
+            if protocol._tls_error:
+                exit_code, msg = format_tls_error(protocol._tls_error)
+                raise InitExit(exit_code, msg) from None
+            raise InitExit(ExitCode.CONNECTION_FAILED, "connection timed out") from None
         except Exception as e:
             log("connect()", exc_info=True)
-            # try to get a more meaningful exception message:
-            einfo = str(e)
-            if not einfo:
-                quic_conn = getattr(protocol, "_quic", None)
-                if quic_conn:
-                    close_event = getattr(quic_conn, "_close_event", None)
-                    log(f"{close_event=}, {dir(close_event)}")
-                    if close_event:
-                        err = close_event.error_code
-                        msg = close_event.reason_phrase
-                        if err & QuicErrorCode.CRYPTO_ERROR:
-                            raise InitExit(ExitCode.CONNECTION_FAILED, msg)
-                        # if (err & 0xFF)==QuicErrorCode.CONNECTION_REFUSED:
-                        #    raise InitExit(ExitCode.CONNECTION_FAILED, msg)
-                        raise RuntimeError(close_event.reason_phrase) from None
-            raise RuntimeError(str(e)) from None
-
-    # protocol.close()
-    # await protocol.wait_closed()
-    # transport.close()
+            # prefer the TLS error captured by datagram_received
+            if protocol._tls_error:
+                exit_code, msg = format_tls_error(protocol._tls_error)
+                raise InitExit(exit_code, msg) from None
+            # check for server-initiated CRYPTO_ERROR in the close event
+            close_event = getattr(protocol._quic, "_close_event", None)
+            if close_event:
+                err = close_event.error_code
+                msg = close_event.reason_phrase
+                if err & QuicErrorCode.CRYPTO_ERROR:
+                    raise InitExit(ExitCode.SSL_CERTIFICATE_VERIFY_FAILURE,
+                                   msg or "SSL certificate verification rejected by server") from None
+                if msg:
+                    raise InitExit(ExitCode.CONNECTION_FAILED, msg) from None
+            exit_code, msg = format_tls_error(e)
+            raise InitExit(exit_code, msg) from None
+        if not fast_open:
+            log(f"{protocol}.open({host!r}, {port}, {path!r})")
+            conn = protocol.open(host, port, path)
+        log(f"websocket connection {conn}")
+        return conn
+
     if len(addresses) == 1:
         return tl.sync(connect, addresses[0])
 
@@ -401,10 +470,3 @@ def quic_connect(host: str, port: int, path: str, fast_open: bool,
         return client_class(QuicConnection(configuration=configuration))
 
     return _quic_connect(create_protocol, host, port, path, fast_open)
-
-
-def verify_connected(protocol) -> None:
-    # this is only useful for debugging,
-    # as we have no way to send a message to the main thread,
-    # I guess it may still show up in some debug log?
-    log("verify_connect(%s) connected=%s", protocol, protocol._connected)
```

andrewachen:
Thanks for picking up the UDP buffer + `pyopenssl_ca_certs` pieces — both look good in their merged form.

## The "still hits the timeout" symptom

The TLS surfacing is actually firing — the `[3b3501892c6fc0b1] Error: 298, reason: self-signed certificate, frame_type: 6` line in your log is `aioquic.quic.connection.QuicConnection.receive_datagram` logging the `QuicConnectionError` (CRYPTO_ERROR + alert 42 = 0x12A = 298) before it calls `self.close()`. `_close_event` is set on the `QuicConnection`, the closing period ends, `_process_events` calls `waiter.set_exception(ConnectionError)`, and `wait_connected()` raises within ~130 ms.

What you're seeing is `_quic_connect`'s tail csv-joining the result of each address attempt:

```python
errors = []
for address in addresses:
    try:
        return tl.sync(connect, address)
    except Exception as ce:
        ...
        errors.append(estr)
raise InitExit(ExitCode.CONNECTION_FAILED, "failed to connect: " + csv(errors))
```

When `get_address_options` returns IPv4 + IPv6 (typical for `localhost` or any A+AAAA name) and your server only binds one family:

- bound address: close-event path → `"self-signed certificate"`
- unbound family: `asyncio.wait_for(wait_connected(), timeout=CONNECT_TIMEOUT)` raises `asyncio.TimeoutError` → `"connection timed out"`

Both strings get csv-joined into the final `InitExit`, producing exactly `"failed to connect: self-signed certificate, connection timed out"`. The TLS error is being reported correctly — it's just sitting next to an unrelated second-address timeout that makes it look broken.

Quickest way to confirm on your side: `XPRA_IPV6=0` (or `-d quic` to see both `connect(addr_info)` calls in the log). With a single resolved address you'll see a clean `"self-signed certificate"` and no trailing timeout. I reproduced this against aioquic 0.9.25 with a v4-only test server.

One follow-up worth considering: short-circuit the address loop when the first address fails with `SSL_FAILURE` / `SSL_CERTIFICATE_VERIFY_FAILURE`. That would drop the misleading tail for anyone debugging a cert issue against a multi-address hostname. The downside is it removes a valid failover path — if a hostname resolves to multiple real endpoints (multi-region, anycast, partial DNS rollout) and only one has a stale cert, the current code retries the next address and can recover; a short-circuit would bail before reaching the working endpoint. The trade-off is "clean error message for the common single-server case" vs. "preserve failover for split-brain deployments". I lean toward leaving the loop alone and letting the user disambiguate from the error string.

## #4873 (`recv_into` / `FrameState`)

Good flag. The per-substream parse work here keeps its own small buffer per substream and shouldn't conflict with a `recv_into` switch on the underlying socket, but `do_read_parse_thread_loop`'s `FrameState` would replace the same accumulated locals that the per-substream `ParseState` is mirroring. I'll re-test against #4873 once it lands and consolidate the two parse-state types if the shape lines up.

## Direction (dedicated socket parser vs. this approach)

I hear you on the WebSocket-over-QUIC weight. One thing to note: in the substreams configuration the high-volume traffic (video, audio, input) is delivered raw on dedicated QUIC streams — transport-layer compression and the raw-packets opcode aren't on that hot path anymore, so for that deployment shape a parser rewrite wouldn't change the runtime profile much. That's not universal though: anything that stays on the legacy WebSocket-over-QUIC path (non-substream sessions, or control traffic in this PR's setup) still pays H3 framing + WS opcode + transport-layer compression overhead, and the maintainability case stands either way.

totaam:
Please re-submit piecemeal.

PR #4866, opened 2026-04-12
Fix QUIC connection byte counters for session info

andrewachen:
## Summary

`XpraQuicConnection.write()` and `read()` bypass the base class `_write()` and `_read()` methods that increment `input_bytecount` and `output_bytecount`. This causes the session info bytes recv/sent graphs to show zero for QUIC connections.

6-line fix — increment the counters in `stream_write()` and `read()` directly.

Sponsored-By: Netflix


totaam:
Thanks, applied but closer to the actual IO call site in 779d5d0da91b8ac1182aeb063ef5b25b75afd5fc

PR #4875, opened 2026-04-21
Fix nvenc cleanup out-of-context leak and nvjpeg fallback guard

andrewachen:
## Summary

Two related fixes in the nvenc / fallback cleanup path, reachable under content-type-change storms on HEVC 4:4:4 nvenc (e.g. Microsoft Edge tab restore with multiple video windows):

- **nvenc `do_clean` leaks pycuda allocations → SEGV.** `cdc.lock.acquire(timeout=5)` with an `_force_context_release()` fallback left `self.inputBuffer` / `self.cudaInputBuffer` / `self.cudaOutputBuffer` live on the Encoder when the acquire timed out. Those pycuda allocations were then GC'd from the daemon cleanup thread without a CUDA context pushed, emitting `pagelocked_host_allocation in out-of-thread context could not be cleaned up` + leaking pinned memory until pycuda SEGV'd. The bug predates PR #4848 (pre-#4848, `do_clean` silently skipped cleanup on lock contention; #4848 fixed counter accounting but not the buffer leak).

  Fix: `do_clean` now uses blocking `cdc.lock.acquire()` and drops `_force_context_release` entirely. `threaded_clean` no longer takes `device_lock` — a new per-encoder `init_complete` `threading.Event` serializes cleanup against init of the *same* encoder without cross-blocking init of other encoders on a stuck compress_image. `self.kernel = None` is moved into `cuda_clean` so the Function/Module release happens under the pushed context (fixes a related "module in out-of-thread context" warning).

- **nvjpeg was never actually excluded from the video fallback list.** The guard added in b39162652f ("don't use nvjpeg as video fallback encoding") checks `encoder_name != "nvjpeg"`, but nvjpeg is registered via `add("enc_nvjpeg")` in `WindowSource.do_init_encoders` — so `encoder_name == "enc_nvjpeg"` at the check site and the guard has never matched. nvjpeg has silently been first in the fallback list for ~3 years, contrary to the commit's stated intent. One-liner: `"nvjpeg"` → `"enc_nvjpeg"`.

A synthetic reproducer is included at `tests/scripts/nvenc_cleanup_repro.py` — builds a pool of initialized Encoders, forces cleanup under cdc-lock contention, counts pycuda out-of-context warnings via subprocess capture. Before the fix: REPRO CONFIRMED. After: REPRO NEGATIVE.

Sponsored-By: Netflix

### DIFF
--- a/tests/scripts/nvenc_cleanup_repro.py
+++ b/tests/scripts/nvenc_cleanup_repro.py
@@ -1,0 +1,294 @@
+#!/usr/bin/env python3
+# This file is part of Xpra.
+# Copyright (C) 2026 Netflix, Inc.
+# Xpra is released under the terms of the GNU GPL v2, or, at your option, any
+# later version. See the file COPYING for details.
+# ABOUTME: Synthetic reproducer for nvenc `_force_context_release` out-of-context
+# ABOUTME: cleanup bug — builds a pool of fully-inited encoders, then tears them
+# ABOUTME: down while cdc.lock is held externally so each cleanup times out.
+
+"""
+Reproducer for the nvenc cleanup bug observed 2026-04-19:
+
+    pagelocked_host_allocation in out-of-thread context could not be cleaned up
+    device_allocation in out-of-thread context could not be cleaned up
+
+Root-cause hypothesis (see audit-findings.md):
+    When `do_clean()`'s `cdc.lock.acquire(timeout=5)` times out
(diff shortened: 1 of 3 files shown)


totaam:
I've made a small change that just reduces the diff and merged it: 1810168009a0932e6725a2e99c42952f4ac86713 + ad1e996e819f0d37d9baabd0d9f8457a5d3a7a16 + c63a286f20f570920bcc9250c50895ffb1343eac

Thanks!

PR #4881, opened 2026-05-06
Load all content-type rules across files and dirs

andrewachen:
## Summary

Two related issues in `xpra/server/window/content_guesser.py` cause user
and fallback rules to be silently dropped when loading
`content-type` / `content-categories` / `content-parent` defs.

### 1. Tilde paths from `do_get_user_conf_dirs` never load

`_load_dict_dir` calls `os.path.exists()` to skip missing directories.
On POSIX, `do_get_user_conf_dirs()` returns paths with literal `~`
(e.g. `"~/.config/xpra"`), and `os.path.exists("~/...")` returns
False without expansion — silent failure. **Any rules under
`~/.config/xpra/...` or `~/.xpra/...` are never loaded.**

The literal `~` in `do_get_user_conf_dirs` is intentional so multi-user
callers like `get_xpra_defaults_dirs(username, uid, gid)` in
`xpra/scripts/config.py` can pass these paths through
`osexpand(actual_username=..., uid=..., gid=...)` and resolve them
against the **target** user's home (proxy/start-as flows running as
root). Expanding `~` inside `do_get_user_conf_dirs` would bind the path
to the current process's HOME and break those flows. Expand at the
point of use in the loader instead, where the server is running as the
user the rules belong to.

### 2. Shallow `dict.update()` discards rules across files and dirs

`parse_content_types` returns a 2-level `{prop_name: {regex: tuple}}`
dict, but `_load_dict_dir` and `_load_dict_dirs` combine results with
`dict.update()`, which replaces the entire inner dict for any
`prop_name` defined in both source and target.

Symptoms:

- `90_fallback.conf`'s `role:browser=browser` replaces all role rules
  from `10_role.conf` (`gimp-dock`, `vlc-playlist`, etc).
- Once issue 1 is fixed and a user's
  `~/.config/xpra/content-type/50_extras.conf` actually loads with
  e.g. `class-instance:kitty=text`, the user's `class-instance` map
  replaces `/etc/xpra/content-type/50_class.conf`'s `class-instance`
  map wholesale — `xterm`, `jetbrains.*`, `gnome-terminal`, etc.
  lose their `text` classification and fall through to video encoding,
  which hurts text rendering quality and latency.

Fix:

- A small `_merge_defs` helper recurses one level when both sides are
  dicts. Behavior is unchanged for parsers that return flat dicts
  (`content-categories`, `content-parent`).
- `_load_dict_dirs` builds system and user tiers separately, then
  combines them so that:
  1. System's `prop_name` iteration order is preserved
     (`role`, `title`, `class-instance`, ...). `guess_content_type_from_defs`
     iterates by property and stops at the first match, so a user
     adding e.g. `class-instance:kitty=text` should not promote the
     class-instance property ahead of system's title rules — that
     would change classifications for unrelated windows like a
     Firefox Gmail tab matched by `title:- Gmail -=text`.
  2. Within each property, user regexes iterate before system regexes
     so a user override beats a broader system regex like
     `class-instance:.*terminal.*=text`.

Tests cover both regressions, the tilde expansion, the user-override
scenario (full `guess_content_type_from_defs` end-to-end), the
prop_name-order preservation, multi-system-dir precedence, and the
flat-parser regression.

Sponsored-By: Netflix

### DIFF
--- a/tests/unittests/unit/server/window/content_guesser_test.py
+++ b/tests/unittests/unit/server/window/content_guesser_test.py
@@ -1,0 +1,216 @@
+#!/usr/bin/env python3
+# This file is part of Xpra.
+# Copyright (C) 2026 Netflix, Inc.
+# Xpra is released under the terms of the GNU GPL v2, or, at your option, any
+# later version. See the file COPYING for details.
+
+import os
+import tempfile
+import unittest
+from unittest.mock import patch
+
+from xpra.os_util import POSIX
+from xpra.server.window import content_guesser
+from xpra.server.window.content_guesser import (
+    _load_dict_dirs,
+    parse_content_types,
+    parse_content_categories_file,
+    guess_content_type_from_defs,
+    _merge_defs,
+)
+
+
+class TestContentGuesser(unittest.TestCase):
+
+    def test_merge_defs_nested_preserves_both_sides(self):
+        # nested {prop: {key: value}} dicts must merge inner dicts,
+        # not replace them — otherwise rules from one file/dir wipe another.
+        target = {"class-instance": {"r1": "a", "r2": "b"}}
+        source = {"class-instance": {"r3": "c"}}
+        _merge_defs(target, source)
+        self.assertEqual(target, {"class-instance": {"r1": "a", "r2": "b", "r3": "c"}})
+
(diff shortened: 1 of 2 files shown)


totaam:
Thanks. Nice catch!

Issue #4877, opened 2026-05-05
wayland hardware-accelerated clients


totaam:
Improve upon #387
Implement [linux-dmabuf-v1](https://wayland.app/protocols/linux-dmabuf-v1) useful for many different types of applications: GPU-rendered GTK4, Firefox, Chromium, MPV, games via gamescope, anything using EGL/Vulkan with GBM.

Once we get GPU buffers directly from the client applications, we can think about the wiring needed for keeping the pixel buffers on the GPU as long as possible.

totaam:
With f012a4a57ccc2322c3e79b6d8322a5328978a139, we can even pass the fd directly to a local client and have it render directly from the `dmabuf`s that the client application used for rendering.

totaam:
Working and backported to v6.5.x branch since this is a self-contained and simple change.

Tested by running `weston-simple-egl`.
With f012a4a57ccc2322c3e79b6d8322a5328978a139, we can see the CPU download with `-d dmabuf`.

PR #4886, opened 2026-05-14
Intel oneVPL HEVC 4:4:4 hardware decoder for Windows

andrewachen:
## Summary

Adds HEVC 4:4:4 (RExt profile) hardware decode on Intel GPUs using
oneVPL, a generic decoder pool, and a `MFXVideoDECODE_Reset` fast path
to amortize encoder-cycling overhead. Complements the MediaFoundation
4:2:0 decoder landed in #4860 — VPL is the missing 4:4:4 path for
Intel hardware.

## Background

xpra's HEVC encoder produces 4:4:4 RExt streams for sharp content
(screen UI, browser text). The Windows client previously had no
hardware decoder for 4:4:4 — MF handles 4:2:0 only — so any 4:4:4
stream fell back to software decode, costing CPU and frame latency.

oneVPL exposes HEVC RExt 4:4:4 hardware decode on Intel Xe-LPG
(Meteor Lake) and newer. This change plumbs it into xpra's decoder
framework.

Concrete user-visible win: with 4:4:4 hardware decode in place, the
server can switch from webp/rgb24 to h265 4:4:4 during high-motion
periods on text-heavy content (browsers, editors) without
compromising text quality on the client side. Prefigured in
[#4860 (comment)](https://github.com/Xpra-org/xpra/pull/4860#issuecomment-4256548355).

## Why a decoder pool?

`MFXVideoDECODE_Init` is ~25 ms at 1920×1033 4:4:4 on Meteor Lake.
xpra cycles decoders frequently — damage rectangles vary in size and
each new size triggers a fresh decoder lifecycle. Without pooling,
every encoder cycle paid the full destroy+create cost (~90 ms first
frame). With pool + Reset fast path, the same cycle drops to ~5-10 ms
first frame. The pool also lets us pre-warm an idle decoder at a
likely envelope (e.g. monitor resolution) so the first frame after a
stream switch isn't gated on a cold Init.

## Why isn't MF using the pool?

The pool is decoder-agnostic by design — `DecoderPool` in
`xpra/codecs/decoder_pool.py` and any decoder can adopt it. MF
deliberately doesn't, today:

1. **ROI is much smaller.** MF's Init is ~10 ms vs VPL's ~25 ms;
   pooling MF would save ~8 ms per cycle, not a perceptual problem.
2. **No drop-in Reset equivalent.** MF's analog is
   `IMFTransform::ProcessMessage(MFT_MESSAGE_COMMAND_FLUSH)` +
   re-`SetInputType`/`SetOutputType` — different API, needs separate
   C-layer work.
3. **D3D11 lifetime.** MF's decoder is bound to a D3D11
   device-context; pooled lifetimes would need additional refcount
   care. VPL outputs to system memory, simpler.

If MF cycling becomes a hotspot later, the pool is ready — only the
codec-specific reset wrapper is missing.

### 1. `Intel oneVPL HEVC 4:4:4 hardware decoder for Windows`
- New `xpra/codecs/vpl/` module: Cython wrapper (`decoder.pyx`) over
  a C layer (`vpl_decode.c` / `vpl_decode.h`).
- Outputs AYUV (8-bit packed 4:4:4) directly to GL — no CSC, packed
  surfaces are uploaded straight to an AYUV→RGB shader in
  `xpra/opengl/shaders.py`.
- Session bound with three filters (hardware impl + HEVC codec +
  RExt profile), one `mfxConfig` per property per oneVPL's canonical
  pattern. Both the startup probe and real-session creation pin
  RExt so SCC-only systems don't false-select.
- oneVPL session is unconditionally hardware
  (`MFX_IMPL_TYPE_HARDWARE` pinned at session creation), so scoring
  is set once at registration with hardware-tier values — no
  equivalent of MF's runtime `HARDWARE_DECODERS` flag-flip needed
  (and avoids the registration-time-vs-runtime ordering issue noted
  on #4860).
- 8-bit only for now. Y410 (10-bit) deliberately not advertised in
  `output_colorspaces` — see deferred work below.
- Registered alongside the existing MF decoder in
  `xpra/codecs/loader.py`, `xpra/codecs/video.py`, `xpra/log.py`.
- `_OPAQUE_SHADER_FORMATS = {"AYUV"}` in `GLWindowBackingBase`
  exempts AYUV from the base-class alpha strip in
  `get_rgb_formats()` for opaque GL windows that advertise an RGBA
  visual (Chromium / Firefox set this even when they render
  opaquely).
- `libvpl` added to `packaging/MSWindows/SETUP.sh`. Build gated on
  `has_header_file("vpl/mfxvideo.h")` and `WIN32`.

### 2. `codecs: generic DecoderPool with TDD`
- `xpra/codecs/decoder_pool.py`: decoder-agnostic pool keyed on
  arbitrary tuple. `min_kept_idle` + prefer-largest reaper protect
  pre-warmed slots from being recycled away.
- 53 unit tests in `tests/unittests/unit/codecs/test_decoder_pool.py`.

### 3. `vpl: add vpl_decoder_reset C API + Cython extern`
- C entry point for reconfiguring an existing decoder session.
  Deferred-Reset state machine (`reset_pending` flag) — the Reset
  doesn't fire until `DecodeHeader` has run on the new bitstream so
  MFX has accurate params. (An earlier acquire-time-Reset design
  failed on real hardware with `MFX_ERR_INCOMPATIBLE_VIDEO_PARAM`
  from stale SPS fields leaking between streams.)
- Cython extern declaration.

### 4. `vpl: thin pool wrapper + env vars`
- `xpra/codecs/vpl/pool.py`: VPL-specific wrapper around
  `DecoderPool`.
- `XPRA_VPL_PREWARM_FULLSCREEN=WxH` for explicit prewarm; GDK
  monitor auto-detect when unset.
- 10 unit tests in `tests/unittests/unit/codecs/test_vpl_pool.py`.

### 5. `vpl: wire Decoder through pool`
- `Decoder.__init__` acquires from pool; `Decoder.clean` releases
  to the pool instead of destroying. Pool release uses
  `vpl_decoder_release_surface` (unmap only, `initialized=1`) — not
  `vpl_decoder_idle` (Close + unmap) — so reused slots skip
  lazy_init on the fast path.

### 6. `vpl: MFXVideoDECODE_Reset fast path for pool hits`
- Skip Close+Init when stream dims change but the session stays
  compatible. `MFXVideoDECODE_Reset` is ~150 µs vs ~25 ms for the
  Close+Init dance.
- Falls back to Close+Init transparently on
  `MFX_ERR_REALLOC_SURFACE` or any other Reset error.
- Close gates check `initialized || reset_pending` so a half-reset
  state still tears down the MFX session correctly.
- Escape hatch: `XPRA_VPL_RESET_FAST=0`.
- Measured (Meteor Lake, 1920×1033 4:4:4): pool cycle drops from
  ~55 ms → ~30 ms on fit cases; first-frame after pool hit drops
  to ~5-10 ms vs ~90 ms with no pool.

## Known gaps (deferred)

- **AYUV not advertised for truly-opaque windows.**
  `WindowBackingBase._get_full_csc_modes()` strips any format
  containing "A" when `_alpha_enabled=False`. Chromium / Firefox
  advertise RGBA visuals so `_alpha_enabled=True` and AYUV
  survives. Genuinely opaque windows (terminals, editors) fall
  back to YUV420P — fine, 4:4:4 only matters for sharp UI / text.

- **10-bit (Y410) not yet supported.** Pool key doesn't partition
  on `bit_depth`, and the reset path hardcodes 8-bit. Plumb
  `bit_depth` through the Cython wrapper and pool key before
  re-adding `Y410` to `output_colorspaces` / `RGB_MODES`. Render-
  side plumbing (shaders, format tables) is already in place so
  re-enabling is a one-line addition when the registration is
  fixed.

- **`MFX_WRN_VIDEO_PARAM_CHANGED` only refreshes width/height.**
  Cached `format`/`bit_depth` would go stale on a bit-depth flip
  (gated by the 10-bit gap above), and the new dims come from
  `GetVideoParam` which oneVPL may report as padded for non-16-
  aligned mid-stream resizes. Latent — xpra always sends a fresh
  stream rather than mid-stream SPS changes.

## Verification

- 63 new unit tests (decoder pool + VPL pool wrapper) pass.
- Live-validated for ~3 weeks against an nvenc-h265 server on a
  Meteor Lake daily-driver setup: Chromium 4:4:4 streams decode in
  hardware, encoder cycling (tab restore, fast scroll, window
  resize) hits the Reset fast path.
- 4 rounds of \`codex review\`; all findings either fixed or
  triaged with rationale above.

Sponsored-By: Netflix


xpra/codecs/vpl/decoder.pyx
```diff
@@ -0,0 +1,496 @@
...
+
+
+def get_type() -> str:
+    return "vpl"
+
+
+def get_info() -> Dict[str, Any]:
+    return {
+        "version": get_version(),
+        "type": "oneVPL",
```
totaam: Please return `type="vpl"`, same as `get_type`.

xpra/codecs/vpl/decoder.pyx
```diff
@@ -0,0 +1,496 @@
...
+    }
+
+
+def get_encodings() -> Sequence[str]:
+    return ("h265", )
```
totaam, line 203: Why only expose `h265` since onevpl supports av1, vp9, etc?
Is it that much different that adding those will cause bigger changes?

xpra/codecs/vpl/decoder.pyx
```diff
@@ -0,0 +1,496 @@
...
+
+def get_min_size(encoding) -> Tuple[int, int]:
+    return 64, 64
+
+
+MAX_WIDTH, MAX_HEIGHT = 16384, 16384
+
+
+def get_specs() -> Sequence[VideoSpec]:
+    # Only 4:4:4 — the MF decoder handles 4:2:0 for all GPU vendors.
```
totaam, line 214: onevpl should also work on Linux, so it makes sense to expose 420

xpra/codecs/vpl/decoder.pyx
```diff
@@ -0,0 +1,496 @@
...
+    # oneVPL could also do 4:2:0 (Main/Main10 → NV12), which would serve as a
+    # fallback if MF fails to load. Not added yet because it would need NV12
```
totaam, line 216: Why not expose NV12 frames directly?
The opengl backend supports NV12, and the cairo backend can use libyuv to convert them.

xpra/codecs/vpl/decoder.pyx
```diff
@@ -0,0 +1,496 @@
...
+            has_lossless_mode=False,
+            codec_class=Decoder,
+            codec_type=get_type(),
+            quality=100, speed=100,
+            size_efficiency=60,
+            setup_cost=0,
+            min_w=64, min_h=64,
+            width_mask=0xFFFE, height_mask=0xFFFE,
+            max_w=MAX_WIDTH, max_h=MAX_HEIGHT,
+            cpu_cost=10, gpu_cost=50,
```
totaam: I see below that there is a `vpl_decoder_is_hardware`.
It would make sense to keep track of the formats that are GPU accelerated during the self-tests so that this `gpu_cost` flag reflects the true hardware capabilities.

xpra/codecs/vpl/decoder.pyx
```diff
@@ -0,0 +1,496 @@
...
+        cdef int w = frame.width
+        cdef int h = frame.height
+        cdef int bpp = 4  # both AYUV and Y410 are 32 bits per pixel
+        cdef int row_bytes = w * bpp
+
+        copy_start = monotonic()
+        if frame.stride == row_bytes:
+            pixels = frame.data[:row_bytes * h]
+        else:
+            # stride != width*bpp due to padding; must compact rows
```
totaam: Why compact rows here?
The ImageWrapper should handle any rowstride correctly.

totaam (commented):
Please see the review comments.

Also, and more generally speaking:
* why we need `vpl_decode.h` and `vpl_decode.c` at all - couldn't these be implemented in Cython, either in `decoder.pyx` or in a pxd?
* looks like the opengl shaders could be merged as-is without fuss, but to add a decoder that produces these formats we need a software fallback path for `AYUV` and `Y410` - libyuv can handle `AYUV`, not sure what to do with `Y410` (I think that we can ignore it as 10 bits per pixel should be pretty rare at the moment - not a default config with the x11 backend)
* the whole pool thing is too complicated for what it's trying to solve (~90 ms of warming up):
  * it's a lot of code
  * it makes the decoder code aware of the pool somewhat - and makes it more complicated
  * I believe that we can find easier solutions: the server already sends jpeg frames for video encoders that use b-frames:
    https://github.com/Xpra-org/xpra/blob/1e11a7022f1417bbc4b8dd50060a04f982a068d5/xpra/server/window/video_compress.py#L2558
     something similar could be re-used for video decoders that expose some kind of "slow-start" flag: the client can paint the jpeg frame(s) and ignore the decoded hevc frame when they do come (and we only need a disposable thread for warming up the decoder)

andrewachen:
Thanks for the substantive review. Responses to the three architectural points:

**vpl_decode.h/c vs pure Cython**: oneVPL's API surface includes `mfxFrameSurface1::FrameInterface` (a vtable of function pointers per surface) and `MFXVideoDECODE_DecodeFrameAsync` which has to be called `nogil` for the sync wait to not stall the GIL. Both are workable in Cython, but the resulting code ends up with `cdef extern from "vpl/mfxstructures.h"` declarations for the whole surface vtable, manual struct field access, and explicit `with nogil:` blocks around every decode/sync/map. The C wrapper isolates the dispatcher/session lifecycle and exposes a flat 6-function surface (`startup`, `create`, `decode`, `release_surface`, `destroy`, `shutdown`). Happy to attempt the .pxd rewrite if you prefer — but I'd want to know whether you'd accept a `MFXVideoDECODE_DecodeFrameAsync` call with a `with nogil:` block holding a raw `mfxBitstream*`, or whether you'd want that path also moved.

**Software fallback for AYUV**: correcting an earlier version of this paragraph that was wrong about cairo. Cairo windows do already handle h265 video today via the libyuv CSC bridge — `xpra/codecs/libyuv/converter.pyx:254` wires up `NV12 → RGB/BGRX/RGBX/YUV420P`, so the per-window `full_csc_modes` negotiation picks h265-NV12 paths for cairo windows when the MF decoder is in play. The same would work for AYUV if libyuv's wrapper exposed `AYUVToARGB` (it exists in libyuv upstream, just not in `xpra/codecs/libyuv/converter.pyx`) and `COLORSPACES` gained `"AYUV": ("BGRX", "RGBX")`. ~30 LOC of wrapping. That's the missing piece for cairo h265-4:4:4 via VPL.

I'd prefer for someone else to drive this — it's not needed for my workflow and it's hard to justify the effort to do it.

**Pool complexity**: fair critique. The pool's actual value over a single-decoder cache turns out to be small — `paint_with_video_decoder` already caches `_video_decoder` per window, and the dim-change pattern within a window is what `MFXVideoDECODE_Reset` solves directly. The multi-slot LRU + prefer-largest reaper + `min_kept_idle` machinery was over-built for the actual lifecycle (one decoder per window, occasional bust). I'll rewrite this as: keep `vpl_decoder_reset` C API + the deferred-DecodeHeader state machine (that's the actual perf win — Close+Init is ~25ms, Reset is ~150µs), drop the generic `DecoderPool` entirely, and let `Decoder.init_context` either reset its existing context or create a new one. ~30-50 LOC instead of ~600 LOC + 53 tests.

Your jpeg-warmup idea is a great complement for the cold-start case, and notably it's not VPL-specific — MF init is ~10ms, nvdec is similar, any future AV1 HW decoder will have comparable cost. A generic "slow-start" flag in the decoder spec that triggers server-side jpeg seed frames would benefit all of them, including the MF path that's already landed. I'd propose landing the simplified VPL decoder here and following up with the slow-start framework as a separate PR — it's a bigger change touching both server (`compress.py`) and client (`backing.py`) and deserves its own review cycle.

xpra/codecs/vpl/decoder.pyx
andrewachen: Done — `type="vpl"` to match `get_type()`.

xpra/codecs/vpl/decoder.pyx
andrewachen: On Linux this is the only HW path for Intel — but I'm not personally targeting that workflow. This PR is scratching the HEVC 4:4:4 itch for Intel Windows clients (where MF already handles h264/h265 4:2:0).

I'd prefer for someone else to drive av1/vp9/h264 expansion — it's not needed for my workflow and it's hard to justify the effort to do it.

That said, the structure is friendly to adding them later: `probe_hevc_profile` generalizes to `probe_codec_profile(codec, profile)` cleanly, and the C decode loop is codec-agnostic (works on `dec->session` regardless of `CodecId`). Adding av1/vp9/h264 doesn't require rewriting the session lifecycle — just probe entries and `get_specs` lines.

xpra/codecs/vpl/decoder.pyx
andrewachen: Agreed that Linux Intel users would benefit since there's no MF fallback there. Same framing as L245 — not in my use case (MF covers 4:2:0 on Windows which is what I run), and I'd prefer for someone else to drive this — it's not needed for my workflow and it's hard to justify the effort to do it.

The work is bounded: C-layer `extract_frame()` is currently packed-32bpp-only; NV12 needs 2-plane stride/offset math + a surface-format dispatch (~50-100 LOC). libyuv already wraps `NV12 → BGRX/RGB/RGBX` (`xpra/codecs/libyuv/converter.pyx:254`), so both opengl and cairo windows would pick up h265-NV12 via the existing full_csc_modes graph once C extraction lands.

(Edit: corrected the cairo claim from my first version — cairo does handle NV12 today via libyuv CSC.)

xpra/codecs/vpl/decoder.pyx
andrewachen: You're right — cairo does handle NV12 via libyuv (`xpra/codecs/libyuv/converter.pyx:254` wires up `NV12 → BGRX/RGB/RGBX`), and the per-window `full_csc_modes` negotiation already picks that path for cairo when MF is in play. Same framing as L245/L256: not in my use case, and I'd prefer for someone else to drive this — it's not needed for my workflow and it's hard to justify the effort to do it.

The work is bounded: C-layer `extract_frame()` is currently packed-32bpp-only; NV12 needs 2-plane stride/offset math + a surface-format dispatch (~50-100 LOC). The CSC paths to both opengl and cairo are already in place via libyuv once C extraction lands.

(Edit: corrected the cairo claim from my first version — I'd incorrectly said cairo couldn't handle NV12.)

xpra/codecs/vpl/decoder.pyx
andrewachen: Reading this: you want `gpu_cost` to reflect actual hardware status at registration time, not be a hardcoded value — the same pattern bug that bit `HARDWARE_DECODERS` in MF (where `init_context` populates it after `get_specs` already ran). Today VPL pins `MFX_IMPL_TYPE_HARDWARE` at session creation, so the answer is trivially "always HW" — but if codecs are added later per L245, some might be HW on one GPU generation and SW on another (e.g., older iGPUs lacking AV1 decode). The right fix is to restructure the startup probe to record per-(codec, profile) hardware status and have `get_specs` read from that table. I'll leave a comment marking this for the codec-expansion follow-up.

xpra/codecs/vpl/decoder.pyx
andrewachen: Done — removed the row-compaction path; passing `frame.stride` directly to `ImageWrapper`. GL upload uses `GL_UNPACK_ROW_LENGTH`, so non-tight strides work natively. Saves a per-frame memcpy of about 8MB at 1920×1033.

xpra/client/subsystem/encoding.py
```diff
@@ -265,6 +265,11 @@ def get_encodings_caps(self) -> dict[str, Any]:
         # we can send different values as part of the map event
         # these are the RGB modes we want (the ones we are expected to be able to paint with):
         rgb_formats = ["RGB", "RGBX", "RGBA"]
+        # opengl_enabled lives on the Display mixin; guard for test
+        # harnesses that instantiate Encodings without Display.
+        if getattr(self, "opengl_enabled", False):
+            # GL shaders can render packed YUV formats directly (no CSC needed)
+            rgb_formats += ["AYUV"]
```
totaam: I'm not sure that it is correct to expose this here.
Especially if we don't have a non-opengl fallback.

xpra/codecs/vpl/decoder.pyx
totaam: I can deal with Linux support once this is merged.

xpra/codecs/vpl/decoder.pyx
totaam: It's not clear to me what the "plane stride/offset math + a surface-format dispatch" is.

xpra/codecs/vpl/decoder.pyx
totaam: That's exactly what I meant.

xpra/opengl/backing.py
```diff
@@ -242,12 +248,31 @@ class GLWindowBackingBase(WindowBackingBase):
...
         "YUV420P16",
         "GBRP", "BGRA", "BGRX", "RGBA", "RGBX",
         "RGB", "BGR",
     )
     HAS_ALPHA: bool = GL_ALPHA_SUPPORTED
+    # Formats rendered via GL shaders that are opaque despite having "A" in the name.
+    # The base class alpha filter strips any format containing "A", which would
+    # incorrectly remove these from non-alpha windows.
+    _OPAQUE_SHADER_FORMATS: frozenset = frozenset(("AYUV", ))
+    _is_intel: bool | None = None  # cached across all windows
```
totaam: This does not belong here.

xpra/opengl/backing.py
```diff
@@ -242,12 +248,31 @@ class GLWindowBackingBase(WindowBackingBase):
...
+
+    def get_rgb_formats(self) -> Sequence[str]:
+        if self._alpha_enabled:
+            # _alpha_enabled means the window advertises an RGBA visual,
+            # not that it actually composites with meaningful transparency
+            # — Chromium/Firefox set this but render opaquely. Keep AYUV
+            # available here so those windows can use VPL HEVC 4:4:4.
+            # Genuinely-transparent windows (rare: tooltips, some docks)
+            # would lose transparency if the server happened to pick
+            # VPL→AYUV, but that's an acceptable edge vs. losing 4:4:4
```
totaam: Why can't we paint with alpha since AYUV carries it?
(or does it not? invalid?)

totaam: Looks to me like the shader drops the alpha.
We could just have 2 shaders, or a parameter.
And keep the alpha when we know it can / should be used.

totaam (commented):
To get us closer to the endgame on this large PR, it may make sense to submit the opengl support separately.

As for vpl: apart from the review comments, the pool code needs removing.

andrewachen:
> the pool code needs removing

Agreed. I have this done in my 6.4.x branch; I need to give it a test before I push the master version for this PR.

xpra/codecs/vpl/decoder.pyx
andrewachen: Fair pushback — Claude's phrasing was vague. Concretely it's ~50 LOC across the C wrapper (second plane pointer + pitch in the extract struct, format dispatch in `extract_frame`) and Cython (multi-plane `ImageWrapper` instead of single-buffer). No manual stride/offset arithmetic; the surface API exposes both pitches directly. Still prefer for someone else to drive NV12, but the boundary is narrower than the earlier estimate suggested.

xpra/codecs/vpl/decoder.pyx
andrewachen: Got it — same shape as the `HARDWARE_DECODERS` ordering bug we have for MF. The fix is part of the master-port refactor I committed to in the conversation thread (pool→cache simplification): `do_test_setup()` probes whether a probe context comes back as hardware, sets a module-level boolean, then `get_specs()` returns the appropriate `gpu_cost`. Will reference the commit when that lands.

xpra/client/subsystem/encoding.py
andrewachen: How about moving the `_OPAQUE_SHADER_FORMATS` override (`xpra/opengl/backing.py`) up to the base backing class as a generic `OPAQUE_FORMATS_WITH_A` class attribute, consulted by all three name-based filter sites in the base (`get_rgb_formats()` at L215, `_get_full_csc_modes()` at L493, `paint_with_pixel_format()` at L623)? Once that's in place, opengl windows naturally include AYUV in their `csc_modes.h265` and cairo windows don't — and this conditional in `encoding.py` becomes unnecessary, so it can be deleted. The cairo gap is real but isolated: cairo windows would negotiate to NV12/4:2:0 just like they do today with MF.

xpra/codecs/vpl/decoder.pyx
andrewachen: 👍 thanks. Currently the gate is just `vpl_decoder_ENABLED = DEFAULT and WIN32 and ...` in `setup.py:325`. Flipping the `WIN32` constraint to also allow Linux + libvpl present should be the only change.

xpra/opengl/backing.py
andrewachen: How about moving the override to the base backing class as a generic `OPAQUE_FORMATS_WITH_A: frozenset = frozenset()` class attribute (default empty), with the GL subclass declaring `OPAQUE_FORMATS_WITH_A = frozenset({"AYUV"})`? The three name-based filter sites in the base (`get_rgb_formats()` at L215, `_get_full_csc_modes()` at L493, `paint_with_pixel_format()` at L623) would all consult it.

Side note: the current GL-subclass-only override is silently incomplete — it only fixes `get_rgb_formats()` but leaves L493 stripping AYUV during csc negotiation. The base-class move would fix that latent bug too.

xpra/opengl/backing.py
andrewachen: AYUV from VPL on HEVC RExt 4:4:4 carries A=0xFF for every pixel — the bitstream is YUV-only and Intel's hardware fills the A byte with full opaque. So the alpha exists in the surface but is always 1.0.

Right — `xpra/opengl/shaders.py:186` currently hardcodes `vec4(r, g, b, 1.0)`. I'll change it to `vec4(r, g, b, t.a)` — single shader, no parameter, no branching. For HEVC RExt 4:4:4 `t.a` is always 1.0 so the rendered output is bit-identical, but the codebase becomes honest about what the surface format carries, and any future codec that produces real alpha in an AYUV surface gets it preserved automatically. The framebuffer's alpha handling (opaque window → RGB-only FBO → alpha dropped naturally; alpha window → alpha preserved) takes care of the per-window-mode decision without a shader-side branch.

xpra/client/subsystem/encoding.py
andrewachen: Quick correction on the proposed deletion. Tracing the per-window encoding-property flow more carefully, `xpra/wayland/subsystem/window.py::_process_window_map` doesnt call `_set_client_properties` — the clients per-window `encoding.full_csc_modes` (with AYUV via the new base-class `OPAQUE_FORMATS_WITH_A`) never reaches the server on Wayland sessions. The connection-level `caps["full_csc_modes"]` from hello is the only signal Wayland gets, so this conditional has to stay.

Revised: kept the `if self.opengl_enabled: rgb_formats += ["AYUV"]` line with a comment explaining the Wayland dependency, and dropped Y410 from the list (the VPL decoder no longer advertises it until 10-bit is plumbed). Commit e9da5727e4 on the v6.4.3 branch; will be in the master push.

xpra/opengl/backing.py
andrewachen: Implemented in [`149205a4c7`](https://github.com/Xpra-org/xpra/pull/4886/commits/149205a4c7) (master) — moves `OPAQUE_FORMATS_WITH_A` to `WindowBackingBase`, consulted at all three name-based filter sites.

xpra/opengl/backing.py
andrewachen: Implemented in [`e15803f70b`](https://github.com/Xpra-org/xpra/pull/4886/commits/e15803f70b) (master) — single shader change `vec4(r, g, b, 1.0)` -> `vec4(r, g, b, t.a)`. For HEVC RExt 4:4:4 `t.a` is always 1.0 so the rendered output is bit-identical, but the shader becomes honest about what AYUV carries and forward-compatible with any future codec that produces real alpha. The framebuffer alpha handling (FBO format, blend modes) is governed by window mode, so no separate shader or uniform parameter is needed.

xpra/codecs/vpl/decoder.pyx
andrewachen: Implemented in [`c346dd93ed`](https://github.com/Xpra-org/xpra/pull/4886/commits/c346dd93ed) (master) — `init_module()` now creates a tiny probe decoder and captures `vpl_decoder_is_hardware()` into a module-level `_HARDWARE_ACCEL` flag. `get_specs()` returns `cpu_cost=10 / gpu_cost=50` if hardware-accelerated, `cpu_cost=100 / gpu_cost=0` if software fallback. Same ordering pattern as the MF `HARDWARE_DECODERS` fix (probe first, register spec second).

xpra/client/subsystem/encoding.py
andrewachen: On master as [`a39a53ae3e`](https://github.com/Xpra-org/xpra/pull/4886/commits/a39a53ae3e) — keeps the `if getattr(self, "opengl_enabled", False):` guard you originally added (test-harness safety) and just drops Y410 + adds a comment explaining the Wayland dependency.

andrewachen:
Quick status on both points:

**Pool removal**: landed on master in [`529c26995c`](https://github.com/Xpra-org/xpra/pull/4886/commits/529c26995c) — the multi-slot LRU + reaper machinery is gone; `init_module()` parks at most one `VPLDecoder*` in a module-level cache, `init_context` pops it via `vpl_decoder_reset` (~150 µs) or creates a fresh one if absent.

**OpenGL separation**: the actual surface in this PR is now ~15 lines total — one fragment-shader function (`gen_AYUV_to_RGB` at `xpra/opengl/shaders.py:155`) and one `OPAQUE_FORMATS_WITH_A` class attribute promoted to the base `WindowBackingBase` ([`149205a4c7`](https://github.com/Xpra-org/xpra/pull/4886/commits/149205a4c7)). I’d push back gently on splitting these into a separate PR: there’s no AYUV producer on master before VPL, so a standalone "opengl AYUV" PR would land code that nothing exercises until VPL follows. Happy to drop the shader + filter attr from this PR and land them with VPL part 2 if you’d prefer, but the most likely "merged in isolation" outcome is just dead code until the decoder lands.

xpra/client/subsystem/encoding.py
totaam: Rather than complicating the client code just because the wayland server fails to account for csc modes, let's just create a ticket for the wayland server and fix that.

xpra/opengl/backing.py
totaam: I meant the line with `is_intel` - which isn't used anywhere.

As for `OPAQUE_FORMATS_WITH_A`, `AYUV` isn't necessarily opaque, it just happens to be opaque coming out of the onevpl decoder.

I think it would make more sense to:
* expose `XYUV` out of the decoder (`AYUV` packing without any alpha)
* support both `AYUV` and `XYUV` in the opengl backend (almost the same shader)

xpra/opengl/backing.py
totaam: If alpha is always `1.0` then you can ignore what I said above about `XYUV` and just support `AYUV` only.

xpra/opengl/backing.py
totaam: As per other comments, I don't understand why we need `OPAQUE_FORMATS_WITH_A` if the `AYUV` data has `A=1.0`

xpra/codecs/vpl/decoder.pyx
totaam: I'll take care of testing YUV420P on Linux.

totaam:
BTW, I have a new incentive to get this merged: I got wayland GPU buffers working in #4877, and one of the easiest ways of avoiding downloading pixel buffers to the CPU for compression is likely onevpl.

xpra/client/subsystem/encoding.py
andrewachen: Done — conditional removed in [`94e3e78111`](https://github.com/Xpra-org/xpra/pull/4886/commits/94e3e78111). Per-window negotiation via the new `ALPHA_FORMATS`-based filter ([`edaed0a0ab`](https://github.com/Xpra-org/xpra/pull/4886/commits/edaed0a0ab)) surfaces AYUV on the X11 server. Filed Wayland ticket [#4896](https://github.com/Xpra-org/xpra/issues/4896) for the per-window encoding-property gap.

xpra/opengl/backing.py
andrewachen: Agreed — replaced with explicit `ALPHA_FORMATS = frozenset({"BGRA","RGBA","ABGR","ARGB"})` in `xpra/codecs/constants.py` ([`44d537ab14`](https://github.com/Xpra-org/xpra/pull/4886/commits/44d537ab14)), consulted at all five strip-by-name sites (three in `WindowBackingBase` at [`edaed0a0ab`](https://github.com/Xpra-org/xpra/pull/4886/commits/edaed0a0ab), one each in `compress.py` and `video_compress.py` at [`df9232a783`](https://github.com/Xpra-org/xpra/pull/4886/commits/df9232a783)). AYUV naturally falls through because it isn't in the set. `OPAQUE_FORMATS_WITH_A` and the dead `_is_intel` attribute removed in [`edaed0a0ab`](https://github.com/Xpra-org/xpra/pull/4886/commits/edaed0a0ab) and [`c0ec0c1e2b`](https://github.com/Xpra-org/xpra/pull/4886/commits/c0ec0c1e2b) respectively.

totaam: That is not what I meant.
Please see: c3aa77e9eb3ad514a807d16e67eca9e8c2d2354e + 6476434eb07dfea4f46ae0bc973df5784d0c1f43

andrewachen:
Thanks for cleaning this up; apologies for not quite getting what you were looking for.

PR #4890, opened 2026-05-22
nvenc: prevent use-after-free SEGV in Encoder __dealloc__

andrewachen:
## Summary

Fix two latent bugs in nvenc Encoder cleanup that combine to cause server SIGSEGV under encoder reinit pressure (observed live on a 4:4:4 HEVC workload):

1. **`Encoder.__dealloc__` use-after-free.** When `threaded_init` is on (the default), `__dealloc__` called `self.clean()`, which spawns `threaded_clean` via `start_thread`. The Thread retains a bound method holding `self`, but Cython's `tp_dealloc` for a cdef class frees the struct after `__dealloc__` returns regardless of refcount resurrection. The spawned thread then dereferences freed memory → SEGV.

2. **`WindowVideoSource.cleanup()` ordering.** `super().cleanup()` ran first, which invokes `init_vars()` (Python method lookup resolves to the subclass override, nulling `self._video_encoder` / `_csc_encoder`). The follow-up `cleanup_codecs()` then saw `None` and skipped the explicit `clean()` call — putting the encoder onto the `__dealloc__` UAF path above.

### `server: clean codecs before init_vars in WindowVideoSource.cleanup`
Reorder `cleanup()` so `cancel_damage(MAX_SEQUENCE)` runs first. `cancel_damage()` in this class already calls `cleanup_codecs()`, which captures the encoder refs and schedules explicit `clean()` in the encode thread. After that, `super().cleanup()` can safely run `init_vars()`.

### `nvenc: prevent use-after-free SEGV in Encoder __dealloc__`
`__dealloc__` now branches on `self.threaded_init`:
- `True`: log a warning and mark `closed=1` (leak GPU resources rather than crash). Callers MUST call `clean()` before dropping the last reference.
- `False`: `clean()` runs `do_clean()` inline (no thread spawn, no UAF), safe.

Also fix the one-shot `_process_encode` path in `xpra/server/encoder/server.py` to call `encoder.clean()` explicitly — it used to rely on `__dealloc__`.

## Reproducer

`tests/scripts/nvenc_dealloc_uaf_repro.py` exercises the bug in isolation:
- Before this PR: 3/3 SEGV
- After this PR: 3/3 clean exits (with the warning firing as a safety net for any caller that forgets `clean()`)

## Known limitations

A residual narrow race remains in `WindowVideoSource.cleanup()`: if the encode thread is mid-`video_encode` when `cancel_damage` fires, and it reaches `check_pipeline()` before observing `_damage_cancelled`, it can assign a fresh encoder to `self._video_encoder` that `init_vars()` then nulls without `clean()`. With this PR the consequence is a leaked NVENC context per occurrence rather than a SEGV — strict improvement over the previous behavior, but not a complete fix.

Attempts to close the race revealed an ordering subtlety worth flagging:

- An `encode_ended` override can clean any encoder still attached when it runs, but only catches assignments that happen AFTER `init_vars()` (and the racing-in-flight encode's local ref drops before that override runs).
- Synchronously draining the encode thread (`call_in_encode_thread(set_event)` + `event.wait()`) in `cleanup()` would close the race for the per-window `remove_window` path, but breaks the client-disconnect path: `ClientConnectionMuxer.close()` iterates `reversed(CC_BASES)`, so `EncodingsConnection.cleanup()` runs first and queues the `None` end-of-queue sentinel before `WindowsConnection.cleanup()` reaches the per-window cleanups. A drain queued after the sentinel never fires; cleanup stalls for the full timeout. (The comment in `xpra/server/source/encoding.py:cleanup()` about "must come AFTER the window mixin" is misleading vs. the actual reversed-cleanup behavior in `factory.py`.)

Fully eliminating the race would require either reworking the encode-thread shutdown sequence so window-source cleanups can drain reliably, or making `_video_encoder` access synchronized between main and encode threads. Both are larger than this PR.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>
Sponsored-By: Netflix

### DIFF
--- a/xpra/server/encoder/server.py
+++ b/xpra/server/encoder/server.py
@@ -199,8 +199,22 @@ class EncoderServer(ServerBase):
                         log(f" supported pixel formats for {encoding!r}: %s", csv(input_cs_options))
                         raise ValueError(msg)
                 add_device_context(ss, options)
-                encoder.init_context(encoding, width, height, pixel_format, typedict(options))
-                bdata, client_options = encoder.compress_image(image, typedict(options))
+                try:
+                    encoder.init_context(encoding, width, height, pixel_format, typedict(options))
+                    bdata, client_options = encoder.compress_image(image, typedict(options))
+                finally:
+                    # Per-request one-shot encoder: clean explicitly so
+                    # GPU resources are released even if init_context or
+                    # compress_image raised. Relying on __dealloc__ is
+                    # unsafe — see nvenc Encoder.__dealloc__. Swallow
+                    # cleanup errors so we don't replace a successful
+                    # encode result (or mask the original exception).
+                    try:
+                        encoder.clean()
+                    except Exception:
+                        # NVENCException from raiseNVENC() inherits from
+                        # Exception (not RuntimeError), so catch broadly.
+                        log.error(f"Error cleaning one-shot encoder {encoder}", exc_info=True)
                 bpp = 24
                 stride = 0
                 coding = encoding
--- a/xpra/server/window/video_compress.py
+++ b/xpra/server/window/video_compress.py
@@ -27,6 +27,7 @@ from xpra.server.window.compress import (
     STRICT_MODE, LOSSLESS_WINDOW_TYPES,
     DOWNSCALE_THRESHOLD, DOWNSCALE, TEXT_QUALITY,
     LOG_ENCODERS,
+    MAX_SEQUENCE,
 )
 from xpra.net.common import Packet, BACKWARDS_COMPATIBLE
 from xpra.util.rectangle import rectangle, merge_all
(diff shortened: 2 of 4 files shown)


xpra/server/window/video_compress.py
```diff
@@ -370,8 +371,19 @@ def suspend(self) -> None:
...
+        # in the encode thread. It must run before super().cleanup()
+        # because the base class's cleanup() calls init_vars() —
+        # resolved (via Python method lookup) to our subclass override,
+        # which nulls _video_encoder and _csc_encoder. If init_vars
+        # runs first, video_context_clean() inside cleanup_codecs sees
+        # None refs and skips the explicit clean(); the encoder is then
+        # GC'd without closed=True, exposing the nvenc __dealloc__
+        # use-after-free path (defended by the companion commit, but
+        # better not to reach it at all).
+        self.cancel_damage(MAX_SEQUENCE)
```
totaam, line 385: That doesn't make sense to me.
The `super().cleanup()` already calls it:
https://github.com/Xpra-org/xpra/blob/91d5977642d66a5b4dc8a38e4e0eb8a246e06e6e/xpra/server/window/compress.py#L524-L525

xpra/server/encoder/server.py
```diff
@@ -199,8 +199,22 @@ def _process_encode(self, proto: SocketProtocol, packet: Packet) -> None:
...
+                    # GPU resources are released even if init_context or
+                    # compress_image raised. Relying on __dealloc__ is
+                    # unsafe — see nvenc Encoder.__dealloc__. Swallow
+                    # cleanup errors so we don't replace a successful
+                    # encode result (or mask the original exception).
+                    try:
+                        encoder.clean()
+                    except Exception:
+                        # NVENCException from raiseNVENC() inherits from
+                        # Exception (not RuntimeError), so catch broadly.
```
totaam, line 216: Fair comment, addressed in 39d9530

xpra/server/window/video_compress.py
andrewachen: You're right — `super().cleanup()` calls `self.cancel_damage(MAX_SEQUENCE)` first (compress.py:524), and `WindowVideoSource.cancel_damage()` already invokes `cleanup_codecs()` before `init_vars()` runs. The reorder was based on a misreading and is redundant. Dropping it.

andrewachen:
Two corrections to my closing note:

- 39d9530 narrows `NVENCException` to `RuntimeError`. It does **not** add the try/finally cleanup in `_process_encode`, so the one-shot encoder there still relies on `__dealloc__`.
- The `__dealloc__` UAF (3/3 SEGV in the reproducer when a threaded-init Encoder is dropped without explicit `clean()`) is therefore still live on master.

Do you want me to reopen with just the `__dealloc__` branch + reproducer (i.e. without the `cleanup_codecs()` reorder you correctly flagged), or leave it for now?

Issue #3376, opened 2021-12-04
quic transport


totaam:
[QUIC](https://en.wikipedia.org/wiki/QUIC) could easily replace the old UDP transport.
There are frameworks and APIs that wrap QUIC with easy to use callbacks, like https://github.com/aiortc/aioquic [documentation](https://aioquic.readthedocs.io/en/latest/)
(see also https://github.com/MagicStack/uvloop)
We would get HTTP/3 and encryption almost for free and a reliable event based framework for managing connection state.

[Examples](https://github.com/aiortc/aioquic/tree/main/examples) include an [http3 server](https://github.com/aiortc/aioquic/blob/main/examples/http3_server.py) and clients, etc..

The re-send / packet-drop callbacks could easily re-use the techniques that were used for UDP (#639) which was removed in 2022238ababa40382d2b8d6b019fb2d757ec9955.

---

Worth noting that _aioquic requires Python 3.7 or better, and the OpenSSL development headers_ so this won't be available on RHEL 8.

totaam:
`aioquic` is not available in Fedora and the number of dependencies required to run the http3 server example is a little bit ridiculous: `python3-wsproto`, `python3-httpbin`, `python3-asgiref`, `python3-starlette` and `python3-aiofiles`.
Not counting all the dependencies that each one of these packages brings...

totaam:
It gets worse since we also need to run the GTK main loop.
There are various ways of doing this:
* [asyncio-glib](https://github.com/jhenstridge/asyncio-glib/)
* [gbulb](https://github.com/beeware/gbulb)
And perhaps soon in pygobject directly: [Implement asyncio event loop based on glib](https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/189)

Next, creating `asyncio` sockets requires using a specific API: [asyncio.loop.create_datagram_endpoint](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_datagram_endpoint) / [loop.create_connection](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_connection), which means that socket sharing is going to be more difficult to implement.

This may help in bridging with the rest of the code:
* [trio](https://github.com/python-trio/trio) : _a friendly Python library for async concurrency and I/O_
* [greenback](https://github.com/oremanj/greenback) : _allows you to call back into async code from a syntactically synchronous function_

TijZwa:
[...]

PR #4891, opened 2026-05-22
asyncio: stop the loop at interpreter shutdown via atexit

andrewachen:
## Background

`ThreadedAsyncioLoop` in `xpra/net/aio/thread.py` runs uvloop's
`run_forever()` on a daemon thread with no shutdown coordination —
no `stop()`, no `atexit`, no caller-side teardown. The
`loop.close()` line after `run_forever()` is unreachable in normal
operation.

The class dates to `072a553e38` (2022-11-06, "#3376 split directory
listing, simplify calling async code"), which introduced the
original `xpra/net/quic/asyncio_thread.py`. It has since been
moved twice (`beab487c7c` to `xpra/net/asyncio/`, then
`7d662ac562` to `xpra/net/aio/` to avoid the stdlib shadow) but the
daemon-thread-no-shutdown design has been unchanged across all
revisions.

## Symptom

During Python interpreter teardown, `_PyGILState_Fini()` zeroes
`gilstate->autoInterpreterState`. If the daemon asyncio thread is
still parked in `uv_run` with a pending libuv callback, the callback
fires, uvloop's C wrapper calls `PyGILState_Ensure()`, and
`new_threadstate(NULL)` segfaults at `Python/pystate.c:1388`.

QUIC-only in practice because aioquic's
`loop.create_datagram_endpoint` is the only path that brings uvloop
into xpra. TCP and WebSocket use xpra's own threaded socket handlers
and never touch uvloop.

Observed twice in production within a week from ordinary
`systemctl restart` cycles with a QUIC client connected:

- SEGV stack: `new_threadstate(interp=0x0)` at `Python/pystate.c:1388`,
  called from `PyGILState_Ensure` (`pystate.c:2218`), called from
  uvloop's C UDP datagram callback inside `uv_run`.
- Same root cause caught later in finalization: uvloop delivered a
  datagram to `XpraWebSocketHandler.http_event_received`'s `log.info`
  call after stdio had been closed, raising
  `ValueError: I/O operation on closed file`.

## Fix

Add `ThreadedAsyncioLoop.stop()` that schedules `loop.stop()` via
`call_soon_threadsafe` and joins the daemon thread with a 1s
timeout. Register it via `atexit` at singleton creation. atexit
fires before `_PyGILState_Fini`, so on the successful-stop path the
daemon thread is gone before the finalization race window opens.

If `stop()` times out at 1s (pathological case where the asyncio
thread is blocked in a long callback), `stop()` logs a warning and
returns; the residual race remains possible in that window but
`daemon=True` keeps process exit from deadlocking. The successful
path is what closes the SEGV.

`stop()` unregisters its own atexit hook on success so re-stops are
idempotent and the bound-method reference does not retain the loop
instance forever.

`stop()` guards on `loop.is_closed()` rather than `loop.is_running()`
because a freshly-constructed loop has `is_running() == False` until
`run_forever()` actually enters its main loop; the `is_running`
guard would drop a stop request fired in that startup window.

Sponsored-By: Netflix

### DIFF
--- a/tests/unittests/unit/net/aio_thread_test.py
+++ b/tests/unittests/unit/net/aio_thread_test.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+# This file is part of Xpra.
+# Copyright (C) 2026 Netflix, Inc.
+# Xpra is released under the terms of the GNU GPL v2, or, at your option, any
+# later version. See the file COPYING for details.
+
+# ABOUTME: Tests for ThreadedAsyncioLoop.stop() — verifies the loop is stopped
+# ABOUTME: and the daemon thread is joined, which is what prevents the uvloop
+# ABOUTME: callback vs Py_FinalizeEx race that crashes in PyGILState_Ensure
+# ABOUTME: -> new_threadstate(NULL).
+
+import unittest
+
+from xpra.net.aio.thread import ThreadedAsyncioLoop
+
+
+class TestThreadedAsyncioLoopStop(unittest.TestCase):
+
+    def test_stop_terminates_loop_and_joins_thread(self):
+        tl = ThreadedAsyncioLoop()
+        thread = tl._thread
+        self.assertIsNotNone(thread)
+        self.assertTrue(thread.is_alive())
+        loop = tl.loop
+        self.assertIsNotNone(loop)
+        # We don't assert loop.is_running() here: wait_for_loop() only
+        # waits for self.loop to be assigned, not for run_forever() to
+        # start. Under scheduler pressure that assertion would flake.
+
+        tl.stop()
+
+        self.assertFalse(thread.is_alive(), "asyncio thread did not exit")
+        # run_forever() returns, then loop.close() runs at the end of
+        # the thread; by the time stop() returns from thread.join() the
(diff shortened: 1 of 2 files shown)


totaam:
I dislike the use of `atexit` in general, so I've used a different approach: client and servers always stop the asyncio loop if it had been started.

Also propagate the handshake errors via an optional `ExitCode` on the connection objects. ie: quic can set it to `SSL_CERTIFICATE_VERIFY_FAILURE` and fail faster.

I believe this should address your  shutdown issue, without using atexit.

andrewachen:
Thanks for the alternative — your explicit-cleanup approach (commits 73ec4f4235, 7b332c64d0, 6fbc9d71fa, a2993a1c5f) is the right call. Closing in favor of those.

I've ported the four #4891 commits to my v6.4.3-achen daily-driver branch, with the minimal adaptations needed for the older paths (`xpra.net.aio` → `xpra.net.asyncio`, the `_watch_handshake` function placed before the monolithic `quic_connect` rather than the master-split `_quic_connect`, and `stop_asyncio_loop()` keyed off `xpra.net.asyncio.thread` in `xpra/common.py`).

PR #4892, opened 2026-05-26
pointer/keyboard scroll-hints + py: external transform loader

hamidthri:
Two self-contained commits.

1. **pointer, keyboard: derive scroll-hint signals from input events.**
   Drag on the right-edge scrollbar zone of a window and a configurable
   set of keyboard keys (`Page_Up`/`Page_Down`, `Home`/`End`, `Up`/`Down`)
   call `record_scroll_event(wid)` so downstream consumers (chiefly image
   filters that want temporal stability across scrolls) learn about
   scrolls beyond wheel events. Drag-as-scroll is on by default;
   keyboard-as-scroll defaults OFF.

2. **codecs/pytorch/filter: support external `py:` transforms.**
   Adds a `py:<module>:<Class>(kw=val,...)` transform option that loads
   any importable callable as the image filter. When the callable sets
   `supports_numpy = True`, frames go through a CPU numpy fast path that
   skips the torchvision GPU pipeline. Same commit fixes a use-after-free
   in the threaded image-filter path (`convert_image()` now returns a new
   `ImageWrapper` rather than mutating the input wrapper that gets freed
   downstream), adds a rowstride-safe BGRX reshape, and an
   `XPRA_IMAGEFILTER_TRANSFORM` env fallback for the option default.

### DIFF
--- a/xpra/codecs/pytorch/filter.py
+++ b/xpra/codecs/pytorch/filter.py
@@ -3,6 +3,7 @@
 # Xpra is released under the terms of the GNU GPL v2, or, at your option, any
 # later version. See the file COPYING for details.
 
+import os
 from typing import Any, Sequence
 
 from xpra.util.objects import typedict
@@ -103,6 +104,29 @@ def torch_init() -> None:
             cuda_info["devices"] = devices
 
 
+def _load_py_transform(transform_str: str) -> tuple[Any, dict]:
+    """Load an external callable via 'py:<module>:<Class>(kwargs...)'."""
+    import importlib
+    rest = transform_str[len("py:"):]
+    if ":" not in rest:
+        raise ValueError(
+            f"Invalid py: transform {transform_str!r}. "
+            "Expected format: py:<module>:<callable>(kwargs...)"
+        )
+    module_path, callable_part = rest.split(":", 1)
+    callable_name, kwargs = parse_function_call(callable_part)
+    log.info("py: loader: importing %r.%r kwargs=%s", module_path, callable_name, kwargs)
+    try:
+        mod = importlib.import_module(module_path)
+    except ImportError as exc:
+        raise ImportError(
+            f"Cannot import {module_path!r} for py: transform. "
+            f"Install the package or set PYTHONPATH. Error: {exc}"
+        ) from exc
+    factory = getattr(mod, callable_name)
+    return factory(**kwargs), {}
+
+
 class Filter:
     __slots__ = ("closed", "width", "height", "device", "transform", "kwargs")
 
@@ -124,7 +148,14 @@ class Filter:
         self.width = src_width
         self.height = src_height
         self.device = "cuda"
-        transform_str = options.strget("transform", "functional.invert")
+        transform_str = (
+            options.strget("transform")
+            or os.environ.get("XPRA_IMAGEFILTER_TRANSFORM", "functional.invert")
+        )
+        if transform_str.startswith("py:"):
+            self.transform, self.kwargs = _load_py_transform(transform_str)
+            log("init_context: loaded py: transform %r", self.transform)
+            return
         import torchvision.transforms.v2 as transforms
         if transform_str.startswith("functional."):
             functional = transforms.functional
@@ -187,6 +218,27 @@ class Filter:
         bgrx_array = np.frombuffer(bgrx, dtype=np.uint8)
         bgrx_array = bgrx_array.reshape(height, rowstride)
         bgrx_array = bgrx_array[:, :width * 4].reshape(height, width, 4)
+
+        if getattr(self.transform, "supports_numpy", False):
+            bgr_np = bgrx_array[:, :, :3].copy()
+            bgr_result = self.transform.process_numpy(
+                bgr_np,
+                target_x=image.get_target_x(),
+                target_y=image.get_target_y(),
+                window_width=self.width,
+                window_height=self.height,
+            )
+            result = np.concatenate([bgr_result, bgrx_array[:, :, 3:4]], axis=2)
+            bgrx = result.tobytes()
+            filtered = ImageWrapper(
+                image.get_x(), image.get_y(), width, height,
+                bgrx, image.get_pixel_format(), image.get_depth(),
+                width * 4, 4, planes=ImageWrapper.PACKED, thread_safe=True,
+            )
+            filtered.set_target_x(image.get_target_x())
+            filtered.set_target_y(image.get_target_y())
+            return filtered
+
         pixels_gpu = torch.tensor(bgrx_array, device=self.device, dtype=torch.uint8)
         # Separate BGR and X channels
         bgr = pixels_gpu[:, :, :3]          # (H, W, 3)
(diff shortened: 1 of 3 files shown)


xpra/server/subsystem/pointer.py
```diff
@@ -24,6 +25,37 @@
...
+# (chiefly the OCR/PII temporal merger, which otherwise only learns about wheel
+# events).  Two complementary signals are used to separate scrollbar-thumb drags
+# from text-selection and other button-1 drags:
+#   (a) press-X close to the right edge of the window (scrollbar zone), and
+#   (b) the drag motion remains near-vertical (low |dx|:|dy| ratio).
+# (a) is the strong signal when window geometry is available.  (b) is the
+# fallback when geometry can't be resolved.  Either signal classifying the
+# drag as 'not a scrollbar drag' suppresses emits for the rest of that drag.
+#
+#   XPRA_DRAG_SCROLL_ENABLED                  master switch
```
totaam, line 39: Should be default OFF since most users don't have any use for this (yet?).

xpra/codecs/pytorch/filter.py
```diff
@@ -179,32 +210,51 @@ def convert_image(self, image: ImageWrapper) -> ImageWrapper:
...
+            filtered = ImageWrapper(
+                image.get_x(), image.get_y(), width, height,
+                bgrx_out, image.get_pixel_format(), image.get_depth(),
+                width * 4, 4, planes=ImageWrapper.PACKED, thread_safe=True,
+            )
+            filtered.set_target_x(image.get_target_x())
+            filtered.set_target_y(image.get_target_y())
+            return filtered
+
+        import torch
```
totaam: Do we really need this branch?
Don't all the pytorch builds out there load with numpy?

totaam (commented):
LGTM!
Just 2 minor comments.

xpra/codecs/pytorch/filter.py
hamidthri: The numpy branch is for `py:` filters that have their own inference
like our OCR masker and going through torch would just add a
wasted tensor round-trip.

The torch branch needs to stay though: the default `functional.invert`
raises TypeError on numpy. Verified on torchvision 0.25.0:

    >>> import numpy as np
    >>> import torchvision.transforms.v2 as T
    >>> T.functional.invert(np.zeros((100, 100, 3), dtype=np.uint8))
    TypeError: Functional F.invert supports inputs of type dict_keys([
      <class 'torchvision.tv_tensors._image.Image'>,
      <class 'torch.Tensor'>,
      <class 'PIL.Image.Image'>,
      <class 'torchvision.tv_tensors._video.Video'>
    ]), but got <class 'numpy.ndarray'> instead.

So dropping the torch branch would break the default config and any
`functional.*` user.

xpra/server/subsystem/pointer.py
hamidthri: Switched to off now.

xpra/server/subsystem/keyboard.py
```diff
@@ -288,6 +307,18 @@ def do_process_keyboard_event(self, proto, wid: int, keyname: str, pressed: bool
                 log.estr(e)
                 log.error(" for keyname=%s, keyval=%i, keycode=%i", keyname, keyval, keycode)
         ss.emit("user-event", "key-action")
+        # Keyboard-as-scroll: notify window sources so the OCR/PII temporal
```
totaam: This commit should not refer to OCR/PII, only to scroll event detection.

xpra/codecs/pytorch/filter.py
```diff
@@ -179,32 +210,51 @@ def convert_image(self, image: ImageWrapper) -> ImageWrapper:
...
-        return image
+        bgrx_out = np.ascontiguousarray(result_cpu).tobytes()
+        filtered = ImageWrapper(
+            image.get_x(), image.get_y(), width, height,
+            bgrx_out, image.get_pixel_format(), image.get_depth(),
+            width * 4, 4, planes=ImageWrapper.PACKED, thread_safe=True,
+        )
+        filtered.set_target_x(image.get_target_x())
+        filtered.set_target_y(image.get_target_y())
+        return filtered
```
totaam, line 264: Is there a reason for using a new `ImageWrapper` instance here?
We normally explicitly free them in the compression logic, but here one just falls out of scope and gets freed through its `__del__` method.

xpra/codecs/pytorch/filter.py
```diff
@@ -179,32 +210,51 @@ def convert_image(self, image: ImageWrapper) -> ImageWrapper:
...
         pixels_gpu = torch.tensor(bgrx_array, device=self.device, dtype=torch.uint8)
-        # Separate BGR and X channels
-        bgr = pixels_gpu[:, :, :3]          # (H, W, 3)
-        x_channel = pixels_gpu[:, :, 3:4]   # (H, W, 1)
-        # Convert to CHW format for torchvision (C, H, W)
-        bgr_chw = bgr.permute(2, 0, 1)  # (3, H, W)
-        # Apply transform - RandomInvert works with uint8
+        bgr = pixels_gpu[:, :, :3]
+        x_channel = pixels_gpu[:, :, 3:4]
+        bgr_chw = bgr.permute(2, 0, 1)
```
totaam: Unless the comments are wrong, please keep them. This also reduces the diff.

xpra/codecs/pytorch/filter.py
```diff
@@ -179,32 +210,51 @@ def convert_image(self, image: ImageWrapper) -> ImageWrapper:
...
-        result_gpu = torch.cat([bgr_hwc, x_channel], dim=2)  # (H, W, 4)
-        # Download to CPU
+        bgr_hwc = bgr_transformed.permute(1, 2, 0)
+        result_gpu = torch.cat([bgr_hwc, x_channel], dim=2)
         result_cpu = result_gpu.cpu().numpy()
-        bgrx = result_cpu.tobytes()
-        image.set_pixels(bgrx)
-        image.set_rowstride(width * 4)
-        return image
+        bgrx_out = np.ascontiguousarray(result_cpu).tobytes()
```
totaam: Why do we need `np.ascontiguousarray` here?
I would expect `numpy.array.tobytes` to handle any array format?

xpra/codecs/pytorch/filter.py
```diff
@@ -179,32 +210,51 @@ def convert_image(self, image: ImageWrapper) -> ImageWrapper:
...
+        x_channel = pixels_gpu[:, :, 3:4]
+        bgr_chw = bgr.permute(2, 0, 1)
         bgr_transformed = self.transform(bgr_chw, **self.kwargs)
-        # Convert back to HWC format
-        bgr_hwc = bgr_transformed.permute(1, 2, 0)  # (H, W, 3)
-        # Recombine with X channel
-        result_gpu = torch.cat([bgr_hwc, x_channel], dim=2)  # (H, W, 4)
-        # Download to CPU
+        bgr_hwc = bgr_transformed.permute(1, 2, 0)
+        result_gpu = torch.cat([bgr_hwc, x_channel], dim=2)
```
totaam: Just as above, it took me a minute to realize that there are no changes here!

xpra/codecs/pytorch/filter.py
```diff
@@ -3,6 +3,8 @@
 # Xpra is released under the terms of the GNU GPL v2, or, at your option, any
 # later version. See the file COPYING for details.
 
+import importlib
```
totaam: Not a biggie, but please move `importlib` into `_load_py_transform`.
Default usage doesn't need it.

xpra/codecs/pytorch/filter.py
hamidthri: Yes: ImageFilter.do_process_image() frees the input wrapper immediately after convert_image() returns. If the pytorch filter mutates and returns that same wrapper, compression receives an already-freed object. Returning a new wrapper gives the filtered pixels their own lifetime, matching the existing pillow filter behavior.

xpra/codecs/pytorch/filter.py
totaam: @hamidthri bump?

Issue #4938, opened 2026-06-22
Shadow ssh fails on xpra v6.5 (Ubuntu)


tribhuvanji:
Hi,
Hoping this is the right place to report/ask about this.  

**Problem:**
Command:    _xpra shadow ssh://server_   
Result:     Time out, SIGTERM  (SSH successful and xpra server starts as expected, 30 second wait for timeout and TERM)

(see below for command debug output)

**Steps to reproduce:**   upgrade to xpra v6.5-r0 and run _xpra shadow ssh://server_
**Step to resolve:**   downgrade to xpra v6.4.4-r0  and run _xpra shadow ssh://server_

**Platform:** Ubuntu 24.04.4 LTS latest patch level using xpra from the repositories

When running xpra client lower than v6.5  _xpra shadow ssh_ is successful to any xpra server, old or new.
This makes me suspect something in xpra-client 6.5 shadow is different.
FYI When using xpra client v6.5  attach works fine but not shadow.

Is something broken with shadow in xpra 6.5-r0 or are there additional configurations required to executing xpra shadow ssh in version 6.5?

My apologies if this has already been addressed.
I can provide debugging info if needed.

Thanks so much, 
-Steve

Snippit from command output when using 6.5 :

`xpra shadow ssh://myserver`
```
2026-06-22 09:18:00,488  SSH: 'Entering daemon mode; any further errors will be reported to:'                           
2026-06-22 09:18:00,489  SSH: "  '/run/user/1000/xpra/0/server.log'"                                                    
2026-06-22 09:18:02,329 created unix domain sockets:                                                                    
2026-06-22 09:18:02,329  '/run/user/1000/xpra/clients/sunroom-30766'                                                    
2026-06-22 09:18:40,351 connection timed out                                                                            
                                                                                                                        
2026-06-22 09:18:40,452 splash: Splash got signal SIGTERM                                                               
2026-06-22 09:18:40,453 0 packets received                                                                              
2026-06-22 09:18:40,453 1 packets sent (5K bytes)
```

totaam:
Is there anything useful in `/run/user/1000/xpra/0/server.log`?
Is this using the paramiko backend? Can you run with `-d ssh`?
What version is installed on the server?
What sessions are running there? How many? `xpra displays` output?

tribhuvanji:
Thanks. 
[...]

PR #4941, opened 2026-06-26
launcher: start the protocol so the client reads the server's replies

andrewachen:
The GTK launcher's built-in connection path (`do_start_xpra_client`) calls `make_protocol()` but never calls `protocol.start()`.

`make_protocol()` builds the `SocketProtocol` and schedules `send_hello()`, but the network read thread is only started by `protocol.start()` — which the command-line client does explicitly (`xpra/scripts/main.py`) while the launcher does not.

As a result, a client started from the GTK launcher sends its hello but never reads anything back: the connection reports "network connection established" and then times out with "0 packets received" on every transport (tcp, ssl, quic).

This regressed in 25d6e80b05c, which split `make_protocol()` out of `setup_connection()` and moved the `start()` responsibility to the callers; the command-line callers were updated but the launcher was missed.

The fix calls `protocol.start()` after `make_protocol()`, matching the command-line client. Verified on Windows (GTK launcher, built-in connect): with this change the launcher connects over both tcp and quic; without it the client hangs after "network connection established".

Sponsored-By: Netflix

### DIFF
--- a/xpra/client/gtk3/launcher.py
+++ b/xpra/client/gtk3/launcher.py
@@ -896,7 +896,12 @@ class ApplicationWindow:
         log("do_start_XpraClient(%s, %s) client=%s", conn, display_desc, self.client)
         self.client.encoding = self.config.encoding
         self.client.display_desc = display_desc
-        self.client.make_protocol(conn)
+        protocol = self.client.make_protocol(conn)
+        # make_protocol() does not start the protocol: the network read thread
+        # is only started by protocol.start() (as the command-line client does).
+        # Without this the client sends its hello but never reads the server's
+        # reply, so the connection times out with 0 packets received.
+        protocol.start()
         self.set_info_text("Network connection established")
         log("start_XpraClient() client initialized")
 


totaam:
Already done in 0cff4450444b64615dfcf925f6c1d8e7d568e09a / da3fa8745e7e67845e63cf6d80ed31c768f97138, details in #4938

andrewachen:
Thanks!

andrewachen:
@totaam thanks for the quick triage. #4938 does address the same underlying symptom (the protocol never gets started → `0 packets received`), but I think it fixes a *different caller* and the GTK launcher path is still affected.

All of the #4938 commits touch `run_remote_server` (`xpra/scripts/main.py`) and `socket_handler.py`:
- master: `da3fa8745e`, `b32003f8f1` (+ `0cff4450` for the idempotent `start()`)
- v6.5.x: `e7dc2a1197`, `c911a3f62a`

None of them touch `xpra/client/gtk3/launcher.py`. The GTK launcher's built-in connect doesn't go through `run_remote_server` — it goes through `do_start_xpra_client`, which on both current master (`6c6e81f46e`) and v6.5.x (`deeeedcc88`) still does:

```python
self.client.make_protocol(conn)
self.set_info_text("Network connection established")
```

i.e. no `protocol.start()`. The regular GTK client's `run()` chain doesn't start it either (only the command client's `run()` does, in `command.py`). So launching a session from the GTK launcher reproduces the exact #4938 symptom — hello sent, then `0 packets received` and an idle timeout after "network connection established" — over tcp, ssl and quic.

This PR's one-liner calls `protocol.start()` after `make_protocol()` in `do_start_xpra_client`, matching what `run_remote_server` now does. Verified on Windows over both tcp and quic: without it the launcher hangs; with it, it connects.

Would you reconsider this PR (or fold the launcher fix into #4938)? Happy to rebase onto master/v6.5.x, or rework it to lean on the idempotent `start()` from `0cff4450`, whichever you prefer.

totaam:
My bad, I thought that the earlier commit had fixed both!

Issue #3703, opened 2022-12-06
nvdec hardware accelerated decoder


totaam:
Split from #202.
There are other frameworks for hardware acceleration, but nvidia's offering is generally the most stable option.

https://docs.nvidia.com/video-technologies/video-codec-sdk/nvdec-video-decoder-api-prog-guide/

totaam:
Now working for `jpeg` decoding without OpenGL - which is not very useful since we have `nvjpeg` (#3504) which does this but better in every way: more lightweight, with opengl acceleration (decoded image stays on the GPU).
But this is a start, hopefully I can figure out how to enable `h264`, `hevc`, `vp8`, `vp9`, etc..
Still TODO:
* fix `no device` errors when used as a video decoder for `jpeg` - allocate the CUDA device once in the decode thread?
* OpenGL - CUDA buffer sharing (very similar to `nvjpeg`)
* add scaling to `libyuv`'s `NV12 to RGB`

totaam:
It paints... but needs a new shader: this one only paints the `Y` plane for now. (better than a corrupted screen, or worse.. crashes)

totaam:
Trying to get `nvdec` to work via gstreamer.

It finds the same valid encodings and chroma formats:
```shell
GST_DEBUG=nvdec*:4 gst-inspect-1.0 nvh264dec 
0:00:00.902346129 655952 0x55f02fa992d0 INFO               nvdecoder gstnvdecoder.c:1171:gst_nv_decoder_check_device_caps: mpegvideo bit-depth 8 with chroma format 0 [48 - 4080] x [16 - 4080]
0:00:00.908428330 655952 0x55f02fa992d0 INFO               nvdecoder gstnvdecoder.c:1171:gst_nv_decoder_check_device_caps: mpeg2video bit-depth 8 with chroma format 0 [48 - 4080] x [16 - 4080]
0:00:00.914007625 655952 0x55f02fa992d0 INFO               nvdecoder gstnvdecoder.c:1171:gst_nv_decoder_check_device_caps: mpeg4video bit-depth 8 with chroma format 0 [48 - 2032] x [16 - 2032]
0:00:00.920519713 655952 0x55f02fa992d0 INFO               nvdecoder gstnvdecoder.c:1094:gst_nv_decoder_check_device_caps: No codec map corresponding to codec 3
[...]

Issue #3504, opened 2022-03-27
nvjpeg decoder


totaam:
Same as #2984 but for client side decoding this time.

This will allow us to achieve the best possible single picture latency.

Once implemented, it would make sense to tell the server to prefer `jpeg` over `webp` in all(?) cases.

totaam:
The inspiration for the GPU code above was not from [nvidia's own documentation](https://docs.nvidia.com/cuda/nvjpeg/index.html) or sample code - which fails to tell you that the buffer must be a GPU buffer... and the API doesn't fail if you use a CPU buffer either.
Some good code samples:
[using pycuda and glumpy to draw pytorch GPU tensors to the screen without copying to host memory](https://gist.github.com/victor-shepardson/5b3d3087dc2b4817b9bffdb8e87a57c4)
[pycuda gl API documentation](https://documen.tician.de/pycuda/gl.html)

Still TODO: move GPU device selection and keep the context alive as creating it is very expensive.
(and perhaps re-create it when the window is moved to another monitor?)

totaam:
Testing with:
```shell
XPRA_OPENGL_DRAW_REFRESH=0 XPRA_OPENGL_NVJPEG=1 xpra attach tcp://127.0.0.1:10000/ \
    --no-speaker --no-mmap --opengl=force -d nvjpeg,opengl --quality=10
```
A `glxgears` 4K screen update:
```
23,158 glXMakeCurrent: xid=0x5c00033, context=<OpenGL.raw.GLX._types.LP_struct___GLXcontextRec object at 0x7f5eeceb1ac0>
23,159 decompress_with_device(RGB, 141256 bytes, None)
23,159 got image info: 3840x2123 YUV420
23,181 nvjpegDecode took 21ms
23,182 copy_buffer(<pycuda._driver.DeviceAllocation object at 0x7f5eecf14a60>, 24456960)
23,186 RegisteredBuffer(1, pycuda._driver.graphics_map_flags.WRITE_DISCARD)=<pycuda._driver.RegisteredBuffer object at 0x7f5eecec9db0>
23,187 copying from <pycuda._driver.DeviceAllocation object at 0x7f5eecf14a60> to mapping=<pycuda._driver.RegisteredMapping object at 0x7f5eecf14c40> at 0x7f5daa400000
[...]

PR #4964, opened 2026-07-17
libva: X11 VADisplay support, size/capacity-aware decoder selection, H.264 multi-reference DPB

Wenri:
This series makes the libva client decoder usable on X11-only VA drivers and implements a proper H.264 multi-reference DPB, replacing the current single-reference behaviour. Developed and validated on a pre-DRM NVIDIA setup (GeForce GT 130, driver 340.xx — no render node) through the VDPAU-backed VA driver, but every change is generic.

### 1. `libva: X11 VADisplay support + register the video decoder`
- `libva_open_display()` accepts device `x11[:<name>]` (XOpenDisplay + `vaGetDisplay`, with symmetric teardown) and the decoder device scan falls back to `x11` when no render node works. VDPAU-backed VA has no DRM constructor at all, so DRM-only display opening made the module unusable there.
- `get_output_frame()` falls back from `vaDeriveImage` to `vaCreateImage`+`vaGetImage` (the ffmpeg hwcontext pattern) for drivers whose surfaces have no linear CPU view.
- `dec_libva` is registered in `CODEC_TO_MODULE` / `ALL_VIDEO_DECODER_OPTIONS` — the module shipped without the decoder being selectable (`--video-decoders=libva` warned "unknown").

### 2. `client paint: size- and capacity-aware video decoder selection`
Two production failure modes motivated this:
- a stream a decoder can *never* handle (over its size envelope) still got selected; each guaranteed `init_context` failure bumped the shared spec's `setup_cost` until the decoder was delisted session-wide — one oversized window permanently demoted the hardware decoder for every window.
- hardware that corrupts *concurrent* streams (measured: two interleaved VDPAU decode streams damage each other's luma) had no client-side concurrency control: `max_instances` existed in `CodecSpec` but the client decoder path neither tracked instances nor read it.

So: new `CodecSpec.max_pixels` (area limits that `max_w/max_h` can't express, e.g. VDPAU feature set A = 8192 macroblocks but 2048/side), a hard envelope filter before instantiation, decoder creation through `make_instance()`, and `choose_decoder()` ranking by `(100 - setup_cost) × get_runtime_factor()` — the same capacity mechanism `video_scoring.py` already applies to encoders/CSC. The libva spec declares its envelope (env-tunable), `max_instances=1`, `setup_cost=0`, and accepts odd display sizes (streams are coded even + SPS crop; the NV12 copy-out now even-rounds the chroma stride).

Note on `setup_cost=0`: with the current cost-only ranking, libva's 50 vs openh264's 0 meant the hardware decoder could never win. Zero makes the tie-break the user's `--video-decoders` order. Happy to adjust if you'd rather express this differently.

### 3. `libva: H.264 multi-reference DPB`
The decoder kept a single reference (`ReferenceFrames[0]`, 4 surfaces, every frame — even non-references — overwrote it). Real encoder output is multi-ref whenever speed drops (x264 faster/fast/medium default ref=2/2/3; NVENC emits `max_num_ref_frames=3` at quality presets), and multi-ref slices against the wrong reference produce whole-window MC garbage.

Implemented for progressive streams: 16-entry DPB + 17-surface pool, POC types 0/2 (type 1 cleanly rejected), `dec_ref_pic_marking` parsed and applied (sliding window + MMCO 1–6 incl. MMCO5, IDR flush, long-term), `nal_ref_idc` gating, `ref_pic_list_modification` parsed with the spec-default P list + modifications filling `RefPicList0`. B slices are rejected at slice-header parse (no display-order reordering here; clean rejection lets the paint code fall back instead of decoding garbage). `decompress_image` now raises `CodecStateException` on mid-stream errors so the paint code restarts the decoder (a `RuntimeError` propagates and leaves a stale decoder in place).

### Validation
- x264 IPP `ref=1/2/3/16` @1280×720 and `ref=3` @1920×1080, 120 frames each: **bit-exact** (per-frame NV12 MD5) vs ffmpeg software decode, through the VDPAU-backed VA driver on the X11 display path.
- Odd display size (1397×785 view of a 1398×786 stream): byte-identical to the cropped even-size decode.
- A VDPAU call interceptor confirmed rotating multi-entry `referenceFrames` reaching `VdpDecoderRender` and decoder creation with `max_references=3`.
- Live xpra session against an NVENC `nrf=3` stream: sustained hardware decode, zero errors.

Happy to split this into separate PRs if preferred.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3689,10 +3689,10 @@ if vpl_encoder_ENABLED:
 toggle_packages(libva_ENABLED, "xpra.codecs.libva")
 if libva_encoder_ENABLED:
     ace("xpra.codecs.libva.encoder,xpra/codecs/libva/va_encode.c",
-        "libva,libva-win32" if WIN32 else "libva,libva-drm")
+        "libva,libva-win32" if WIN32 else "libva,libva-drm,libva-x11,x11")
 if libva_decoder_ENABLED:
     ace("xpra.codecs.libva.decoder,xpra/codecs/libva/va_decode.c",
-        "libva,libva-win32" if WIN32 else "libva,libva-drm")
+        "libva,libva-win32" if WIN32 else "libva,libva-drm,libva-x11,x11")
 toggle_packages(gstreamer_ENABLED, "xpra.gstreamer")
 toggle_packages(remote_encoder_ENABLED, "xpra.codecs.remote")
 
--- a/xpra/client/gui/window/backing.py
+++ b/xpra/client/gui/window/backing.py
@@ -125,16 +125,21 @@ def rgba_text(text: str, width: int = 64, height: int = 32, x: int = 20, y: int 
 
 
 def choose_decoder(decoders_for_cs: list[CodecSpec], max_setup_cost=100) -> CodecSpec:
-    # for now, just rank by setup-cost, so slow-to-initialize decoders come last:
-    scores: dict[int, list[int]] = {}
+    # rank by setup-cost (so slow-to-initialize decoders come last),
+    # scaled by the spec's runtime factor - the same mechanism the
+    # server's pipeline scoring uses (video_scoring.py): a spec at its
+    # `max_instances` capacity returns 0 and loses to any available
+    # alternative, but remains eligible if there is nothing else
+    scores: dict[float, list[int]] = {}
     for index, decoder_spec in enumerate(decoders_for_cs):
         cost = decoder_spec.setup_cost
         if cost > max_setup_cost:
             continue
-        scores.setdefault(cost, []).append(index)
+        score = (100 - cost) * decoder_spec.get_runtime_factor()
+        scores.setdefault(score, []).append(index)
     if not scores:
         raise RuntimeError("no decoders available!")
-    best_score = sorted(scores)[0]
+    best_score = sorted(scores)[-1]
     options_for_score = scores[best_score]
     # if multiple decoders have the same score, just use the first one:
     chosen = decoders_for_cs[options_for_score[0]]
@@ -820,12 +825,41 @@ class WindowBackingBase:
                 if not all_decoders_for_cs:
                     raise RuntimeError(f"no video decoders for {coding!r} and {input_colorspace!r}")
                 decoders_for_cs = list(all_decoders_for_cs)
+
+                # filter out decoders whose hard limits this stream
+                # exceeds (min/max dimensions, max_pixels area, ie:
+                # driver decode envelopes), so that such streams fall
+                # back to another decoder instead of poisoning the
+                # shared spec's setup_cost via guaranteed init_context
+                # failures (which at >100 delists the decoder for the
+                # entire session).
+                # (`max_instances` capacity is handled by
+                # choose_decoder() via get_runtime_factor, like the
+                # server's pipeline scoring - and dimension masks are
+                # deliberately NOT checked here: real decoders handle
+                # padded+cropped sizes regardless of their alignment
+                # masks, and pre-filtering on them can leave an
+                # at-capacity decoder as the only candidate)
+                def within_limits(ds) -> bool:
+                    return (ds.min_w <= enc_width <= ds.max_w
+                            and ds.min_h <= enc_height <= ds.max_h
+                            and (ds.max_pixels <= 0 or enc_width * enc_height <= ds.max_pixels))
+
+                hard_ok = [ds for ds in decoders_for_cs if within_limits(ds)]
+                if hard_ok:
+                    decoders_for_cs = hard_ok
+                else:
+                    videolog.warn(f"Warning: no decoder is rated for {enc_width}x{enc_height} {coding} frames")
+                    videolog.warn(" trying anyway with: " + csv(d.codec_type for d in decoders_for_cs))
                 while not self._video_decoder:
                     decoder_spec = choose_decoder(decoders_for_cs)
                     videolog("paint_with_video_decoder: new %s%s",
                              decoder_spec.codec_type, (coding, enc_width, enc_height, input_colorspace))
                     try:
-                        vd = decoder_spec.codec_class()
+                        # make_instance (not codec_class()) so the spec's
+                        # instance WeakSet tracks live decoders — that is
+                        # what max_instances gating counts
+                        vd = decoder_spec.make_instance()
                         vd.init_context(coding, enc_width, enc_height, input_colorspace, options)
                         self._video_decoder = vd
                         break
@@ -833,7 +867,7 @@ class WindowBackingBase:
                         log("%s.init_context(..)", vd, exc_info=True)
                         log.warn(f"Warning: failed to initialize decoder {decoder_spec.codec_type}: {e}")
                         decoder_spec.setup_cost += 10
-                    except RuntimeError as e:
+                    except (RuntimeError, ValueError) as e:
                         log("%s.init_context(..)", vd, exc_info=True)
                         log.warn(f"Warning: failed to initialize decoder {decoder_spec.codec_type}: {e}")
                         decoder_spec.setup_cost += 50
--- a/xpra/codecs/libva/decoder.pyx
+++ b/xpra/codecs/libva/decoder.pyx
@@ -12,7 +12,7 @@ from time import monotonic
 from typing import Any, Dict, Tuple
 from collections.abc import Sequence
 
-from xpra.codecs.constants import VideoSpec, EncodingNotSupported
+from xpra.codecs.constants import VideoSpec, EncodingNotSupported, CodecStateException
 from xpra.codecs.vacommon import config_libva_logging
 from xpra.codecs.image import ImageWrapper
 from xpra.common import SizedBuffer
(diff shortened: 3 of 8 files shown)


xpra/client/gui/window/backing.py
```diff
@@ -125,16 +125,21 @@ def rgba_text(text: str, width: int = 64, height: int = 32, x: int = 20, y: int
 
 
 def choose_decoder(decoders_for_cs: list[CodecSpec], max_setup_cost=100) -> CodecSpec:
-    # for now, just rank by setup-cost, so slow-to-initialize decoders come last:
-    scores: dict[int, list[int]] = {}
+    # rank by setup-cost (so slow-to-initialize decoders come last),
```
totaam, line 128: Note: That's not necessarily what we want here.
The server does this, but eventually settles on efficient encoders after a while, even sending non-video frames during the encoder's warmup.
So we may have to change this part too...

xpra/client/gui/window/backing.py
```diff
@@ -820,20 +825,49 @@ def restart(msg: str, *args) -> None:
...
                         break
                     except TransientCodecException as e:
                         log("%s.init_context(..)", vd, exc_info=True)
                         log.warn(f"Warning: failed to initialize decoder {decoder_spec.codec_type}: {e}")
                         decoder_spec.setup_cost += 10
-                    except RuntimeError as e:
+                    except (RuntimeError, ValueError) as e:
                         log("%s.init_context(..)", vd, exc_info=True)
                         log.warn(f"Warning: failed to initialize decoder {decoder_spec.codec_type}: {e}")
                         decoder_spec.setup_cost += 50
```
totaam, line 873: This chunk is very useful and can be merged as-is.
(I just never hit those limits myself)

xpra/codecs/libva/decoder.pyx
```diff
@@ -155,7 +156,12 @@ cdef bint supports(str encoding, str colorspace):
     return bool(libva_decode_supports(enc, cs))
 
 
-MAX_WIDTH, MAX_HEIGHT = 8192, 8192
+# driver limits, overridable for hardware whose VA driver cannot
+# express them (ie: the VDPAU bridge on feature-set-A NVIDIA: H264
+# maxes at 2048x2048 AND 8192 macroblocks = 2097152 pixels):
```
totaam, line 161: Why would you use VDPAU with nvidia when we have nvjpeg (#3504) and nvdec  (#3703) ?
These should be more reliable.

xpra/codecs/libva/decoder.pyx
```diff
@@ -174,10 +180,25 @@ def get_specs() -> Sequence[VideoSpec]:
...
                 quality=80, speed=90,
                 size_efficiency=70,
-                setup_cost=50,
+                # setup_cost 0 (was 50): choose_decoder() ranks by
+                # setup_cost alone and openh264 declares 0, so any
+                # nonzero value here means the hardware decoder can
+                # NEVER be selected when openh264 is present.  With
+                # equal cost the tie-break is list order = the user's
+                # --video-decoders order, which is the right authority.
+                setup_cost=0,
```
totaam, line 189: I don't think it's right to lower the setup-cost to allow this encoder to be selected.
libva decoders are slower to initialize than openh264.
Maybe we should select based on cpu-cost / gpu-cost instead.

xpra/codecs/libva/decoder.pyx
```diff
@@ -203,8 +224,15 @@ cdef class Decoder:
...
-            raise ValueError("invalid odd width %i or height %i" % (width, height))
+        # odd display sizes are fine: the bitstream is coded at the
+        # padded even size (SPS crop carries the display size) and the
+        # NV12 copy-out handles odd width/height (ceil'd chroma rows,
+        # even-rounded chroma stride)
+        if width > MAX_WIDTH or height > MAX_HEIGHT or (MAX_PIXELS > 0 and width * height > MAX_PIXELS):
+            # belt-and-braces for callers that bypass the size-aware
+            # decoder selection: fail here (cleanly, before creating a
+            # decoder that would error on every submit)
+            raise RuntimeError("%ix%i exceeds this device's decode limits" % (width, height))
```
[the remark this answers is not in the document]
totaam: This copilot suggestion is overkill: we don't currently support the huge picture sizes that could ever overflow this.
But I guess the extra care is cheap.

xpra/codecs/libva/decoder.pyx
```diff
@@ -298,7 +326,12 @@ cdef class Decoder:
         if status != LIBVA_DEC_OK:
             detail = libva_decoder_get_last_error(self.context).decode("utf-8", "replace")
             last_sts = libva_decoder_get_last_status(self.context)
-            raise RuntimeError("libva decode error: %s (detail: %s, sts=%d)" % (
+            # CodecStateException (not RuntimeError): the paint code
+            # restarts the decoder on it (backing.py), whereas other
+            # exception types propagate and leave a stale decoder in
+            # place; a mid-stream failure here always invalidates the
+            # decoder state (DPB / reference chain)
+            raise CodecStateException("libva decode error: %s (detail: %s, sts=%d)" % (
```
totaam, line 334: Are there other places where we end up leaving a stale decoder behind?

xpra/codecs/libva/va_common.h
```diff
@@ -12,6 +12,8 @@
 #include <va/va_win32.h>
 #else
 #include <va/va_drm.h>
+#include <va/va_x11.h>
+#include <X11/Xlib.h>
```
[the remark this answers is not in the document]
totaam: This is a valid concern - xpra can be built `--without-x11` and codecs should not link to X11 in that case.

`#ifdef`s are a pain in Cython, alternatively you could move the x11 glue code to a file and only include it in the `setup.py` compilation when `x11_ENABLED` is `True`. (and have an empty shim for building without)

xpra/codecs/libva/va_common.h
```diff
@@ -167,17 +169,75 @@ static inline int libva_open_display(const char *device, int *fd_out, VADisplay
...
+            XCloseDisplay(dpy);
+            return 0;
+        }
+        vstr = vaQueryVendorString(display);
+        snprintf(vendor, vendor_size, "%s", vstr ? vstr : "");
+        libva_x11_register(display, dpy);
+        *display_out = display;
+        return 1;
+    }
+    fd = open(device, O_RDWR | O_CLOEXEC);
```
totaam, line 240: This chunk could easily be separated out.

xpra/codecs/libva/va_decode.c
```diff
@@ -1201,8 +1594,14 @@ static LibVADecodeStatus map_output(LibVADecoder *dec, VASurfaceID surface,
...
     if (image.format.fourcc == VA_FOURCC_NV12) {
+        /* NV12 chroma rows hold 2 bytes per 2-pixel pair, so an odd
+         * display width needs w+1 bytes per chroma row; the source
+         * surface is coded-size aligned so the bytes always exist.
+         * (Odd display sizes are normal: h264 codes the padded even
+         * size and crops via the SPS.) */
+        int cw = (w + 1) & ~1;
         size_t ysize = (size_t)w * h;
-        size_t uvsize = (size_t)w * ((h + 1) / 2);
+        size_t uvsize = (size_t)cw * ((h + 1) / 2);
```
totaam, line 1604: This looks reasonable and ties with the width_mask change, right?

xpra/codecs/libva/va_decode.c
```diff
@@ -2182,23 +2581,40 @@ static LibVADecodeStatus h264_decoder_decode(LibVADecoder *dec,
...
 
     memset(&pic, 0, sizeof(pic));
     pic.CurrPic.picture_id = surface;
     pic.CurrPic.frame_idx = (uint32_t)first->frame_num;
-    pic.CurrPic.flags = VA_PICTURE_H264_SHORT_TERM_REFERENCE;
-    pic.CurrPic.TopFieldOrderCnt = first->poc_lsb;
-    pic.CurrPic.BottomFieldOrderCnt = first->poc_lsb;
+    pic.CurrPic.flags = first->nal_ref_idc ? VA_PICTURE_H264_SHORT_TERM_REFERENCE : 0;
+    pic.CurrPic.TopFieldOrderCnt = top_foc;
+    pic.CurrPic.BottomFieldOrderCnt = bottom_foc;
```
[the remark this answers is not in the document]
totaam: This copilot comment is over my head, but looks legit.

xpra/codecs/video.py
```diff
@@ -76,7 +77,7 @@ def try_import_modules(prefix: str, *codec_names: str) -> list[str]:
                                             "nvenc", "vpl", "libva", "nvjpeg", "jpeg", "webp", "remote")
 HARDWARE_ENCODER_OPTIONS: Sequence[str] = ("nvenc", "vpl", "libva", "nvjpeg", "vt")
 ALL_CSC_MODULE_OPTIONS: Sequence[str] = ("cython", "libyuv")
-ALL_VIDEO_DECODER_OPTIONS: Sequence[str] = ("vpl", "mf", "nvdec", "openh264", "vpx", "aom", "de265")
+ALL_VIDEO_DECODER_OPTIONS: Sequence[str] = ("vpl", "mf", "nvdec", "libva", "openh264", "vpx", "aom", "de265")
```
totaam, line 80: As per above,  I don't think that the list order is the right way of choosing a decoder, spec details are - that's what they're for.

setup.py
```diff
@@ -3689,10 +3689,10 @@ def bundle_tests() -> None:
 toggle_packages(libva_ENABLED, "xpra.codecs.libva")
 if libva_encoder_ENABLED:
     ace("xpra.codecs.libva.encoder,xpra/codecs/libva/va_encode.c",
-        "libva,libva-win32" if WIN32 else "libva,libva-drm")
+        "libva,libva-win32" if WIN32 else "libva,libva-drm,libva-x11,x11")
```
totaam, line 3692: As per above, `x11` bits should only be enabled if `x11_ENABLED`.

totaam (changes requested):
Please try to split this PR so it is more manageable.
ie: the validation / size mask stuff should be easy to merge.
The x11 hooks are harder.

totaam:
See also these changes which will require a rebase:
* 8fc29367f550ec0a70b7c077bd36ea757871d23f
* d12d7d03f557695d3bc05f38a13718bc3f79f01c
* 6bba997299c3eb752e4170a3bab9f50ba3a28170
* 900e57c327f0079ad6e9ac8b8d941587851448c1

PR #5002, opened 2026-08-19
Fix Wayland frame callbacks for empty-damage commits

kogeler:
## Summary

- Prevent native Wayland seamless clients from stalling visual updates after a mapped commit with an empty damage list, even though pointer and keyboard events continue to arrive. A manual refresh advances one frame because the application is still waiting for its Wayland frame callback.
- Acknowledge mapped empty-damage commits after surface, size, colourspace, and subsurface bookkeeping. With no rectangles, no encoding work is queued, so this direct acknowledgement supplies `frame_done()` and flushes the client.
- Leave unknown windows and unmapped empty commits unchanged. Commits with damage rectangles still use the normal damage/encoding acknowledgement path, without an extra immediate acknowledgement.
- Add deterministic coverage for mapped empty-damage, normal damaged, unmapped empty-damage, unknown-window, and subsurface-ordering behavior. No documentation change is needed for this internal backend correctness fix.
- GitHub searches found no issue or pull request describing the same frame-callback failure. The closest context is [#4877](https://github.com/Xpra-org/xpra/issues/4877), which covers hardware-accelerated Wayland clients and dmabuf rather than empty-damage stalls.

## Test plan

- [x] Focused Wayland window unit tests (4 passed)
- [x] Xpra unit-test runner with native Wayland support (4 passed)
- [x] Cython-enabled server and Wayland build (4 passed)
- [x] Ruff and flake8 checks
- [x] Python compile check
- [x] `git diff --check`

### DIFF
--- a/tests/unittests/unit/wayland/window_test.py
+++ b/tests/unittests/unit/wayland/window_test.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python3
+# This file is part of Xpra.
+# Copyright (C) 2026 Antoine Martin <antoine@xpra.org>
+# Xpra is released under the terms of the GNU GPL v2, or, at your option, any
+# later version. See the file COPYING for details.
+
+import sys
+import unittest
+from types import ModuleType
+from unittest.mock import Mock, patch
+
+
+def load_window_server_class():
+    modules = {}
+    for module_name, class_name in (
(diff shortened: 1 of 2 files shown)


totaam:
Thanks!

PR #5003, opened 2026-08-19
Link Wayland helper extensions against wayland-server

kogeler:
## Summary

- Link the Wayland `events` and `display` helpers directly against
  `wayland-server`, which provides the symbols they call.
- Avoid clean-process import failures for `wl_list_insert`, `wl_list_remove`,
  and `wl_display_flush_clients`; relying on transitive linkage or import order
  is unreliable with `--as-needed` linking.
- Verify each helper import in a separate fresh subprocess so another Wayland
  extension cannot mask a missing direct dependency.

## Test plan

- [x] Clean Wayland build, isolated imports, ELF dependency checks, and complete
  Wayland unit-test subset
- [x] Full supported unit-test checklist: 272 successful, 0 failed, 7 requested
  slow-test exclusions
- [x] Ruff, Flake8, and `git diff --check`
- [ ] Additional `CYTHONIZE_MORE=with` build: current upstream stops before the
  tests in `xpra/x11/subsystem/window.py` because `ss` is referenced before
  assignment; this code is outside this change

No documentation or release-note changes are needed for this build-linkage
correction.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3771,8 +3771,8 @@ if wayland_server_ENABLED:
         print("generating %r" % header_path)
         subprocess.run(["wayland-scanner", "server-header", xml_path, header_path])
     wlr_args = ["-DWLR_USE_UNSTABLE", "-I./xpra/wayland/server/"]
-    ace("xpra.wayland.server.events", "wlroots-0.19", extra_compile_args=wlr_args)
-    ace("xpra.wayland.server.display", "wlroots-0.19", extra_compile_args=wlr_args)
+    ace("xpra.wayland.server.events", "wlroots-0.19,wayland-server", extra_compile_args=wlr_args)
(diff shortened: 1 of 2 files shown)


kogeler:
The optional `unittest (with)` job failure is unrelated to this change.

The new `unit.wayland.linkage_test` passed in that job. The run completed 271 tests successfully and had one failure: `unit.net.paramiko_client_test`, where `test_matching_key_returns_empty_string` failed at line 85. The same Paramiko test passed in the required `unittest (without)` job.

This failure is in the existing SSH host-key path when the extra Cythonization leg is enabled. This PR only changes the link dependencies of the two Wayland extensions and adds their isolated-import regression test; it does not touch the SSH or Paramiko code.

PR #5004, opened 2026-08-19
Detect libyuv installations without pkg-config metadata

kogeler:
## Summary

- Detect the standard `libyuv.h` header on supported Unix systems when pkg-config metadata is unavailable, allowing Ubuntu 26.04's `libyuv-dev` package to enable the converter.
- Keep pkg-config authoritative when available and otherwise pass the direct `-lyuv` linkage using the same inline fallback pattern as the existing PAM build logic.
- Preserve the existing platform gates and explicit enable or disable behavior.

## Test plan

- [x] Clean Ubuntu 26.04 fallback build without `libyuv.pc`, including converter import, ELF dependency checks, and BGRX-to-NV12 conversion
- [x] Explicit enable without usable dependencies fails clearly; explicit disable remains disabled
- [x] Focused Cython-enabled client/libyuv build and relevant codec tests
- [x] Full interpreted unit-test suite: 272 successful, 0 failed, 7 intended skips
- [x] Ruff, Flake8, and `git diff --check`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -536,7 +536,9 @@ nvdec_ENABLED           = False
 nvfbc_ENABLED           = nvidia_ENABLED and not ARM and pkg_config_exists("nvfbc")
 cuda_kernels_ENABLED    = nvidia_ENABLED and (nvenc_ENABLED or nvjpeg_encoder_ENABLED)
 cuda_rebuild_ENABLED    = None if (nvidia_ENABLED and not WIN32) else False
-csc_libyuv_ENABLED      = DEFAULT and pkg_config_exists("libyuv")
+libyuv_pkgconfig        = pkg_config_exists("libyuv")
+libyuv_fallback         = not libyuv_pkgconfig and POSIX and not OSX and has_header_file("libyuv.h")
+csc_libyuv_ENABLED      = DEFAULT and (libyuv_pkgconfig or libyuv_fallback)
 gstreamer_ENABLED       = DEFAULT
 example_ENABLED         = DEFAULT
 win32_tools_ENABLED     = WIN32 and DEFAULT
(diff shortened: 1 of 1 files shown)


totaam:
This change is too big for what it does - AI style.
Please keep it simple, we have other libs that have fallbacks and they're not that hard to read.
It's not necessarily a prime example of how to do things, but at least it's easier to parse IMO:
https://github.com/Xpra-org/xpra/blob/d97c6d9c951d616f69b7215c43673d4eabaa1cb8/setup.py#L3530-L3538

kogeler:
Just a note on the previous CI run: the libyuv extension built successfully, and its codec tests passed. The only failure was `unit.net.paramiko_client_test` (`TestVerifyHostkey.test_matching_key_returns_empty_string`), which appears unrelated to this build change; the non-Cython job passed.

kogeler:
Thanks for the concrete pointer to the existing PAM fallback. It made the preferred style clear, and I have simplified this change to follow the same local pattern.

For context, the underlying issues came from hands-on testing, not static analysis by AI agents. I ran into them while setting up the latest Xpra release for an application on a remote server that uses Vulkan through the native Wayland backend, with hardware-accelerated H.264 encoding and decoding for the stream. That setup exposed a few other problems as well; I am still working through them and will prepare separate focused pull requests later.

totaam:
This caused a regression on some of the older build hosts, like the CI for the 6.5.x branch:
```
...
    File "/usr/lib/python3/dist-packages/setuptools/_distutils/command/build_ext.py", line 345, in run
      self.build_extensions()
    File "/usr/lib/python3/dist-packages/setuptools/_distutils/command/build_ext.py", line 467, in build_extensions
      self._build_extensions_serial()
    File "/usr/lib/python3/dist-packages/setuptools/_distutils/command/build_ext.py", line 493, in _build_extensions_serial
      self.build_extension(ext)
    File "/usr/local/lib/python3.12/dist-packages/Cython/Distutils/build_ext.py", line 136, in build_extension
      super().build_extension(ext)
    File "/usr/lib/python3/dist-packages/setuptools/_distutils/command/build_ext.py", line 575, in build_extension
      libraries=self.get_libraries(ext),
                ^^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/lib/python3/dist-packages/setuptools/_distutils/command/build_ext.py", line 788, in get_libraries
      return ext.libraries + py37compat.pythonlib()
             ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
  TypeError: can only concatenate tuple (not "list") to tuple
```
The fix has been committed - just not pushed yet.

PR #5008, opened 2026-08-20
x11 bindings: only match `PropertyNewValue` events in `get_server_time`

zardus:
`get_server_time` changes a property and waits for the `PropertyNotify`,
but deleting the property afterwards queues a second notify with state
`PropertyDeleted`. The predicate matched both, so every call after the
first consumed the previous call's leftover event and returned a stale
timestamp — two successive calls return the same value. The X server
silently ignores `XSetInputFocus` with a timestamp older than the last
focus change, so focus handling could quietly stop working.

The `xlib.pxd` declarations (`XPropertyEvent.state`, `PropertyNewValue`)
ride along because the predicate cannot compile without them.

### DIFF
--- a/xpra/x11/bindings/window.pyx
+++ b/xpra/x11/bindings/window.pyx
@@ -16,7 +16,7 @@ from xpra.x11.bindings.xlib cimport (
     Expose,
     XEvent,
     XWMHints, XSizeHints,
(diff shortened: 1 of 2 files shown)


totaam:
What a great find - this would explain some of the focus issues we've had.
Thanks!

Issue #4963, opened 2026-07-17
German Keyboard altgr + q, don't work anymore


JuergenGleiss:
**Describe the bug**
Since the upgrade of 6.5.1, I can't use ALTGR+q (@) on a German keyboard. Instead of **@** a **"** is printed. If I downgrade the Xpra Server Version to xpra v6.4.4-r0, I get @. 
It doesn't matter which client version I'm using.  

**To Reproduce**
1. server command
2. client command  xpra start ssh://username@ip:/201 --audio=no --opengl=no --start=xterm
3. German keyboard -> xterm or mouspad editor --> ALTGR +Q  --> "

**System Information (please complete the following information):**
 - Server OS: Fedora Linux 44 and Debian Linux 13 (trixie)
 - Client OS: MS Windows 11
 - Xpra Server Version: xpra v6.5.1-r0
 - Xpra Client Version: xpra v6.5.1-r0 and  v.6.4.4-r0

totaam:
Not the first time we've unintentionally broken `AltGr` with the German keyboard: #3029, #3675, #4449 (long!), #2648 (old!)

I will assume that your keyboard looks something like this:
<img width="2880" height="960" alt="Image" src="https://github.com/user-attachments/assets/805e6aa1-c52d-42a8-96f8-34c60a690bfa" />

totaam:
Initial findings.

Client:
* `AltGr_modifiers(['control'], True) AltGr=, add=[], clear=['mod1', 'mod2', 'control']`
* we send a spurious `Control_L` event for `AltGr`: `do_send_keyboard('key-action', 6291468, 'Control_L', True, [], 65507, '\x00', 17, 1)`
* `AltGr` actual event: `do_send_keyboard('key-action', 6291468, 'Alt_R', True, [], 65514, '\x00', 165, 1)`
* correctly sending `@`: `do_send_keyboard('key-action', 6291468, 'at', True, [], 64, '@', 81, 1)`

Server: 
* we set the layout-group 1: `set_keyboard_layout_group(1) config=KeyboardConfig(de /  / ), self.current_keyboard_group=0`
* we unpress control immediately after pressing it: `make_keymask_match([]) current mask: {'control'}, wanted: set(), ignoring=113/['Alt_R'], keys_pressed={37: 'Control_L'}`
* correctly press: `handle keycode pressing   113: key 'Alt_R'`
[...]

PR #5009, opened 2026-08-20
keyboard: always send the `sync` value in the keymap properties

zardus:
`get_keymap_properties` only includes truthy values, so `sync=False` was
dropped and the server's `parse_options` fell back to
`boolget("sync", True)`: `--keyboard-sync=no` was ignored by the server
for every client. Set `sync` explicitly after the truthiness-filtered
loop instead of listing it with the optional properties.

This is the client-side counterpart of #4963. Old servers parse the extra key fine (`boolget`).
Behavior change: users with `keyboard-sync=no` in a config file get it honored for the first time.

Verified with `-d keyboard` on both ends: the server now logs
`sync=False` in `handle_key` and does the immediate press-and-release.

### DIFF
--- a/xpra/client/gui/keyboard_helper.py
+++ b/xpra/client/gui/keyboard_helper.py
@@ -406,7 +406,6 @@ class KeyboardHelper:
                 "backend", "backend_name",
                 "layout", "layouts", "variant", "variants", "options",
                 "raw", "layout_groups",
-                "sync",
                 "query_struct", "mod_meanings",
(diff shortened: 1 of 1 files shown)


totaam:
I prefer c32af8d600559fc96f02eb431e70f58c6b0e1ab6 as it matches the intent better.

zardus:
Sounds good. I'll rebase the PRs that depended on this.

PR #5010, opened 2026-08-20
desktop: add the missing RandR mode for any virtual display, not just dummy

zardus:
`ScreenDesktopModel.do_resize` only added the requested mode on the dummy
driver, so resizing a desktop session running on Xvfb silently did
nothing: `set_screen_size` cannot switch to a mode that does not exist,
and Xvfb only pre-defines its initial resolution. Repro: `xpra
start-desktop --resize-display=yes` on Xvfb, resize the client window —
the log shows `wanted WxH - got <initial size>` and nothing changes.

The seamless server already adds missing modes whenever
`randr_exact_size` is set (covers both dummy and Xvfb, see
`do_get_best_screen_size`) — do the same here.

### DIFF
--- a/xpra/x11/desktop/desktop_model.py
+++ b/xpra/x11/desktop/desktop_model.py
@@ -120,7 +120,13 @@ class ScreenDesktopModel(DesktopModelBase):
             if ow == rw and oh == rh:
                 return
             with xsync:
-                if RandR.is_dummy16() and (rw, rh) not in RandR.get_xrr_screen_sizes():
+                if (rw, rh) not in RandR.get_xrr_screen_sizes():
+                    # the mode must exist before it can be set: this is not
+                    # specific to the dummy driver, Xvfb needs it too
+                    # (the seamless server adds missing modes whenever
+                    # `randr_exact_size` is set, which covers both the dummy
+                    # driver and Xvfb - see `do_get_best_screen_size`
+                    # in `xpra.x11.subsystem.display`):
                     RandR.add_screen_size(rw, rh)
             with xsync:
                 if not RandR.set_screen_size(rw, rh):


totaam:
`add_screen_size`:
https://github.com/Xpra-org/xpra/blob/2852e531dd07566dc728e97ce13d8a985db7392a/xpra/x11/bindings/randr.pyx#L853-L866
Is gated on dummy 1.6 because we need randr support for `XRRAddOutputMode`:
https://github.com/Xpra-org/xpra/blob/2852e531dd07566dc728e97ce13d8a985db7392a/xpra/x11/bindings/randr.pyx#L865
I'm surprised to see that Xvfb now supports it.

I'm going to have to verify which version of Xvfb can handle this before merging, as this may need a different kind of check: perhaps we should have checked the randr extension's version all along instead.

totaam:
AI review: Single hunk in [`xpra/x11/desktop/desktop_model.py`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/desktop/desktop_model.py#L114)
(`ScreenDesktopModel.do_resize`): the `RandR.is_dummy16()` guard before
`RandR.add_screen_size()` is removed, making the mode creation unconditional.

Removed ([lines 122-124](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/desktop/desktop_model.py#L122-L124)):

```python
with xsync:
    if RandR.is_dummy16() and (rw, rh) not in RandR.get_xrr_screen_sizes():
        RandR.add_screen_size(rw, rh)
```

(Line links point at master
[`f718e273`](https://github.com/Xpra-org/xpra/commit/f718e273fd9b47be9a2fec7ec81b33f270e38388),
i.e. the state *before* this PR.)

## Findings

### 1. Guard removed instead of being replaced by the `resize_exact` check (medium/high)

[`xpra/x11/desktop/desktop_model.py:123`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/desktop/desktop_model.py#L123)

The comment justifies the change by pointing at the seamless server, but that path
only adds modes when `randr_exact_size` is true
([`display.py:438`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/subsystem/display.py#L438)),
and `randr_exact_size` is only set for dummy (randr >= 1.6), Xvfb (single
pre-defined size), or when the `_XPRA_RANDR_EXACT_SIZE` root property is present
([`display.py:248-259`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/subsystem/display.py#L248-L259)).

The model already receives that flag as `self.resize_exact` (passed from
[`window.py:195`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/desktop/window.py#L195),
stored at
[`desktop_model.py:79`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/desktop/desktop_model.py#L79),
already used at
[`desktop_model.py:164`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/desktop/desktop_model.py#L164)),
but the new code ignores it, so mode creation now also happens on displays with a
fixed mode list: a real Xorg / `--use-display=:0` session, Xephyr/Xnest, Xvnc.

Failure scenario: with `resize_exact` false, `update_size_hints()` advertises
legacy hints with a 128px increment, so a client will routinely request sizes that
are not in `get_xrr_screen_sizes()`. xpra then issues `XRRCreateMode` +
`XRRAddOutputMode`
([`randr.pyx:853`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/bindings/randr.pyx#L853))
against
[`get_current_output()`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/bindings/randr.pyx#L977)
(output 0 = the physical monitor output) and permanently attaches junk modes to the
user's real output, where previously it just resized to the closest listed size.

Suggested fix: gate on `self.resize_exact` (and/or `RandR.get_version() >= (1, 2)`),
which keeps the Xvfb fix while preserving the old behaviour elsewhere.

### 2. An X error from `add_screen_size()` now aborts the whole resize (medium)

[`xpra/x11/desktop/desktop_model.py:123`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/desktop/desktop_model.py#L123)

The call sits inside `with xsync:` with no `try/except XError`, so a failing
`XRRCreateMode`/`XRRAddOutputMode` raises `XError` at block exit, which is caught
only by the outer `except Exception` at the end of `do_resize()`
([`desktop_model.py:133-136`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/desktop/desktop_model.py#L133-L136)).
`set_screen_size()` is never attempted and the user gets
`Error: failed to resize desktop display to WxH`. Previously, non-dummy displays
skipped the call and always reached `set_screen_size()`, which picks the closest
oversized mode.

Note that `resize()` only guards on `has_randr()`
([`desktop_model.py:87`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/desktop/desktop_model.py#L87)),
which is true for RandR 1.1 servers where `XRRCreateMode` (a 1.2 request) fails, and
`get_current_output()` additionally does `assert rsc != NULL` / logs
"this display has no outputs" and returns 0
([`randr.pyx:981-985`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/bindings/randr.pyx#L981-L985)).

The seamless counterpart handles exactly this by wrapping the call in
`try/except XError`, re-querying the size list and falling back to the closest size
([`display.py:438-455`](https://github.com/Xpra-org/xpra/blob/f718e273fd9b47be9a2fec7ec81b33f270e38388/xpra/x11/subsystem/display.py#L438-L455));
the desktop path should do the same rather than let the failure kill the resize.

totaam:
The review above is mostly nonsense.
The change makes sense, we no longer support any version of Xvfb that doesn't have randr 1.2 or later.

PR #5011, opened 2026-08-20
desktop: keep the RandR monitor in sync when resizing

zardus:
On the dummy driver the vfb setup creates a RandR 1.6 monitor, and window
managers lay the desktop out from that monitor's geometry: resizing only
the screen leaves the monitor at its old size. Repro: `xpra start-desktop
--resize-display=yes` on dummy running i3 — after a resize,
`xrandr --listmonitors` still shows the old geometry and i3 tiles new
windows and its bar off-screen.

Use `set_crtc_config` for the resize on the dummy driver when the display
has a single monitor: it resizes the screen, crtc and monitor together.
Other drivers keep the plain screen-size path.

Applies on top of #5010.


totaam:
Will review after #5010 - odd that this didn't get spotted earlier!
Maybe `xpra monitor ..` mode works differently.

totaam:
Like #5010, this is totally correct!

PR #5012, opened 2026-08-20
terminal client: new backend rendering via the kitty graphics protocol

zardus:
`xpra attach --backend=terminal` renders the session into the terminal it
runs in: pixels via the kitty graphics protocol, keyboard via the kitty
keyboard protocol, pointer via SGR pixel mouse reports. Works over ssh
like any attach; needs a terminal implementing the kitty graphics
protocol (kitty, Ghostty, WezTerm, Konsole). Pure python, stdlib only,
no new dependencies; inert unless `--backend=terminal` is selected.

Trying for graceful fallback in challenging setups:
- pixel data goes through POSIX shared memory (`t=s`) when a startup
  probe shows the terminal is local, falling back to direct base64
  transmission (a full attach + 20s session is ~12KB of escape stream
  with shm, ~100KB direct);
- damaged regions are patched via shm frame edits; direct `a=f` edits
  are opt-in (`XPRA_TERMINAL_FRAME_EDITS`) because kitty 0.45 misfiles a
  spec-conformant chunked frame edit as a new animation frame (worked
  around by repeating `i`/`r` on continuation chunks; standalone
  reproducer available), so the default is flicker-free full retransmits
  using two alternating image ids per window;
- terminals without the kitty keyboard protocol fall back to legacy
  escape parsing with synthesized key releases.

The build switch defaults to off like the other alternative clients
(qt6/pyglet/tk), but the rpm and debian packages opt in with
`--with-terminal_client`, and the unit test workflow builds it so the
tests run. Second commit adds the rpm/debian packaging (without it the
rpm build fails on the unpackaged directory).

The new packages depend on `-client-gtk3` because `xpra/client/gui/` and
`xpra/client/subsystem/` are packaged there. The terminal client itself
needs no GTK, so those directories could move to the base client package
at some point, but that is a topic for another day.
Happy to vibe that in a separate PR if there's interest.

**Dependencies:** Depends on #5009 (keyboard sync-off needs `sync=False` to reach the
server); desktop sessions resize to fit the terminal, which needs
#5010/#5011 server-side.

Tested on kitty (X11 and Wayland), Ghostty and pty-emulated terminals,
POSIX only; ~380 unit tests included, passing on both build variants.

I have not looked at a single line of this code, but I've been using the new backend and it's pretty neat!


totaam:
LGTM!

I broke it with some refactoring, I will fix it.

totaam:
Fixed.

---

This is so cool. Thank you!

I launched the xpra terminal client after clearing `DISPLAY` and it just worked, with kitty graphics rendering.

I don't really know how useful this is going to be, but it's amazing that it works.

zardus:
I know, right??? Super wild to experience in the terminal :-)

totaam:
This screenshot is just so disappointing seeing how amusing it is to run an xterm in firefox in kitty on my desktop:

2026-08-27-002953_1361x1045_scrot

PR #5013, opened 2026-08-20
Fix Debian upgrades to the split Wayland package

kogeler:
## Summary

- Allow `xpra-wayland` to take ownership of Wayland files shipped by older `xpra-common` packages, preventing dpkg overwrite failures during upgrades from Xpra 6.5.3 to the 6.6 beta packages on Ubuntu 26.04.
- Add versioned `Breaks` and `Replaces` relationships while retaining the exact-version `xpra-common` dependency, so the current packages remain installed together.

The upgrade failed with:

```text
dpkg: error processing archive /tmp/apt-dpkg-install-j0TSB3/36-xpra-wayland_6.6-r42294-1_amd64.deb (--unpack):
 trying to overwrite '/usr/lib/python3/dist-packages/xpra/wayland/__init__.py', which is also in package xpra-common (6.5.3-r0-1)
```

## Test plan

- [x] Generate the binary package relationships with `dpkg-gencontrol` and verify the versioned transition fields.
- [x] Reproduce the pre-fix overwrite failure, then verify the same unpack order transfers file ownership and leaves both current packages configured.
- [x] Run the full interpreted unit-test suite: 273 successful, 0 failed, 7 requested slow-test exclusions.

### DIFF
--- a/packaging/debian/xpra/control
+++ b/packaging/debian/xpra/control
@@ -235,6 +235,8 @@ Description: tool to detach/reattach running X programs,
 
 Package: xpra-wayland
 Architecture: any
+Breaks: xpra-common (<< ${binary:Version})
+Replaces: xpra-common (<< ${binary:Version})
 Depends: xpra-common (= ${binary:Version})
         ,${misc:Depends}
         ,${python3:Depends}


totaam:
Strange how dpkg doesn't resolve that the same transaction will also remove these files from the conflicting package!

kogeler:
Yes, that surprised me too. It appears that dpkg checks file ownership one archive at a time: when `xpra-wayland` was unpacked, the installed 6.5.3 `xpra-common` still owned the path, and the newer `xpra-common` archive had not replaced its file list yet. Debian Policy documents `Breaks` plus `Replaces` for this exact package-split case: https://www.debian.org/doc/debian-policy/ch-relationships.html#overwriting-files-in-other-packages

totaam:
Very confusing indeed, in my mind ` Replaces` is what I would use if `xpra-common` was no longer needed, not "a couple of files live somewhere else now".

PR #5019, opened 2026-08-25
Serialize video pipeline cleanup with encoding

### LINKED ISSUE #5018: Video pipeline references are cleared outside the encode thread
**Describe the bug**

This is a focused follow-up to the video-pipeline lifecycle observations in
the [original PR report](https://github.com/Xpra-org/xpra/pull/5016#issuecomment-5398159908).
The cleanup ordering problem is still present on canonical Xpra
[`24c6f07`](https://github.com/Xpra-org/xpra/commit/24c6f07f41e3461b334eaad0221fa20f9f5aef08).

`video_context_clean()` reads and clears the current CSC and video encoder
references on its caller thread, then queues only the captured objects for the
encode thread to close. Encoder initialization also resets those fields
directly. This leaves an ordering window where encode-thread work can publish
or use a pipeline while another thread clears it or fails to capture it for
cleanup.

Depending on the interleaving, a partially published pipeline can survive the
queued cleanup, or its references can become unreachable before the queued
callback uses and closes them. The ordinary physical workload does not hit
this timing on every run.

**To Reproduce**

The race can be reproduced with the current `WindowVideoSource` by forcing
this ordering:

1. Pause encode-thread pipeline setup immediately before the new CSC or video
   encoder reference is published.
2. Call `video_context_clean()`, encoder reinitialization, or full source
   cleanup from the caller thread.
3. Resume the encode worker and let the queued cleanup callback run.
4. Inspect the active references and codec cleanup calls.

If publication happens after the caller-thread snapshot, the queued callback
does not own the newly published object. If publication happens before that
snapshot, the caller thread clears the shared field before the FIFO callback
has reached it. Full cleanup and encoder reinitialization expose the same
ordering boundary.

- Server OS: Ubuntu 26.04, Wayland seamless backend, RADV on
  `/dev/dri/renderD128`
- Client OS: Debian 13, GTK3 under Sway/Xwayland for the physical integration
  run; no client is required to exercise the server-side ordering
- Xpra Server Version: `v6.6-r42464 (g24c6f07f41e) beta`
- Xpra Client Version: `v6.6-r42464 (g24c6f07f41e) beta`

**Additional context**

The current code already states that codec cleanup must run on the encode
thread, but the shared references are read and cleared before work is queued:
[`video_compress.py`](https://github.com/Xpra-org/xpra/blob/24c6f07f41e3461b334eaad0221fa20f9f5aef08/xpra/server/window/video_compress.py#L369-L393).
The encode thread should own the point where active pipeline references are
read, cleared, and closed, with cleanup, reinitialization, and shutdown ordered
against other encode-thread work.

kogeler:
Closes https://github.com/Xpra-org/xpra/issues/5018

## Problem

The cleanup changes now on master close the original empty-snapshot race and
cancel pending video timers and flushes synchronously. Three reachable edges
remain: encoder reinitialization drops an attached codec pair, full cleanup and
suspension queue a duplicate cleanup barrier, and a CSC cleanup exception can
prevent both encoder cleanup and the late-pipeline sweep.

## Change

Build on the current cleanup implementation by requesting video-context
cleanup before a real encoder reinitialization, relying on the inherited
`cancel_damage()` path for one full-cleanup or suspension barrier, and using
nested `finally` blocks for the captured codecs and late sweep.

Codec selection and the upstream timer, flush, and direct encode-thread policy
are unchanged.

## Validation

The final patch was tested on canonical Xpra
[`1d8e112`](https://github.com/Xpra-org/xpra/commit/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b),
which includes the maintainer's `d1ca74f` and `aae59ec` cleanup commits.

The test-only upstream verification failed all three residual production-path
checks before this patch. With the patch, the atomic cleanup module passed
8/8, the combined focused selection passed 29/29, and the native Wayland gate
passed 15/15.

Baseline, Cython-expanded, and no-backwards-compatibility full runs each
completed 292 modules successfully. Their sole failed module was the unrelated
upstream `unit.client.x11_client_paint_test`, also the sole failure in
[upstream's three-mode CI run](https://github.com/Xpra-org/xpra/actions/runs/32964054292)
on the exact unmodified master; no X11 path was touched or bypassed.

The physical H.264 scenario passed with 18 H.264-only updates, no startup or
pre-negotiation pipeline error, a contiguous 16-frame VA-API
`EncSlice`/`VLD` stream presented as NV12 through native OpenGL, matching
pixels, pointer interaction, lifecycle, and cleanup.

<details>
<summary>Exact named test runs</summary>

- negative upstream-only verification: `current-1d8-cleanup-verification-02`
- atomic focused: `current-1d8-cleanup-residual-focused-02`
- combined focused: `current-1d8-residuals-focused-01`
- native Wayland: `current-1d8-residuals-wayland-01`
- full modes: `current-1d8-residuals-full-01`,
  `current-1d8-residuals-full-cython-01`, and
  `current-1d8-residuals-full-no-compat-01`
- physical H.264: `current-1d8-residuals-live-h264-01`

</details>

### DIFF
--- a/xpra/server/window/video_compress.py
+++ b/xpra/server/window/video_compress.py
@@ -276,6 +276,11 @@ class WindowVideoSource(WindowSource):
         if "scroll" in self.server_core_encodings:
             add("scroll", self.scroll_encode)
 
+    def init_encoders(self) -> None:
+        if hasattr(self, "_encoders"):
+            self.video_context_clean()
+        super().init_encoders()
+
     def do_set_auto_refresh_delay(self, min_delay, delay) -> None:
         super().do_set_auto_refresh_delay(min_delay, delay)
         if r := self.video_subregion:
@@ -357,15 +362,6 @@ class WindowVideoSource(WindowSource):
             "src_format"    : src_format
         }
 
-    def suspend(self) -> None:
-        super().suspend()
-        # we'll create a new video pipeline when resumed:
-        self.video_context_clean()
-
-    def cleanup(self) -> None:
-        super().cleanup()
-        self.video_context_clean()
-
     def video_context_clean(self, encode_thread: bool = False) -> None:
         """Detach the video context and clean it from the encode thread."""
         self.cancel_video_encoder_flush()
@@ -381,16 +377,20 @@ class WindowVideoSource(WindowSource):
         def clean() -> None:
             if DEBUG_VIDEO_CLEAN:
                 log.warn("video_context_clean() done")
-            if csce:
-                self.csc_clean(csce)
-            if ve:
-                self.ve_clean(ve)
-            # this function always runs from the encode thread
-            # but the `video_context_clean` may have been called from another thread,
-            # in which case we want to run it again to close video contexts
-            # that may have been instantiated in the meantime:
-            if not encode_thread:
-                self.video_context_clean(True)
+            try:
+                try:
+                    if csce:
+                        self.csc_clean(csce)
+                finally:
+                    if ve:
+                        self.ve_clean(ve)
+            finally:
+                # this function always runs from the encode thread
+                # but the `video_context_clean` may have been called from another thread,
+                # in which case we want to run it again to close video contexts
+                # that may have been instantiated in the meantime:
+                if not encode_thread:
+                    self.video_context_clean(True)
 
         if encode_thread:
             # already in the correct thread
--- a/tests/unittests/unit/server/window/video_compress_test.py
+++ b/tests/unittests/unit/server/window/video_compress_test.py
@@ -4,9 +4,93 @@
 # later version. See the file COPYING for details.
 
 import unittest
-from unittest.mock import Mock
+from queue import Queue
+from threading import Event, Thread
+from types import SimpleNamespace
+from unittest.mock import Mock, patch
 
+from xpra.server.window import compress, video_compress
 from xpra.server.window.video_compress import WindowVideoSource
+from xpra.util.objects import typedict
+
+
+class PipelineElement:
+
+    def __init__(self, clean_error: bool = False) -> None:
+        self.clean_count = 0
+        self.clean_error = clean_error
+
+    def clean(self) -> None:
+        self.clean_count += 1
+        if self.clean_error:
+            raise RuntimeError("cleanup failed")
+
+
+class Converter(PipelineElement):
+
+    def init_context(self, src_width, src_height, _src_format,
+                     dst_width, dst_height, _dst_format, _options) -> None:
+        self.src_size = src_width, src_height
+        self.dst_size = dst_width, dst_height
+
+    def get_info(self) -> dict:
+        return {}
+
+
+class Encoder(PipelineElement):
+
+    def __init__(self, init_started=None, finish_init=None) -> None:
+        super().__init__()
+        self.init_started = init_started
+        self.finish_init = finish_init
+
+    def init_context(self, _encoding, _width, _height, _src_format, _options) -> None:
+        if self.init_started:
+            self.init_started.set()
+        if self.finish_init and not self.finish_init.wait(2):
+            raise RuntimeError("encoder setup was not released")
+
+    def get_info(self) -> dict:
+        return {}
+
+
+class EncodeWorker:
+
+    def __init__(self) -> None:
+        self.queue: Queue = Queue()
+        self.optional: list[bool] = []
+        self.callbacks: list = []
+        self.errors: list[BaseException] = []
+        self.thread = Thread(target=self.run, name="test-encode", daemon=True)
+        self.thread.start()
+
+    def call(self, optional: bool, callback, *args) -> None:
+        self.optional.append(optional)
+        self.callbacks.append(callback)
+        self.queue.put((callback, args))
+
+    def run(self) -> None:
+        while True:
+            item = self.queue.get()
+            try:
+                if item is None:
+                    return
+                callback, args = item
+                callback(*args)
+            except BaseException as e:
+                self.errors.append(e)
+            finally:
+                self.queue.task_done()
+
+    def drain(self) -> None:
+        self.queue.join()
+
+    def close(self) -> None:
+        self.queue.put(None)
+        self.queue.join()
+        self.thread.join(2)
+        if self.thread.is_alive():
+            raise RuntimeError("encode worker did not stop")
 
 
 class VideoContextCleanTest(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


xpra/server/window/video_compress.py
```diff
@@ -185,6 +185,8 @@ class WindowVideoSource(WindowSource):
 
     def __init__(self, *args):
         self.supports_scrolling: bool = False
+        self._csc_encoder: ColorspaceConverter | None = None
+        self._video_encoder: VideoEncoder | None = None
```
totaam: Why the move?

xpra/server/window/video_compress.py
```diff
@@ -228,14 +230,12 @@ def init_vars(self) -> None:
         self.scroll_data = None
         self.last_scroll_time = 0.0
 
-        self._csc_encoder: ColorspaceConverter | None = None
-        self._video_encoder: VideoEncoder | None = None
         self._last_pipeline_check = 0
 
     def do_init_encoders(self) -> None:
+        if hasattr(self, "_encoders"):
+            self.cleanup_codecs()
```
totaam: Why is this needed?

xpra/server/window/video_compress.py
```diff
@@ -373,24 +373,28 @@ def cleanup_codecs(self) -> None:
             (the encoder and csc module may be in use by that thread)
         """
         self.cancel_video_encoder_flush()
+        self.cancel_video_encoder_timer()
```
totaam, line 368: This looks valid, and may need separating out.

xpra/server/window/video_compress.py
```diff
@@ -373,24 +373,28 @@ def cleanup_codecs(self) -> None:
...
-                    log.warn("video_context_clean() done")
+            try:
                 self.csc_clean(csce)
+            finally:
                 self.ve_clean(ve)
-            self.call_in_encode_thread(False, clean)
+            if DEBUG_VIDEO_CLEAN:
+                log.warn("video_context_clean() done")
+
+        self.call_in_encode_thread(False, clean)
```
totaam: This may look like a good solution, but I don't think that it is.
We don't want these csc and encoders to be used any more, so we clear them early.
We have to do the actual clean from the encode thread, but clearing the reference early is what we want.

totaam (commented):
The race is tiny, can you actually trigger it?
If not, then I don't think we should be changing this.

kogeler:
Thanks. I can reproduce the race deterministically, and I have reworked the patch around all of the review points.

The codec references remain in `init_vars()` and are still cleared immediately when cleanup is requested. Reinitialization requests cleanup only when encoder tables already exist, so initial construction adds no barrier. The encoder timer is cancelled synchronously because it participates in the same teardown boundary. The redundant `suspend()` and `cleanup()` overrides are gone.

Cleanup now always queues one non-optional FIFO sweep. It closes the pair captured and cleared by the caller as well as any identity-distinct pair published by setup work that was already ahead of the cleanup request. This preserves early invalidation without losing the in-flight pipeline.

<details>
<summary>Deterministic trigger and ordering proof</summary>

On [`521db02`](https://github.com/Xpra-org/xpra/commit/521db02f143d9f0a57c656658d9ee12572622276), the regression uses the real `WindowVideoSource.setup_pipeline_option()` path and pauses the encode worker inside encoder initialization. Cleanup is then requested from the caller thread before setup is allowed to finish.

With the current code and no existing pipeline, cleanup sees no published references and queues nothing. Setup resumes and publishes a CSC converter and video encoder that remain active and unclosed. With an existing pipeline, the callback closes only the captured old pair and misses the new pair published by the setup task that was already running.

The revised ordering clears the visible pair immediately and places a mandatory sweep after the already-running setup task. That sweep clears and closes the late pair. Work already ahead of the cleanup request finishes first; later work sees cleared references.

</details>

<details>
<summary>Reinitialization and timer coverage</summary>

`EncodingServer.reinit_encodings()` reaches `init_encoders()` on each live window source through `WindowsConnection.reinit_encoders()`. The old `do_init_encoders()` replaced live codec references with `None` without closing the old pipeline. The revised guard requests cleanup before the base method rebuilds the encoder tables, while leaving initial construction unchanged.

The regression calls the complete `init_encoders()` wrapper and verifies its CSC parsing and encoding-selection calls. The full-cleanup case uses the production timer-scheduling path, blocks the encode worker, requests cleanup, and verifies that the pending encoder timer is cancelled before the worker is released.

</details>

<details>

The revised change was validated together with the two Wayland follow-ups on `521db02`. The combined focused suite passed 24/24 tests, and the native Wayland suite passed 15/15.

The physical-GPU scenario passed both strict profiles. RGB delivered 11 RGB-only updates. H.264 delivered 18 H.264-only updates with no startup or pre-negotiation errors, including a 16-frame hardware stream with VA-API `EncSlice` and `VLD` plus native NV12 OpenGL rendering. Both runs rendered the expected pixels, accepted real input, and completed the expected lifecycle cleanup.

</details>

xpra/server/window/video_compress.py
kogeler: fixed

xpra/server/window/video_compress.py
kogeler: fixed

xpra/server/window/video_compress.py
kogeler: fixed

xpra/server/window/video_compress.py
kogeler: fixed

totaam:
I much prefer d1ca74f49d16ffb37e3f5a7b462967a1cc3fbc57 + aae59ecc0792b50eed07ef43d730d51e683e9a84

kogeler:
> I much prefer [d1ca74f](https://github.com/Xpra-org/xpra/commit/d1ca74f49d16ffb37e3f5a7b462967a1cc3fbc57) + [aae59ec](https://github.com/Xpra-org/xpra/commit/aae59ecc0792b50eed07ef43d730d51e683e9a84)

These commits are not fully cover my case that I'm fixing here. I will prepare new commit for this PR.

kogeler:
@totaam 

Thank you. I rebased the PR on the current cleanup implementation and kept the
parts from `d1ca74f` and `aae59ec` intact. Those commits do close the empty
snapshot race, cancel timers and flushes synchronously, and add the mandatory
late encode-thread sweep.

Before reducing the PR, I ran a test-only verification against unmodified
[`1d8e112`](https://github.com/Xpra-org/xpra/commit/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b), which contains both commits. It still fails three reachable lifecycle checks.
The updated PR now contains only the corresponding residual changes: cleanup
before live encoder reinitialization, removal of duplicate cleanup requests
from the video `cleanup()` and `suspend()` overrides, and exception-safe codec
cleanup plus late sweep.

<details>
<summary>Why these paths are reachable</summary>

`EncodingServer.reinit_encodings()` calls
`WindowsConnection.reinit_encoders()` after threaded codec discovery and on a
new encoding configuration. That method calls `init_encoders()` on every live
window source. On current master, `WindowVideoSource.do_init_encoders()` sets
the attached CSC and video-encoder references to `None` without first closing
the pair.

For shutdown and suspension, the base methods already call the virtual
`cancel_damage()`. The video override of `cancel_damage()` requests
`video_context_clean()`, and the video `cleanup()` / `suspend()` overrides
request it a second time after the base call. Since current master now always
queues a sweep, this produces two cleanup barriers for one operation.

Finally, the queued callback calls CSC cleanup, then encoder cleanup, then the
late sweep sequentially. If CSC cleanup raises, Python never reaches either of
the following operations. Nested `finally` blocks preserve all three cleanup
obligations.

</details>

<details>
<summary>Negative verification on the two upstream commits</summary>

The verification adds no production change and drives the real
`setup_pipeline_option()`, complete `init_encoders()`, and inherited
`cleanup()` paths on a FIFO worker.

On unmodified `1d8e112`, all three tests fail:

- reinitialization enters base encoder initialization while the old CSC is
  still attached;
- a captured CSC cleanup exception leaves the late-published pipeline
  attached and prevents its encoder cleanup;
- one full cleanup queues six callbacks instead of the expected five because
  it requests two cleanup barriers.

```text
Ran 3 tests in 0.004s
FAILED (failures=3)
```

The retained complete-log SHA-256 is
`990db481ed2235836f3db258c516c66383a5a4579e9b0bca21f11f6da8950d7b`.

</details>

<details>
<summary>Validation of the reduced PR</summary>

- atomic cleanup module: 8/8;
- combined focused selection: 29/29;
- native Wayland gate: 15/15 plus build/import/linkage preflight;
- each full mode: 292 successful modules and the sole unrelated
  `unit.client.x11_client_paint_test` failure already present in
  [upstream's exact-master three-mode run](https://github.com/Xpra-org/xpra/actions/runs/32964054292);
- physical H.264: 18 H.264-only updates, no startup or pre-negotiation error,
  a contiguous 16-frame VA-API `EncSlice`/`VLD` stream, native NV12 OpenGL,
  pixels, input, lifecycle, and cleanup.

No X11 test, configuration, skip, codec-selection rule, or maintainer timer,
flush, direct-thread, or late-sweep behavior is changed by this revision.

</details>

kogeler:
Thank you for your time and for all the discussion around these PRs.

For context, I validate each of these fixes against a real set of live end-to-end tests, covering actual data transfer as well as encoding and decoding on real hardware. I’ve already automated this as part of a CI/CD pipeline that continuously exercises the full flow, so the changes I proposed are driven by issues I can reproduce in my specific environment.

That said, I think that by trying to adapt the upstream project to fit my rather narrow use case, I’m starting to spend both your time and mine on requirements that may simply not belong in the upstream project.

So rather than continuing the discussion, I’ll close these PRs and maintain a fork instead. I’ll keep it in sync with upstream while carrying the small set of patches needed for my particular use case.

Thanks again for your help, for the explanations that helped me better understand the project, and especially for the time you’ve spent reviewing and discussing my changes. I really appreciate it.

Issue #5020, opened 2026-08-25
Wayland configure packets discard one-shot client properties


kogeler:
**Describe the bug**

This is a narrower follow-up to the Wayland negotiation problem in the
[original PR report](https://github.com/Xpra-org/xpra/pull/5016#issuecomment-5398159908).
The old RGB format failure is fixed, but the Wayland server still discards the
backing-specific properties carried by a configure packet.

I reproduced this on canonical Xpra
[`24c6f07`](https://github.com/Xpra-org/xpra/commit/24c6f07f41e3461b334eaad0221fa20f9f5aef08).
The GTK client places its pending backing properties in
`config["properties"]` and then clears its local dictionary. The X11 configure
handler applies that dictionary before processing geometry. The Wayland
handler processes geometry but never reads the properties, so the window
source does not receive the RGB, CSC, transparency, or render-size values from
that configure event.

Current RGB rendering still succeeds because the newer default RGB formats
mask the old visible symptom. The broken operation is the configure update
itself: the client sends one-shot state and the Wayland window source drops it.

**To Reproduce**

1. Build the server and client from `24c6f07` without local changes.
2. Start the X11 desktop used by the client.
3. Start a Wayland seamless server with Zed and attach the GTK3 client in
   RGB/Cairo mode.
4. Inspect the server trace during the initial configure and map sequence.
5. Observe that no configure properties reach `do_set_client_properties()`;
   only the later map event does.

<details>
<summary>Server and client commands</summary>

The client desktop was started with:

```bash
Xvfb :0 -screen 0 1600x1200x24 +extension Composite -nolisten tcp -noreset
DISPLAY=:0 openbox --sm-disable
DISPLAY=:0 picom --backend xrender --config /dev/null --log-level debug
DISPLAY=:0 xsetroot -solid '#000000'
```

The server ran:

```bash
env \
  XDG_RUNTIME_DIR=/tmp/server-runtime \
  XPRA_SCREEN_UPDATES_DIRECTORY=/artifacts/screen-updates \
  xpra seamless :150 \
    --minimal \
    --backend=wayland \
    --daemon=no \
    --displayfd=1 \
    --bind-tcp=0.0.0.0:14500,auth=none \
    --socket-dir=/tmp/server-runtime/xpra-sockets \
    --socket-dirs=/tmp/server-runtime/xpra-sockets \
    --sessions-dir=/tmp/server-runtime/xpra-sessions \
    --session-name=zed-rgb-lab \
[...]

PR #5021, opened 2026-08-25
Test Wayland configure client properties

kogeler:
Follow-up to https://github.com/Xpra-org/xpra/issues/5020 and the fix in
https://github.com/Xpra-org/xpra/commit/e1aaa640d5f49e1f1f82b896a8a94fc698f347ec.

## Problem

The configure-property fix added in `e1aaa640` has no focused regression test.

## Change

Add regression coverage for configure client-property forwarding on the
Wayland server. The tests cover property-only updates, verify that properties
are applied before geometry, and keep the missing-window and missing-surface
paths as no-ops.

This PR changes tests only.

## Validation

The three current changes were validated together on canonical Xpra
[`521db02`](https://github.com/Xpra-org/xpra/commit/521db02f143d9f0a57c656658d9ee12572622276).
The focused regressions passed 24/24 tests, and the native Wayland gate passed
15/15 tests plus its build, import, and linkage checks.

In each full mode—baseline, Cython-expanded, and no backward compatibility—
all 292 other test modules and every module changed by this patch set passed.
The sole failing module was the new upstream
`unit.client.x11_client_paint_test`; it fails on the unmodified commit and in
all three jobs of [upstream's own Actions run](https://github.com/Xpra-org/xpra/actions/runs/32871771151),
so it is not caused by these changes.

In the physical Zed/RADV workload, unmodified master passed RGB but reproduced
the strict H.264 startup error before map properties arrived. With the three
changes applied, RGB passed with 11 RGB-only updates, and H.264 passed with 18
H.264-only updates, no pre-negotiation pipeline error, and a contiguous
16-frame VA `EncSlice`/`VLD` stream presented as NV12 through AMD OpenGL.
Pixels, pointer input, lifecycle, and cleanup passed in both patched runs.

<details>
<summary>Focused regression command</summary>

```bash
CFLAGS='-O0 -g0' CXXFLAGS='-O0 -g0' \
EXTRA_ARGS='--with-terminal_client --with-wayland_server' \
python3 setup.py unittests unit/wayland/window_test.py
```

</details>

<details>
<summary>Wayland boundary command</summary>

```bash
CFLAGS='-O0 -g0' CXXFLAGS='-O0 -g0' \
EXTRA_ARGS='--minimal --with-modules --with-server --with-wayland_server' \
python3 setup.py unittests unit/wayland
```

</details>

### DIFF
--- a/tests/unittests/unit/wayland/window_test.py
+++ b/tests/unittests/unit/wayland/window_test.py
@@ -9,6 +9,8 @@ import unittest
 from types import ModuleType
 from unittest.mock import Mock, patch
 
+from xpra.util.objects import typedict
+
 
 def load_window_server_class():
     modules = {}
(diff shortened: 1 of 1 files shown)


totaam:
Sorry, I had already committed the fix in e1aaa640d5f49e1f1f82b896a8a94fc698f347ec

Please submit the tests alone, or close the PR.

kogeler:
Thanks, you are right: [`e1aaa640`](https://github.com/Xpra-org/xpra/commit/e1aaa640d5f49e1f1f82b896a8a94fc698f347ec) already fixes the production path and is present in current master. I have reduced this PR to tests only.

The tests cover a property-only configure, verify that properties are applied before resize, `frame_done`, and compositor flush, and preserve the no-op behavior when either the window or surface is missing.

<details>

The tests-only revision was validated together with the two related fixes on [`521db02`](https://github.com/Xpra-org/xpra/commit/521db02f143d9f0a57c656658d9ee12572622276). The combined focused suite passed 24/24 tests, and the native Wayland suite passed 15/15.

The combined physical-GPU scenario passed both strict profiles. RGB delivered 11 RGB-only updates. H.264 delivered 18 H.264-only updates with no startup or pre-negotiation errors, including a 16-frame hardware stream with VA-API `EncSlice` and `VLD` plus native NV12 OpenGL rendering. Both runs rendered the expected pixels, accepted real input, and completed the expected lifecycle cleanup.

</details>

PR #5023, opened 2026-08-25
Wait for valid Wayland initial window state

### LINKED ISSUE #5022: Wayland initial windows can be announced or damaged before they are ready
**Describe the bug**

This is a follow-up to the Wayland seamless case in the
[original PR report](https://github.com/Xpra-org/xpra/pull/5016#issuecomment-5398159908).
Most of the codec problems from that report are fixed, but initial Wayland
windows can still be announced or damaged before they are ready.

I reproduced this on canonical Xpra
[`24c6f07`](https://github.com/Xpra-org/xpra/commit/24c6f07f41e3461b334eaad0221fa20f9f5aef08).
There are two observable outcomes at the same startup boundary:

- If the surface is still 0x0, the GTK client creates a 1x1 window and later
  rejects a second create packet for the same window ID when the real size
  becomes available.
- If the surface already has a valid size, its first H.264 damage can reach the
  encoder before the client's map packet supplies the per-window CSC modes.
  The server raises a pipeline-selection exception and then recovers after map.

The 0x0 timing is intermittent. The H.264 startup failure was retained in the
current clean-master run.

**To Reproduce**

1. Build the server and client from `24c6f07` without local changes.
2. Start a Wayland seamless server with Zed as its child.
3. Attach the GTK3 client with strict hardware H.264 settings.
4. Observe the first damage before the client's map properties arrive.
5. Repeating the same early attach can also expose the 0x0 duplicate-create
   path.

<details>
<summary>Server and client commands</summary>

The server ran:

```bash
env \
  XDG_RUNTIME_DIR=/tmp/server-runtime \
  XPRA_SCREEN_UPDATES_DIRECTORY=/artifacts/screen-updates \
  LIBVA_DRIVER_NAME=radeonsi \
  XPRA_LIBVA_DEVICE=/dev/dri/renderD128 \
  LIBVA_TRACE=/artifacts/server-va \
  xpra seamless :150 \
    --minimal \
    --backend=wayland \
    --daemon=no \
    --displayfd=1 \
    --bind-tcp=0.0.0.0:14500,auth=none \
    --socket-dir=/tmp/server-runtime/xpra-sockets \
    --socket-dirs=/tmp/server-runtime/xpra-sockets \
    --sessions-dir=/tmp/server-runtime/xpra-sessions \
    --session-name=zed-h264-lab \
    --use-display=no \
    --exit-with-client=no \
    --exit-with-children=yes \
    --terminate-children=yes \
    --start-child=/opt/start_zed.sh \
    --video-scaling=0 \
    --auto-refresh-delay=0 \
    --html=off \
    --video=yes \
    --encodings=h264 \
    --video-encoders=libva \
    --csc-modules=libyuv \
    -d wayland,damage,encoding,encoder,argb
```

The client ran under Sway/Xwayland with native AMD GLX:

```bash
env \
  XDG_RUNTIME_DIR=/tmp/client-runtime \
  DISPLAY=:0 \
  GDK_BACKEND=x11 \
  PYOPENGL_PLATFORM=x11 \
  LIBVA_DRIVER_NAME=radeonsi \
  XPRA_LIBVA_DEVICE=/dev/dri/renderD128 \
  LIBVA_TRACE=/artifacts/client-va \
  xpra attach tcp://xpra-server:14500 \
    --minimal \
    --compressors=none \
    --quality=100 \
    --speed=100 \
    --reconnect=no \
    --video=yes \
    --encodings=h264 \
    --opengl=force:native \
    --video-decoders=libva \
    --csc-modules=none \
    -d draw,paint,cairo,window,gtk,alpha,libva
```

</details>

<details>
<summary>Observed clean-master output</summary>

The retained 0x0 run failed on the duplicate create:

```text
ValueError: we already have a window 0x1: ClientWindow(0x1)
```

The clean H.264 run failed before the map properties arrived:

```text
Error: no video pipeline options found
 h264 encoding with source format 'RGBX'
 1536x1095 32-bit
 all encoders: libva
 client supported CSC modes for 'h264':
ValueError: no options to choose from
Error: failed to create data packet
```

The map packet carrying `encoding.full_csc_modes` arrived about 28 ms later.
The session recovered and rendered, but the startup exception had already
occurred.

</details>

- Server OS: Ubuntu 26.04, Wayland seamless backend, RADV on
  `/dev/dri/renderD128`
- Client OS: Debian 13, GTK3 under Sway/Xwayland, native AMD GLX
- Xpra Server Version: `v6.6-r42464 (g24c6f07f41e) beta`
- Xpra Client Version: `v6.6-r42464 (g24c6f07f41e) beta`

**Additional context**

A Wayland window should not be announced with non-positive dimensions, and
its first damage should not be encoded before the attaching client has
supplied the window's backing-specific encoding state. Other window backends
should keep their existing initial-window behavior.

kogeler:
Closes https://github.com/Xpra-org/xpra/issues/5022

## Problem

When a client attaches to a Wayland session, the initial-window pass can emit
state before the registered model is ready in either of two ways. A `0x0`
model can be announced before `update_size()` performs its normal first valid
announcement, or a valid-size video-only window can be damaged before the map
packet supplies its concrete backing's CSC modes. The latter enters H.264
pipeline selection with no usable mode and no picture fallback.

## Change

Leave an exact `0x0` Wayland model for the existing `update_size()` transition
to announce. For valid windows, delay damage only while a normal top-level,
video-only source has no fallback and its per-window CSC state is unresolved.
A map event or explicit per-window CSC dictionary releases damage. RGB,
picture fallback, usable global CSC, OR, tray, shadow, and subsurface paths
remain immediate.

## Validation

The atomic change was validated on canonical Xpra
[`1d8e112`](https://github.com/Xpra-org/xpra/commit/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b).
The same test-only verification failed all five new production-path assertions
on unmodified master, while the atomic patch passed 21/21 focused tests. The
native Wayland gate passed its build, import, and linkage checks plus 15/15
tests.

Each full mode completed 292 modules successfully and reported one failure,
the unrelated `unit.client.x11_client_paint_test`. Canonical
[Actions run 32964054292](https://github.com/Xpra-org/xpra/actions/runs/32964054292)
on exact unmodified `1d8e112` has the same sole failed module and seven skips
in baseline, Cython-expanded, and no-backwards-compatibility jobs. This change
neither touches nor skips that test.

On the physical Zed/RADV workload, strict H.264 on unmodified master reproduced
the startup error before map properties arrived. Both proposed picture
fallback configurations made the application visible, but carried every
packet as `rgb32`: `--encodings=h264,rgb` produced 20 RGB and zero H.264
updates, while adding `--encoding=h264` produced seven RGB and zero H.264
updates. The patched strict run produced 19 H.264 and zero RGB updates with no
startup or pre-negotiation error. Hardware VA-API encode/decode, native NV12
OpenGL, pixels, pointer input, application capture, lifecycle, and cleanup all
passed. The patched RGB control also passed with 11 RGB-only updates.

<details>
<summary>Focused regression command</summary>

```bash
CFLAGS='-O0 -g0' CXXFLAGS='-O0 -g0' \
EXTRA_ARGS='--with-terminal_client' \
python3 setup.py unittests \
  unit/server/subsystem/window_test.py \
  unit/server/window/initial_damage_test.py \
  unit/wayland/window_test.py
```

</details>

### DIFF
--- a/xpra/server/subsystem/window.py
+++ b/xpra/server/subsystem/window.py
@@ -634,9 +634,15 @@ class WindowServer(StubSubsystem):
                     window.move_resize(-200, -200, w, h)
             else:
                 # code more or less duplicated from _send_new_window_packet:
+                x, y, w, h = window.get_property("geometry")
+                # The Wayland backend registers a 0x0 model before its first
+                # valid commit. Its update_size() path sends the create packet
+                # when that model becomes ready.
+                if w == h == 0:
+                    log("not sending initial window %#x with zero dimensions", wid)
+                    continue
                 if not sharing and not window.is_OR():
                     window.hide()
-                x, y, w, h = window.get_property("geometry")
                 wprops = self.client_properties.get(wid, {}).get(ss.uuid, {})
                 packet_type = "new-override-redirect" if (window.is_OR() and BACKWARDS_COMPATIBLE) else WINDOW_CREATE
                 ss.new_window(packet_type, wid, window, x, y, w, h, wprops)
--- a/xpra/server/window/video_compress.py
+++ b/xpra/server/window/video_compress.py
@@ -231,6 +231,7 @@ class WindowVideoSource(WindowSource):
         self._csc_encoder: ColorspaceConverter | None = None
         self._video_encoder: VideoEncoder | None = None
         self._last_pipeline_check = 0
+        self._client_csc_modes_resolved = False
 
     def do_init_encoders(self) -> None:
         super().do_init_encoders()
@@ -280,6 +281,27 @@ class WindowVideoSource(WindowSource):
         super().do_set_auto_refresh_delay(min_delay, delay)
         if r := self.video_subregion:
             r.set_auto_refresh_delay(self.base_auto_refresh_delay)
+
+    def set_client_properties(self, properties: typedict) -> None:
+        if properties.strget("event") == "map" or isinstance(properties.get("encoding.full_csc_modes"), dict):
+            self._client_csc_modes_resolved = True
+        super().set_client_properties(properties)
+
+    def waiting_for_client_csc_modes(self) -> bool:
+        return all((
+            not self._client_csc_modes_resolved,
+            not getattr(self, "parent_wid", 0),
+            not (self.is_OR or self.is_tray or self.is_shadow),
+            bool(self.common_video_encodings),
+            not self.non_video_encodings,
+            not any(self.full_csc_modes.strtupleget(encoding) for encoding in self.common_video_encodings),
+        ))
+
+    def damage(self, x: int, y: int, w: int, h: int, options=None) -> None:
+        if self.waiting_for_client_csc_modes():
+            videolog("dropping damage for window %#x until client CSC modes are known", self.wid)
+            return
+        super().damage(x, y, w, h, options)
 
     def update_av_sync_frame_delay(self) -> None:
         self.av_sync_frame_delay = 0
--- a/tests/unittests/unit/server/subsystem/window_test.py
+++ b/tests/unittests/unit/server/subsystem/window_test.py
@@ -8,12 +8,47 @@ import unittest
 from unittest.mock import Mock, patch
 
 from xpra.net.common import Packet
+from xpra.net.packet_type import WINDOW_CREATE
 from xpra.util.objects import AdHocStruct, typedict
 from unit.test_util import stubbable
 from unit.server.subsystem.servermixintest_util import ServerMixinTest
 
 
 class WebcamMixinTest(ServerMixinTest):
+
+    @staticmethod
+    def make_initial_window_server(geometry):
+        from xpra.server.subsystem.window import WindowServer
+
+        server = object.__new__(WindowServer)
+        window = Mock()
+        window.is_managed.return_value = True
+        window.is_tray.return_value = False
+        window.is_OR.return_value = False
+        window.get_property.return_value = geometry
+        server._id_to_window = {1: window}
+        server.client_properties = {}
+        return server, window
+
+    def test_initial_window_waits_for_nonzero_dimensions(self):
+        server, window = self.make_initial_window_server((10, 20, 0, 0))
+        source = Mock(uuid="client")
+
+        server.send_initial_windows(source)
+
+        window.hide.assert_not_called()
+        source.new_window.assert_not_called()
+        source.damage.assert_not_called()
+
+        server, window = self.make_initial_window_server((10, 20, 800, 600))
+        source = Mock(uuid="client")
+
+        server.send_initial_windows(source)
+
+        source.new_window.assert_called_once_with(
+            WINDOW_CREATE, 1, window, 10, 20, 800, 600, {},
+        )
+        source.damage.assert_called_once_with(1, window, 0, 0, 800, 600)
 
     def test_window_stacking_packet(self):
         from xpra.server.subsystem.window import WindowServer
--- a/tests/unittests/unit/server/window/initial_damage_test.py
+++ b/tests/unittests/unit/server/window/initial_damage_test.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+# This file is part of Xpra.
+# Copyright (C) 2026 Antoine Martin <antoine@xpra.org>
+# Xpra is released under the terms of the GNU GPL v2, or, at your option, any
+# later version. See the file COPYING for details.
+
+import unittest
+from unittest.mock import patch
+
+from xpra.server.window.compress import WindowSource
+from xpra.server.window.subsurface_source import SubsurfaceWindowSource
+from xpra.server.window.video_compress import WindowVideoSource
+from xpra.util.objects import typedict
+
+
+class InitialDamageTest(unittest.TestCase):
+
+    @staticmethod
+    def make_source(source_class: type[WindowVideoSource] = WindowVideoSource) -> WindowVideoSource:
+        source = object.__new__(source_class)
+        source.wid = 1
+        source._client_csc_modes_resolved = False
+        source.is_OR = False
+        source.is_tray = False
+        source.is_shadow = False
+        source.common_video_encodings = ("h264",)
+        source.non_video_encodings = ()
+        source.full_csc_modes = typedict()
+        return source
+
+    @staticmethod
+    def damage(source: WindowVideoSource) -> bool:
+        with patch.object(WindowSource, "damage") as damage:
+            source.damage(0, 0, 800, 600)
+        return damage.called
+
+    def set_properties(self, source: WindowVideoSource, properties: dict) -> None:
+        with patch.object(WindowSource, "set_client_properties") as set_properties:
+            source.set_client_properties(typedict(properties))
+        set_properties.assert_called_once()
+
+    def test_video_only_damage_waits_for_window_csc_modes(self):
+        source = self.make_source()
+
+        self.assertFalse(self.damage(source))
+
+        self.set_properties(source, {"event": "map"})
+        self.assertTrue(self.damage(source))
+
+    def test_explicit_empty_window_csc_modes_resolve_state(self):
+        source = self.make_source()
+
+        self.set_properties(source, {"encoding.full_csc_modes": {}})
+
+        self.assertTrue(source._client_csc_modes_resolved)
+        self.assertTrue(self.damage(source))
+
+    def test_subsurface_damage_does_not_wait_for_toplevel_map(self):
+        source = self.make_source(SubsurfaceWindowSource)
+        source.parent_wid = 9
+
+        self.assertTrue(self.damage(source))
+
+    def test_safe_initial_damage_paths_remain_immediate(self):
+        cases = {
+            "picture fallback": {"non_video_encodings": ("rgb24",)},
+            "global csc modes": {"full_csc_modes": typedict({"h264": ("YUV420P",)})},
+            "override redirect": {"is_OR": True},
+            "tray": {"is_tray": True},
+            "shadow": {"is_shadow": True},
+            "no video encoding": {"common_video_encodings": ()},
+        }
+        for name, values in cases.items():
+            with self.subTest(name=name):
+                source = self.make_source()
+                for key, value in values.items():
+                    setattr(source, key, value)
+                self.assertTrue(self.damage(source))
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/unittests/unit/wayland/window_test.py
+++ b/tests/unittests/unit/wayland/window_test.py
@@ -30,6 +30,99 @@ def load_window_server_class():
 WaylandWindowServer = load_window_server_class()
 
 
+class WaylandWindowServerInitialWindowTest(unittest.TestCase):
+
+    @staticmethod
+    def make_server(geometry):
+        server = object.__new__(WaylandWindowServer)
+        window = Mock()
+        window.is_managed.return_value = True
+        window.is_tray.return_value = False
+        window.is_OR.return_value = False
+        window.get_property.return_value = geometry
+        server._id_to_window = {7: window}
+        server.client_properties = {}
+        return server, window
+
+    def test_zero_size_window_is_announced_once_when_ready(self):
+        from xpra.net.packet_type import WINDOW_CREATE
+
+        server, window = self.make_server((0, 0, 0, 0))
+        source = Mock(uuid="client")
+
+        with patch.object(WaylandWindowServer, "_do_send_new_window_packet") as send_window:
+            server.send_initial_windows(source)
+            server.update_size(window, (800, 600))
+            window.get_property.return_value = (0, 0, 800, 600)
+            server.update_size(window, (800, 600))
+
+        window.hide.assert_not_called()
+        source.new_window.assert_not_called()
+        source.damage.assert_not_called()
+        send_window.assert_called_once_with(
+            WINDOW_CREATE, window, (0, 0, 800, 600),
+        )
+
+    def test_map_applies_properties_before_first_refresh(self):
+        from xpra.net.common import Packet
+        from xpra.server.subsystem.window import WindowServer
+        from xpra.server.window.compress import WindowSource
+        from xpra.server.window.video_compress import WindowVideoSource
+
+        proto = Mock()
+        window = Mock()
+        surface = Mock()
+        server = Mock()
+        server.get_window.return_value = window
+        server.get_surface.return_value = surface
+        source = object.__new__(WindowVideoSource)
+        source.wid = 7
+        source._client_csc_modes_resolved = False
+        source.is_OR = False
+        source.is_tray = False
+        source.is_shadow = False
+        source.common_video_encodings = ("h264",)
+        source.non_video_encodings = ()
+        source.full_csc_modes = typedict()
+        connection = Mock(uuid="client")
+        server.get_server_source.return_value = connection
+        events = []
+
+        def set_client_properties(_wid, _window, properties):
+            events.append("properties")
+            source.set_client_properties(properties)
+
+        connection.set_client_properties.side_effect = set_client_properties
+        server._set_client_properties.side_effect = (
+            lambda *args: WindowServer._set_client_properties(server, *args)
+        )
+        surface.resize.side_effect = lambda *_args: events.append("resize")
+        server.server.compositor.flush.side_effect = lambda: events.append("flush")
+
+        def refresh_window(_window):
+            events.append("refresh")
+            source.damage(0, 0, 800, 600)
+
+        server.refresh_window.side_effect = refresh_window
+        properties = {"encoding.full_csc_modes": {"h264": ("YUV420P",)}}
+
+        with patch.object(WindowSource, "set_client_properties"), \
+                patch.object(WindowSource, "damage") as damage:
+            source.damage(0, 0, 800, 600)
+            damage.assert_not_called()
+
+            WaylandWindowServer._process_map(
+                server, proto, Packet("window-map", 7, 0, 0, 800, 600, properties),
+            )
+
+        self.assertEqual(events, ["properties", "resize", "flush", "refresh"])
+        self.assertTrue(source._client_csc_modes_resolved)
+        damage.assert_called_once_with(0, 0, 800, 600, None)
+        server._set_client_properties.assert_called_once_with(
+            proto, 7, window, properties | {"event": "map"},
+        )
+
+
 class WaylandWindowServerCommitTest(unittest.TestCase):
 
     @staticmethod


xpra/server/subsystem/window.py
```diff
@@ -634,13 +636,17 @@ def send_initial_windows(self, ss, sharing=False) -> None:
                     window.move_resize(-200, -200, w, h)
             else:
                 # code more or less duplicated from _send_new_window_packet:
+                x, y, w, h = window.get_property("geometry")
+                if self.WAIT_FOR_INITIAL_WINDOW_SIZE and (w <= 0 or h <= 0):
```
totaam: There's no need for this toggle.
X11 servers will never have width or height set to 0.

xpra/server/subsystem/window.py
```diff
@@ -634,13 +636,17 @@ def send_initial_windows(self, ss, sharing=False) -> None:
...
+                    continue
                 if not sharing and not window.is_OR():
                     window.hide()
-                x, y, w, h = window.get_property("geometry")
                 wprops = self.client_properties.get(wid, {}).get(ss.uuid, {})
                 packet_type = "new-override-redirect" if (window.is_OR() and BACKWARDS_COMPATIBLE) else WINDOW_CREATE
                 ss.new_window(packet_type, wid, window, x, y, w, h, wprops)
-                ss.damage(wid, window, 0, 0, w, h)
+                if self.SEND_INITIAL_DAMAGE:
+                    ss.damage(wid, window, 0, 0, w, h)
```
totaam: Sending the damage later means that the window may appear blank for one whole roundtrip.
And on a slow connection, this can take a while.

Is this really needed? or can it be gated with `w==0 or h==0`?

xpra/wayland/server/subsystem/window.py
```diff
@@ -41,6 +41,11 @@
 
 
 class WaylandWindowServer(WindowServer):
+    # The client's map packet carries the backing-specific encoding properties
```
totaam: This doesn't explain why we would ever need to wait for the map packet.
What does it carry that is so important that we have to show a blank window waiting for it?

kogeler:
Thanks. I removed the broad backend toggles and narrowed both parts of the change.

The common initial-window path now skips non-positive dimensions before hiding, announcing, or damaging a window. As noted in the review, X11 never supplies such a size, so its behavior is unchanged. A size check alone does not solve the H.264 case, because the clean failure used a valid 1536x1095 window.

Initial damage is now delayed only for a normal top-level, video-only source with no picture fallback and no usable global or per-window CSC mode. RGB and other picture-capable clients remain immediate, as do clients with usable global CSC modes and OR, tray, shadow, and subsurface windows. The map applies the backing properties before its refresh, and that refresh releases the first frame.

<details>
<summary>Why the valid-size H.264 window still has to wait</summary>

On clean [`521db02`](https://github.com/Xpra-org/xpra/commit/521db02f143d9f0a57c656658d9ee12572622276), initial damage for the 1536x1095 RGBX window was queued at `17:12:58.009`. Pipeline selection failed at `.010` with `no video pipeline options found` because the window source had no usable client CSC mode and no picture fallback.

The map arrived at `.039` with the backing-specific value `encoding.full_csc_modes = {'h264': ('YUV420P',), ...}`. That value is calculated after the client window backing exists and is what allows the hardware H.264 pipeline to be selected. An explicit per-window CSC dictionary, including an empty one, or the map event resolves the wait; this does not impose a round-trip delay on the immediate paths listed above.

</details>

<details>

The narrowed change was validated together with the two related follow-ups on `521db02`. The combined focused suite passed 24/24 tests, and the native Wayland suite passed 15/15.

The physical-GPU scenario passed both strict profiles. RGB delivered 11 RGB-only updates. H.264 delivered 18 H.264-only updates with no startup or pre-negotiation errors, including a 16-frame hardware stream with VA-API `EncSlice` and `VLD` plus native NV12 OpenGL rendering. Both runs rendered the expected pixels, accepted real input, and completed the expected lifecycle cleanup.

</details>

xpra/server/subsystem/window.py
kogeler: fixed

xpra/server/subsystem/window.py
kogeler: fixed

xpra/wayland/server/subsystem/window.py
kogeler: fixed

xpra/server/subsystem/window.py
```diff
@@ -634,9 +634,12 @@ def send_initial_windows(self, ss, sharing=False) -> None:
                     window.move_resize(-200, -200, w, h)
             else:
                 # code more or less duplicated from _send_new_window_packet:
+                x, y, w, h = window.get_property("geometry")
+                if w <= 0 or h <= 0:
+                    log("not sending initial window %#x with invalid dimensions %ix%i", wid, w, h)
+                    continue
```
totaam, line 643: What is going to send the "new-window" packet for this window then?
We have already created a model for it, so it will not go through this method again.

xpra/server/window/video_compress.py
```diff
@@ -231,6 +231,7 @@ def init_vars(self) -> None:
         self._csc_encoder: ColorspaceConverter | None = None
         self._video_encoder: VideoEncoder | None = None
         self._last_pipeline_check = 0
+        self._client_csc_modes_resolved = False
```
totaam, line 234: This is unrelated to the "new-window" issue above and should be separated out.

xpra/server/window/video_compress.py
```diff
@@ -281,6 +282,27 @@ def do_set_auto_refresh_delay(self, min_delay, delay) -> None:
...
+            not self.non_video_encodings,
+            not any(self.full_csc_modes.strtupleget(encoding) for encoding in self.common_video_encodings),
+        ))
+
+    def damage(self, x: int, y: int, w: int, h: int, options=None) -> None:
+        if self.waiting_for_client_csc_modes():
+            videolog("dropping damage for window %#x until client CSC modes are known", self.wid)
+            return
+        super().damage(x, y, w, h, options)
+
```
totaam, line 305: That's a huge amount of code for something that should be extremely rare.
You should never configure an xpra client without having some fallback picture encodings. Xpra is not designed to work well without them.
And I don't want to have to maintain this level of complication for a use case that is blatantly doing something wrong - unless you can convince me otherwise.

totaam (changes requested):

kogeler:
@totaam 

Thanks. I tested the suggested picture fallback on current master before
rebasing the change. It is a valid way to keep the application visible with an
adaptive transport, but it cannot satisfy this profile's requirement to use
H.264.

The result is unambiguous: `--encodings=h264,rgb` sent 20 `rgb32` updates and
zero H.264 updates. Adding `--encoding=h264` still sent seven `rgb32` updates
and zero H.264 updates. The latter option stores H.264 as a preference, but the
new window source retains the default `strict=False` selection policy. Because
this window has alpha, the normal selector then uses the transparent-encoding
path, which chooses `rgb32` when RGB is available. The workaround succeeds by
changing the wire encoding to RGB; it does not make H.264 reachable.

Removing RGB does force this workload onto H.264, and that is the control that
reproduces the pre-map pipeline error. Enabling the server-wide strict selector
would take the same strict H.264 branch; it would not make the still-missing
backing CSC modes arrive before map. Thus the proposed client fallback and a
clean H.264 start cannot be obtained together without fixing this ordering.

This distinction matters for the reported setup: it deliberately verifies
native H.264 encode and decode, server-side CSC, and direct NV12 OpenGL without
a picture or client-CSC fallback. With only H.264 allowed, the server has no
RGB escape path and initial damage reaches pipeline selection before the map
packet supplies the concrete backing's H.264 CSC modes. With this patch, that
same strict profile produces 19 H.264 updates and zero RGB updates without the
startup error.

The gate is limited to an unresolved normal top-level video-only source with
no picture fallback and no usable global CSC mode. Picture-capable clients,
usable global modes, override-redirect, tray, shadow, and subsurface sources
remain immediate. Map properties are applied before the existing full refresh,
which releases the first strict frame; there is no timer, retry, or polling.

I also kept the `0x0` guard in this PR. It is a different readiness condition,
not the cause of the H.264 failure, but both bugs are in the same initial
attach operation: `send_initial_windows()` must not emit create/damage state
before the registered Wayland model is usable. The existing release paths are
different and remain unchanged: `update_size()` announces a `0x0` model when
it first becomes valid, while map refresh releases the valid-size video frame
after applying the backing properties.

The deferred create already has a concrete sender in current master:
[`new_surface()`](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/wayland/server/subsystem/window.py#L77-L104)
registers a `0x0` model without sending `WINDOW_CREATE`, and
[`update_size()`](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/wayland/server/subsystem/window.py#L301-L309)
sends that packet on its first `0x0 -> nonzero` transition. The model does not
need to enter `send_initial_windows()` again.

<details>
<summary>Why explicit H.264 still becomes RGB</summary>

The behavior follows directly from current master:

1. The client places its selected encoding in the encoding capabilities
   ([`get_encodings_caps()`](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/client/subsystem/encoding.py#L315-L335)).
2. The server parses that value through `set_encoding()` without setting its
   `strict` argument, whose default is `False`
   ([cap parsing](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/source/encoding.py#L319-L335),
   [`set_encoding()`](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/source/encoding.py#L458-L490)).
   At startup it stores the preference on the connection; the new window
   source receives that encoding while its strict flag keeps the default false
   value
   ([source construction](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/source/window.py#L506-L540),
   [`STRICT_MODE` default](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/window/compress.py#L84-L92),
   [source initialization](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/window/compress.py#L457-L475)).
   Runtime encoding changes likewise pass `False` to existing sources.
3. Because strict selection is not enabled and the window wants alpha,
   `get_best_encoding_impl()` selects `get_transparent_encoding()`
   ([selector](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/window/compress.py#L1035-L1089)).
4. When RGB is among the common encodings, that function selects `rgb32` for
   this 32-bit lossless window
   ([transparent choice](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/window/compress.py#L1148-L1176)).

The concrete H.264 CSC answer cannot be sent earlier: the client calculates it
from the actual backing created for the new window
([backing setup](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/client/gui/window_base.py#L321-L326),
[backing properties](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/client/gui/window/backing.py#L542-L575))
and sends it with the map packet
([map packet](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/client/gtk3/window/base.py#L1283-L1304)).

The forced-fallback server log confirms the same path after H.264 was selected:

```text
non video encodings=('rgb24', 'rgb32')
wvs.update_encoding_selection(h264, None, True) full_csc_modes={}
using transparent encoding
```

</details>

<details>
<summary>Current-master A/B and regression results</summary>

All physical runs used exact `1d8e112`, the same Zed/RADV workload, render
node, native OpenGL client, and default-alpha scenario.

| Server change / client policy | Saved packets | Result |
| --- | ---: | --- |
| no initial-state patch; `--encodings=h264` | 19 H.264, 0 RGB | startup pipeline error before map, then recovery |
| no initial-state patch; `--encodings=h264,rgb` | 0 H.264, 20 RGB | visible through RGB fallback |
| no initial-state patch; `--encodings=h264,rgb --encoding=h264` | 0 H.264, 7 RGB | visible through RGB fallback |
| patched; `--encodings=h264` | 19 H.264, 0 RGB | no startup or pre-negotiation error |
| patched RGB control | 0 H.264, 11 RGB | pass |

Every run exercised real pixels, pointer input, application lifecycle, and
cleanup. Both strict H.264 runs exercised hardware VA-API encode/decode and
native NV12 OpenGL; only the patched run met the clean-start boundary.

The test-only verification failed all five new production-path assertions on
unmodified `1d8e112`; the patch passed 21/21 focused tests. The native Wayland
gate passed its build/import/linkage checks and 15/15 tests. Each complete test
mode completed 292 modules and retained only the unrelated
`unit.client.x11_client_paint_test` failure. Canonical
[Actions run 32964054292](https://github.com/Xpra-org/xpra/actions/runs/32964054292)
shows the same sole failure on unmodified `1d8e112` in all three modes. This
PR neither changes nor skips that test.

</details>

xpra/server/subsystem/window.py
totaam: Why was this comment marked as resolved?
Where? How?

xpra/server/window/video_compress.py
totaam: Again, how was this resolved?

totaam:
Please don't just mark review comments as resolved without an explanation.
I don't have time read wall of AI generated text and not get answers in return.

xpra/server/subsystem/window.py
kogeler: For the exact `0x0` case, the model does not enter `send_initial_windows()` again. `new_surface()` registers it without sending `WINDOW_CREATE`. The first map or commit that changes its size from `0x0` to a size with both dimensions non-zero calls
`update_size()`, and that method sends `WINDOW_CREATE` to the current window sources. That existing `update_size()` call is the deferred sender: https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/wayland/server/subsystem/window.py#L301-L309

xpra/server/window/video_compress.py
kogeler: The `0x0` create guard and the H.264 pre-map damage guard are independent conditions. I grouped them because both originate in the `send_initial_windows()` loop used when a client attaches, and both are required to fix the complete scenario reported in #5022. Their release paths and tests remain separate.

If you consider that scope too broad, I can split the `0x0` guard and its tests from the H.264 wait and its tests into two PRs.

xpra/server/window/video_compress.py
kogeler: I tested the suggested fallback on the exact PR base and workload. With
`--encodings=h264,rgb`, the client received 20 `rgb32` updates and no H.264
updates. Adding `--encoding=h264` still produced seven `rgb32` updates and no
H.264 updates. In this path `--encoding=h264` is applied with `strict=False`,
so the alpha window selects the transparent `rgb32` path when RGB is available.

The patched `--encodings=h264` control produced 19 H.264 updates and no RGB
updates without the startup error. Picture fallback therefore keeps the window
visible, but it changes this workload from H.264 to RGB and does not satisfy
the reported H.264-only profile.

<details>
<summary>Why selecting H.264 still reaches the RGB path</summary>

1. During startup, the server parses the client's selected encoding by calling
   `set_encoding()` without a `strict` argument, whose default is `False`
   ([capability parsing](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/source/encoding.py#L319-L335),
   [`set_encoding()`](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/source/encoding.py#L458-L492)).
2. A new window source receives that connection-level H.264 selection, while
   its strict flag keeps the default `False` value
   ([source construction](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/source/window.py#L506-L540),
   [strict default](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/window/compress.py#L88-L90),
   [source initialization](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/window/compress.py#L457-L474)).
3. Because selection is not strict and this window requires alpha, the normal
   selector chooses `get_transparent_encoding()` instead of the selected H.264
   encoder
   ([selection branch](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/window/compress.py#L1035-L1089)).
4. With RGB available, that transparent selector chooses `rgb32` for this
   32-bit lossless window
   ([transparent encoding choice](https://github.com/Xpra-org/xpra/blob/1d8e112c4a03c2ac6d56f9488843fc6ef2f7e57b/xpra/server/window/compress.py#L1148-L1176)).

kogeler:
> Please don't just mark review comments as resolved without an explanation. I don't have time read wall of AI generated text and not get answers in return.

Sorry, done.

xpra/server/subsystem/window.py
totaam: OK, this one makes sense - and I will happily merge this fix.
On its own, without the rest.

totaam:
> Sorry, done.

Not quite. Not all comments were addressed properly.

Ignoring windows with 0x0 dimensions makes sense since we do send the "new-window" packet later. That's one thing.

---

As for the rest, it does not make any sense at all.
Please remove all of it from this PR and start again.
`h264` does not and will not handle transparency.

If your window claims to use transparency, then xpra will correctly use an encoding that supports that.
Which is rgb32 when you only specify `encodings=h264,rgb32`.

As documented extensively elsewhere, and mentioned in this very ticket - specifying a very limited set of encodings is pathological and should never be done. This will only cause problems and lead you down terrible paths.
Don't do that.

There are other ways of dealing with windows with transparency:
* refuse to honour the transparency
* detect that the window content-type is in fact not transparent (we do this for "browser" application types)
* use video encodings that do support transparency
* etc

Trying to damage the xpra source code in order to make it so something that is fundamentally wrong will not be merged.

kogeler:
Thank you for your time and for all the discussion around these PRs.

For context, I validate each of these fixes against a real set of live end-to-end tests, covering actual data transfer as well as encoding and decoding on real hardware. I’ve already automated this as part of a CI/CD pipeline that continuously exercises the full flow, so the changes I proposed are driven by issues I can reproduce in my specific environment.

That said, I think that by trying to adapt the upstream project to fit my rather narrow use case, I’m starting to spend both your time and mine on requirements that may simply not belong in the upstream project.

So rather than continuing the discussion, I’ll close these PRs and maintain a fork instead. I’ll keep it in sync with upstream while carrying the small set of patches needed for my particular use case.

Thanks again for your help, for the explanations that helped me better understand the project, and especially for the time you’ve spent reviewing and discussing my changes. I really appreciate it.

kogeler:
> As documented extensively elsewhere, and mentioned in this very ticket - specifying a very limited set of encodings is pathological and should never be done. This will only cause problems and lead you down terrible paths.
> Don't do that.

If the project allows combinations of CLI options that can lead to unpredictable behavior or violations of its own contracts, then perhaps the CLI model itself is worth reconsidering.

totaam:
> I’ll keep it in sync with upstream while carrying the small set of patches needed for my particular use case.

FYI: I've seen this done before, and the results were... problematic for their authors.
Your patches take the wrong approach to begin with and will inevitably be:
* hard to maintain
* break other things

So please, if you ever make other bug reports - make sure to remove your patches first.

totaam:
> then perhaps the CLI model itself is worth reconsidering.

Extra warnings will be added.
But you are expected to RTFM before diving and making code changes.
This IS documented, thoroughly.

kogeler:
> So please, if you ever make other bug reports - make sure to remove your patches first.

My testing flow always includes verification against a clean master branch. I never report issues that I haven’t first reproduced and confirmed on a clean master.

totaam:
The valid fixes are in 057eb805f33f109ac27af637cdecac8f01b74edf and dbc575b063abbc6df314a63ee530fdb218776327 and will be backported if appropriate.

kogeler:
> The valid fixes are in [057eb80](https://github.com/Xpra-org/xpra/commit/057eb805f33f109ac27af637cdecac8f01b74edf) and [dbc575b](https://github.com/Xpra-org/xpra/commit/dbc575b063abbc6df314a63ee530fdb218776327) and will be backported if appropriate.

Thanks! Yes it is useful but it still doesn't fully fix confirmed real issue including the case where many encoding specified (tested on clean upstream).

